PR #58, opened 2020-05-14
[WIP] Replace unwrap calls in asyncgit with error handling - closes #53

### LINKED ISSUE #53: more error handling
I started `asyncgit` crate ignoring errors `unwrap` is all over the place but the goal would be to handle them correctly like with `sync::get_commits_info`

MCord:
This pull request tries to replace `unwrap()` calls in asyncgit with error handling via a custom error type defined in `error.rs`. changes has been split into separate commits.

* Added a new error type in `error.rs` file (dependent on `thiserror` crate)
* Added a `Result<T>` type alias for `Result<T,Error>`
* Removed all `unwrap()` calls from `asyncgit` lib except a few in callbacks that I don't know how to deal with for the moment.

I have not touched `unwrap()` inside tests as I am not sure yet if we want to remove them too. I assume they are alright in tests. I will continue to change other files when I get feedback about the code.

### DIFF
--- a/asyncgit/src/error.rs
+++ b/asyncgit/src/error.rs
@@ -6,14 +6,14 @@ pub enum Error {
     Generic(String),
 
     #[error("io error")]
-    Io {
-        #[from]
-        source: std::io::Error,
-    },
+    Io(#[from] std::io::Error),
 
     #[error("git error")]
-    Git {
+    Git(#[from] git2::Error),
+
+    #[error("encoding error.")]
+    Encoding {
         #[from]
-        source: git2::Error,
+        source: std::string::FromUtf8Error,
     },
 }
--- a/asyncgit/src/error.rs
+++ b/asyncgit/src/error.rs
@@ -1,0 +1,22 @@
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("`{0}`")]
+    Generic(String),
+
+    #[error("io error")]
+    Io {
+        #[from]
+        source: std::io::Error,
+    },
+
+    #[error("git error")]
+    Git {
+        #[from]
+        source: git2::Error,
+    },
+
+    #[error(transparent)]
+    Other(#[from] Box<dyn std::error::Error>),
+}
--- a/asyncgit/src/error.rs
+++ b/asyncgit/src/error.rs
@@ -16,7 +16,4 @@ pub enum Error {
         #[from]
         source: git2::Error,
     },
-
-    #[error(transparent)]
-    Other(#[from] Box<dyn std::error::Error>),
 }
--- a/asyncgit/src/error.rs
+++ b/asyncgit/src/error.rs
@@ -13,20 +13,9 @@ pub enum Error {
 
     #[error("encoding error:{0}")]
     Encoding(#[from] std::string::FromUtf8Error),
-
-    #[error("unspecified error:{0}")]
-    Unspecified(Box<dyn std::error::Error>),
 }
 
-pub type Returns<T> = std::result::Result<T, Error>;
-
-// impl From<&dyn std::error::Error> for Error {
-//     fn from(error: &dyn std::error::Error) -> Self {
-//         let e = error.clone().to_owned();
-//         let b = Box::new(e);
-//         Error::Unspecified(b)
-//     }
-// }
+pub type Result<T> = std::result::Result<T, Error>;
 
 impl<T> From<std::sync::PoisonError<T>> for Error {
     fn from(error: std::sync::PoisonError<T>) -> Self {
--- a/asyncgit/src/error.rs
+++ b/asyncgit/src/error.rs
@@ -5,15 +5,31 @@ pub enum Error {
     #[error("`{0}`")]
     Generic(String),
 
-    #[error("io error")]
+    #[error("io error:{0}")]
     Io(#[from] std::io::Error),
 
-    #[error("git error")]
+    #[error("git error:{0}")]
     Git(#[from] git2::Error),
 
-    #[error("encoding error.")]
-    Encoding {
-        #[from]
-        source: std::string::FromUtf8Error,
-    },
+    #[error("encoding error:{0}")]
+    Encoding(#[from] std::string::FromUtf8Error),
+
+    #[error("unspecified error:{0}")]
+    Unspecified(Box<dyn std::error::Error>),
+}
+
+pub type Returns<T> = std::result::Result<T, Error>;
+
+// impl From<&dyn std::error::Error> for Error {
+//     fn from(error: &dyn std::error::Error) -> Self {
+//         let e = error.clone().to_owned();
+//         let b = Box::new(e);
+//         Error::Unspecified(b)
+//     }
+// }
+
+impl<T> From<std::sync::PoisonError<T>> for Error {
+    fn from(error: std::sync::PoisonError<T>) -> Self {
+        Error::Generic(format!("poison error: {}", error))
+    }
 }
--- a/asyncgit/src/error.rs
+++ b/asyncgit/src/error.rs
@@ -10,9 +10,6 @@ pub enum Error {
 
     #[error("git error:{0}")]
     Git(#[from] git2::Error),
-
-    #[error("encoding error:{0}")]
-    Encoding(#[from] std::string::FromUtf8Error),
 }
 
 pub type Result<T> = std::result::Result<T, Error>;
--- a/asyncgit/src/lib.rs
+++ b/asyncgit/src/lib.rs
@@ -5,6 +5,7 @@
 #![deny(clippy::all)]
 
 mod diff;
+mod error;
 mod revlog;
 mod status;
 pub mod sync;
--- a/asyncgit/src/lib.rs
+++ b/asyncgit/src/lib.rs
@@ -20,6 +20,7 @@ pub use crate::{
         utils::is_repo,
     },
 };
+use error::{Error, Returns};
 use std::{
     collections::hash_map::DefaultHasher,
     hash::{Hash, Hasher},
@@ -48,9 +49,11 @@ pub fn hash<T: Hash + ?Sized>(v: &T) -> u64 {
 }
 
 /// helper function to return the current tick since unix epoch
-pub fn current_tick() -> u64 {
-    SystemTime::now()
+pub fn current_tick() -> Returns<u64> {
+    Ok(SystemTime::now()
         .duration_since(UNIX_EPOCH)
-        .unwrap()
-        .as_millis() as u64
+        .map_err(|_| {
+            Error::Generic("can't get system time".to_string())
+        })?
+        .as_millis() as u64)
 }
--- a/asyncgit/src/lib.rs
+++ b/asyncgit/src/lib.rs
@@ -20,7 +20,6 @@ pub use crate::{
         utils::is_repo,
     },
 };
-use error::{Error, Result};
 use std::{
     collections::hash_map::DefaultHasher,
     hash::{Hash, Hasher},
@@ -49,11 +48,9 @@ pub fn hash<T: Hash + ?Sized>(v: &T) -> u64 {
 }
 
 /// helper function to return the current tick since unix epoch
-pub fn current_tick() -> Result<u64> {
-    Ok(SystemTime::now()
+pub fn current_tick() -> u64 {
+    SystemTime::now()
         .duration_since(UNIX_EPOCH)
-        .map_err(|_| {
-            Error::Generic("can't get system time".to_string())
-        })?
-        .as_millis() as u64)
+        .unwrap()
+        .as_millis() as u64
 }
--- a/asyncgit/src/lib.rs
+++ b/asyncgit/src/lib.rs
@@ -20,7 +20,7 @@ pub use crate::{
         utils::is_repo,
     },
 };
-use error::{Error, Returns};
+use error::{Error, Result};
 use std::{
     collections::hash_map::DefaultHasher,
     hash::{Hash, Hasher},
@@ -49,7 +49,7 @@ pub fn hash<T: Hash + ?Sized>(v: &T) -> u64 {
 }
 
 /// helper function to return the current tick since unix epoch
-pub fn current_tick() -> Returns<u64> {
+pub fn current_tick() -> Result<u64> {
     Ok(SystemTime::now()
         .duration_since(UNIX_EPOCH)
         .map_err(|_| {
--- a/asyncgit/src/sync/diff.rs
+++ b/asyncgit/src/sync/diff.rs
@@ -118,7 +118,6 @@ pub(crate) fn get_diff_raw<'a>(
     Ok(diff)
 }
 
-//TODO: return Option
 ///
 pub fn get_diff(
     repo_path: &str,
--- a/asyncgit/src/sync/diff.rs
+++ b/asyncgit/src/sync/diff.rs
@@ -304,21 +304,21 @@ mod tests {
         let root = repo.path().parent().unwrap();
         let repo_path = root.as_os_str().to_str().unwrap();
 
-        assert_eq!(get_statuses(repo_path).unwrap(), (0, 0));
+        assert_eq!(get_statuses(repo_path), (0, 0));
 
         File::create(&root.join(file_path))
             .unwrap()
             .write_all(b"test\nfoo")
             .unwrap();
 
-        assert_eq!(get_statuses(repo_path).unwrap(), (1, 0));
+        assert_eq!(get_statuses(repo_path), (1, 0));
 
         assert_eq!(
             stage_add_file(repo_path, file_path).unwrap(),
             true
         );
 
-        assert_eq!(get_statuses(repo_path).unwrap(), (0, 1));
+        assert_eq!(get_statuses(repo_path), (0, 1));
 
         let diff = get_diff(
             repo_path,
@@ -362,7 +362,7 @@ mod tests {
         let root = repo.path().parent().unwrap();
         let repo_path = root.as_os_str().to_str().unwrap();
 
-        assert_eq!(get_statuses(repo_path).unwrap(), (0, 0));
+        assert_eq!(get_statuses(repo_path), (0, 0));
 
         let file_path = root.join("bar.txt");
 
@@ -381,7 +381,7 @@ mod tests {
         let res =
             stage_add_file(repo_path, Path::new("bar.txt")).unwrap();
         assert_eq!(res, true);
-        assert_eq!(get_statuses(repo_path).unwrap(), (0, 1));
+        assert_eq!(get_statuses(repo_path), (0, 1));
 
         // overwrite with next content
         {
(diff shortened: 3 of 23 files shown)


extrawurst:
Hey @MCord thanks for getting at this!
Let me annotate some questions that I have. Please keep in mind as you see in the codebase, I am also no expert on error handling in Rust yet :)

asyncgit/src/error.rs
```diff
@@ -0,0 +1,22 @@
...
+    },
+
+    #[error("git error")]
+    Git {
+        #[from]
+        source: git2::Error,
+    },
+
+    #[error(transparent)]
+    Other(#[from] Box<dyn std::error::Error>),
```
extrawurst, line 21: what is this guy used for? 🙈

asyncgit/src/sync/diff.rs
```diff
@@ -113,13 +120,21 @@ pub(crate) fn get_diff_raw<'a>(
 
 //TODO: return Option
```
extrawurst, line 121: this can go away now :)

asyncgit/src/sync/diff.rs
```diff
@@ -113,13 +120,21 @@ pub(crate) fn get_diff_raw<'a>(
...
+) -> Result<FileDiff, Error> {
     scope_time!("get_diff");
 
     let repo = utils::repo(repo_path);
-    let repo_path = repo.path().parent().unwrap();
-
-    let diff = get_diff_raw(&repo, &p, stage, false).unwrap();
+    let repo_path = repo.path().parent().ok_or_else(|| {
+        Error::Generic(
+            "repositories located at root are not supported."
```
extrawurst, line 133: funny I never really thought about how this can fail. need to test that :D

asyncgit/src/error.rs
MCord: I thought this would cover the conversion from any Error trait but indeed it does not work the way I thought it would. If we want a catch all error we need to add a dependency on `anyhow` crate. I am going to remove this variant.

asyncgit/src/sync/diff.rs
MCord: 👍 I missed that one.

asyncgit/src/sync/diff.rs
MCord: I admit I did not understand what the code is doing and why we need to get the parent folder. I have to read that code again but maybe we can write that differently so that we don't have this restriction ?

asyncgit/src/error.rs
```diff
@@ -0,0 +1,19 @@
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("`{0}`")]
+    Generic(String),
+
+    #[error("io error")]
+    Io {
```
extrawurst, line 9: How do you feel about making them `Io(#[from]std::io::Error)` and `Git` accordingly? I mean if we only have one element I usually use `struct Foo(bool)` aswell.

asyncgit/src/sync/diff.rs
extrawurst: I dug into it: `workdir` is actually what we want here instead of `path` and this we can unwrap safely since `utils::repo` will panic if it is a bare repo.

extrawurst:
Really good work!

Two more things:
1. According `unwrap` in tests is fine
2. Will you convert `sync::utils` aswell? then we have repo returning Result which is used *everywhere* and also allow the `unwrap` for `repo.workdir` (see other conversation)

asyncgit/src/error.rs
MCord: Good idea, there was also a backtrace in there but I removed it since it's unstable. I will do the change.

MCord: done.

MCord:
Thanks ❤️ !

I will remove every unwrap outside the tests and update the PR. I don't mind changing `sync::utils` everywhere since it gives me chance to read more of the project.

asyncgit/src/sync/diff.rs
MCord: makes sense, do you want to do it in a separate PR or you want me to fix it in this one ?

extrawurst: Yeah good point, let's split it off and try to reduce the changes aside of the errors to a minimum.

asyncgit/src/lib.rs
```diff
@@ -47,9 +49,11 @@ pub fn hash<T: Hash + ?Sized>(v: &T) -> u64 {
 }
 
 /// helper function to return the current tick since unix epoch
-pub fn current_tick() -> u64 {
-    SystemTime::now()
+pub fn current_tick() -> Returns<u64> {
```
extrawurst, line 52: I wonder isn't it ok to unwrap here instead of polluting the API with a result? I mean what else must be wrong if `SystemTime::now()` is before UNIX_EPOCH 🙈

asyncgit/src/error.rs
```diff
@@ -0,0 +1,35 @@
...
+    Git(#[from] git2::Error),
+
+    #[error("encoding error:{0}")]
+    Encoding(#[from] std::string::FromUtf8Error),
+
+    #[error("unspecified error:{0}")]
+    Unspecified(Box<dyn std::error::Error>),
+}
+
+pub type Returns<T> = std::result::Result<T, Error>;
```
extrawurst, line 21: I have seen many libs doing this and naming their thing `Result` - what is the advantage of naming it differently? I mean people know the concept of `Result` right when they see it

asyncgit/src/error.rs
```diff
@@ -0,0 +1,35 @@
...
+    Io(#[from] std::io::Error),
+
+    #[error("git error:{0}")]
+    Git(#[from] git2::Error),
+
+    #[error("encoding error:{0}")]
+    Encoding(#[from] std::string::FromUtf8Error),
+
+    #[error("unspecified error:{0}")]
+    Unspecified(Box<dyn std::error::Error>),
```
extrawurst, line 18: I still need to understand where this is used?

asyncgit/src/error.rs
```diff
@@ -0,0 +1,35 @@
...
+}
+
+pub type Returns<T> = std::result::Result<T, Error>;
+
+// impl From<&dyn std::error::Error> for Error {
```
extrawurst, line 23: I strongly believe commented code does not need to be committed :)

asyncgit/src/error.rs
```diff
@@ -0,0 +1,35 @@
...
+//     fn from(error: &dyn std::error::Error) -> Self {
+//         let e = error.clone().to_owned();
+//         let b = Box::new(e);
+//         Error::Unspecified(b)
+//     }
+// }
+
+impl<T> From<std::sync::PoisonError<T>> for Error {
```
extrawurst, line 31: can't we use the `#[from]` to generate this?

asyncgit/src/sync/hooks.rs
```diff
@@ -14,37 +15,40 @@ const HOOK_COMMIT_MSG: &str = ".git/hooks/commit-msg";
...
     if hook_runable(repo_path, HOOK_COMMIT_MSG) {
-        let mut file = NamedTempFile::new().unwrap();
+        let mut file = NamedTempFile::new()?;
 
-        write!(file, "{}", msg).unwrap();
+        write!(file, "{}", msg)?;
 
-        let file_path = file.path().to_str().unwrap();
+        let file_path = file.path().to_str().ok_or_else(|| {
+            Error::Generic("can't get temp file's path".to_string())
```
extrawurst, line 27: reading [Path::to_str docs}(https://doc.rust-lang.org/std/path/struct.Path.html#method.to_str) I believe this can only be None for invalid unicode, we should reflect this in our generic error

asyncgit/src/lib.rs
MCord: I went for removing all unwrap but I agree I might have gone too far :D. Fixed

asyncgit/src/sync/hunks.rs
```diff
@@ -12,21 +12,22 @@ pub fn stage_hunk(
     repo_path: &str,
     file_path: String,
     hunk_hash: u64,
-) -> bool {
+) -> Result<bool, Error> {
```
extrawurst: let's convert this to Result<()> since the bool really was only used to indicate failure before and will be handled by the Result type now

asyncgit/src/error.rs
MCord: I thought it might conflict with the name from standard lib but this seems to be the accepted convention so I renamed.

asyncgit/src/error.rs
MCord: I was still experimenting with a generic error for special cases, was not supposed to be commited. fixed

asyncgit/src/error.rs
MCord: we agree. my mistake.

MCord: fixed.

asyncgit/src/error.rs
MCord: My rust knowledge fails me here. `PoisonError` is generic and so I need to make my `Error` struct generic to add this case ?

asyncgit/src/sync/mod.rs
```diff
@@ -25,80 +25,86 @@ pub use utils::{
...
                 .current_dir(path)
-                .output()
+                .output()?
         };
 
-        let output = output.unwrap();
-        let stdout = String::from_utf8(output.stdout).unwrap();
-        let stderr = String::from_utf8(output.stderr).unwrap();
-        format!(
+        let stdout = String::from_utf8(output.stdout)?;
```
extrawurst, line 100: lets keep the unwraps here. this is test only code. and if that was the reason to add the utf8 conversion errors, lets remove them - what do u think?

asyncgit/src/sync/reset.rs
```diff
@@ -1,45 +1,55 @@
 use super::utils::repo;
+use crate::error::{Error, Returns};
 use git2::{build::CheckoutBuilder, ObjectType, Status};
 use scopetime::scope_time;
 use std::{fs, path::Path};
 
 ///
-pub fn reset_stage(repo_path: &str, path: &Path) -> bool {
+pub fn reset_stage(repo_path: &str, path: &Path) -> Returns<bool> {
```
extrawurst: same as above `Result<()>` should be enough

asyncgit/src/sync/reset.rs
```diff
@@ -1,45 +1,55 @@
...
+
+    Ok(true)
 }
 
 ///
-pub fn reset_workdir_file(repo_path: &str, path: &str) -> bool {
+pub fn reset_workdir_file(
+    repo_path: &str,
+    path: &str,
+) -> Returns<bool> {
```
extrawurst: see above

asyncgit/src/sync/reset.rs
```diff
@@ -53,17 +63,22 @@ pub fn reset_workdir_file(repo_path: &str, path: &str) -> bool {
...
+        Ok(false)
     }
 }
 
 ///
-pub fn reset_workdir_folder(repo_path: &str, path: &str) -> bool {
+pub fn reset_workdir_folder(
+    repo_path: &str,
+    path: &str,
+) -> Returns<bool> {
```
extrawurst: ditto

asyncgit/src/sync/hooks.rs
MCord: updated.

asyncgit/src/sync/status.rs
```diff
@@ -67,32 +67,35 @@ impl Into<StatusShow> for StatusType {
...
-            )
-        } else {
-            e.path().unwrap().to_string()
+        let path = match e.head_to_index() {
+            Some(diff) => diff
+                .new_file()
+                .path()
+                .and_then(|x| x.to_str())
+                .map(String::from)
+                .ok_or_else(|| Error::Generic("".to_string()))?,
```
extrawurst, line 95: we should come up with something here and below

asyncgit/src/sync/hunks.rs
MCord: updated.

asyncgit/src/sync/mod.rs
MCord: you mean just the calls or the whole debug methods ?

asyncgit/src/sync/reset.rs
MCord: changed.

asyncgit/src/sync/reset.rs
MCord: this is a bit more tricky as we have branches that return false. I have to figure out an error message for those cases.

asyncgit/src/error.rs
extrawurst: good point lets keep it like that

asyncgit/src/sync/hooks.rs
```diff
@@ -64,20 +71,24 @@ pub enum HookResult {
...
 
     let output = output.expect("general hook error");
 
     if output.status.success() {
-        HookResult::Ok
+        Ok(HookResult::Ok)
     } else {
-        let err = String::from_utf8(output.stderr).unwrap();
-        let out = String::from_utf8(output.stdout).unwrap();
+        let err = String::from_utf8(output.stderr)?;
```
extrawurst, line 87: I am actually tempted to say lets use [from_utf8_lossy](https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf8_lossy) here and not add the utf8 error type to ours, what do you think?

asyncgit/src/sync/reset.rs
MCord: done

asyncgit/src/sync/mod.rs
extrawurst: I was wondering if we could remove the `FromUtf8Error` type from our Error when we unwrap here and use the lossy conversion in hooks?

asyncgit/src/sync/reset.rs
extrawurst: good point, lets keep it for now

asyncgit/src/sync/reset.rs
MCord: fixed

src/app.rs
```diff
@@ -220,10 +220,12 @@ impl App {
...
                         CWD,
                         reset_item.path.as_str(),
-                    ) {
+                    )
+                    .unwrap()
+                    {
                         flags.insert(NeedsUpdate::ALL);
                     }
-                } else if sync::reset_workdir_file(
+                } else if let Ok(()) = sync::reset_workdir_file(
```
extrawurst, line 226: I believe clippy will not like it :)

extrawurst: or not. I would still suggest using `sync::reset_workdir_file().is_ok()` or?

asyncgit/src/sync/status.rs
MCord: added a message but not sure of the wording.

asyncgit/src/sync/hooks.rs
MCord: I agree, UTF8 error types feel too specific here. I think it makes sense to provide a partially converted output with some miss interpreted chars instead of failing the hook if the script uses another encoding.  fixed.

asyncgit/src/sync/mod.rs
MCord: done.

src/app.rs
MCord: done.

asyncgit/src/sync/hooks.rs
```diff
@@ -64,20 +71,24 @@ pub enum HookResult {
     NotOk(String),
 }
 
-fn run_hook(path: &str, cmd: &str, args: &[&str]) -> HookResult {
+fn run_hook(
+    path: &str,
+    cmd: &str,
+    args: &[&str],
+) -> Result<HookResult> {
```
extrawurst, line 78: if we cannot fail anymore lets remove the `Result` return again, right? 🙈

asyncgit/src/sync/mod.rs
```diff
@@ -25,80 +25,81 @@ pub use utils::{
...
+        Ok((td, repo))
     }
 
     /// helper returning amount of files with changes in the (wd,stage)
-    pub fn get_statuses(repo_path: &str) -> (usize, usize) {
-        (
-            get_status(repo_path, StatusType::WorkingDir).len(),
-            get_status(repo_path, StatusType::Stage).len(),
-        )
+    pub fn get_statuses(repo_path: &str) -> Result<(usize, usize)> {
```
extrawurst, line 72: test code, no Result required

asyncgit/src/sync/mod.rs
```diff
@@ -25,80 +25,81 @@ pub use utils::{
...
+        Ok((
+            get_status(repo_path, StatusType::WorkingDir)?.len(),
+            get_status(repo_path, StatusType::Stage)?.len(),
+        ))
     }
 
     ///
-    pub fn debug_cmd_print(path: &str, cmd: &str) {
-        eprintln!("\n----\n{}", debug_cmd(path, cmd))
+    pub fn debug_cmd_print(path: &str, cmd: &str) -> Result<()> {
```
extrawurst, line 80: ditto

asyncgit/src/sync/mod.rs
```diff
@@ -25,80 +25,81 @@ pub use utils::{
...
+        let cmd = debug_cmd(path, cmd)?;
+        eprintln!("\n----\n{}", cmd);
+        Ok(())
     }
 
-    fn debug_cmd(path: &str, cmd: &str) -> String {
+    fn debug_cmd(path: &str, cmd: &str) -> Result<String> {
```
extrawurst, line 86: ditto

asyncgit/src/sync/hooks.rs
MCord: well we are cheating a bit in that function with expect ! should we keep it as is or forward the result ?

extrawurst:
Sorry to be such a pain in the ass 🙈! I am super glad you do all of this!

asyncgit/src/sync/mod.rs
MCord: fixed

asyncgit/src/sync/hooks.rs
extrawurst: good point, do a `if let` and return a generic `HookResult::NotOk("hook execution error")` in that case

asyncgit/src/sync/mod.rs
MCord: fixed.

asyncgit/src/sync/mod.rs
MCord: fixed.

MCord:
> 
> 
> Sorry to be such a pain in the ass 🙈! I am super glad you do all of this!

My pleasure! sorry you have to micro mange this PR :) It's my first lets hope It gets better with time 😂

asyncgit/src/sync/hooks.rs
MCord: I have used match to handle the `Err` case separately by formating the underlying error as the failure message. is that ok ?

extrawurst: sure looks good!

extrawurst:
Looks good to me! I will squash merge it if that's ok by you?

MCord:
Perfect, Thank you.

extrawurst:
Thank you @MCord! Great job!👍

PR #60, opened 2020-05-16
Implement git hook support on windows- closes 14

MCord:
This PR modifies the hook support to enable hook management on windows platform. in order for the code to work on windows I have made sure:
1. we don't try to chmod script files on windows (not needed / not supported)
2. we don't verify if scripts are executable on windows.
3. we use relative path for temp files to prevent path mapping issues for bash on windows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - introduced proper changelog
-- hook support on windows 
+- hook support on windows ([#14](https://github.com/extrawurst/gitui/issues/14))
 
 ### Changed
 - show longer commit messages in log view
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - introduced proper changelog
+- hook support on windows 
 
 ### Changed
 - show longer commit messages in log view
--- a/asyncgit/src/sync/hooks.rs
+++ b/asyncgit/src/sync/hooks.rs
@@ -1,17 +1,21 @@
-use crate::error::{Error, Result};
+use crate::error::Result;
 use scopetime::scope_time;
+use std::fs::File;
 use std::path::PathBuf;
 use std::{
     io::{Read, Write},
     path::Path,
     process::Command,
 };
-use tempfile::NamedTempFile;
 
 const HOOK_POST_COMMIT: &str = ".git/hooks/post-commit";
 const HOOK_COMMIT_MSG: &str = ".git/hooks/commit-msg";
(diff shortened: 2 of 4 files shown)


asyncgit/src/sync/hooks.rs
```diff
@@ -1,41 +1,41 @@
...
+use crate::error::Result;
 use scopetime::scope_time;
+use std::fs::File;
+use std::path::PathBuf;
 use std::{
     io::{Read, Write},
     path::Path,
     process::Command,
 };
-use tempfile::NamedTempFile;
```
extrawurst, line 9: IMO removing the usage of `NamedTempFile` allows us to move `tempfile` to the dev-dependencies?

asyncgit/src/sync/hooks.rs
```diff
@@ -70,20 +70,36 @@ pub enum HookResult {
...
+        .args(bash_args)
+        .current_dir(path)
+        .output();
+
+    let output = output.expect("general hook error");
+
+    if output.status.success() {
+        HookResult::Ok
+    } else {
+        let err = String::from_utf8(output.stderr).unwrap();
```
extrawurst, line 98: any reason to switch from `from_utf8_lossy` to `from_utf8`+`unwrap`?

asyncgit/src/sync/hooks.rs
```diff
@@ -193,3 +209,16 @@ exit 0
...
+#[cfg(not(windows))]
+fn is_executable(path: PathBuf) -> bool {
+    use is_executable::IsExecutable;
+    path.is_executable()
+}
+
+#[cfg(windows)]
+/// windows does not consider bash scripts to be executable so we consider everything
+/// to be executable (which is not far from the truth for windows platform.)
+fn is_executable(_: PathBuf) -> bool {
```
extrawurst, line 222: we expect the hooks to be run in the bash that comes with git, right? can we therefore remove the is_executable crate and do the is_executable check manually using stock rust tools and make sure we do the same file metadata check: https://github.com/fitzgen/is_executable/blob/master/src/lib.rs#L90

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - introduced proper changelog
+- hook support on windows
```
extrawurst, line 10: can you add the reference to the issue (see below changes)?

asyncgit/src/sync/hooks.rs
MCord: done.

asyncgit/src/sync/hooks.rs
MCord: I mistakenly reverted this in a rebase conflict! fixed

CHANGELOG.md
MCord: done

asyncgit/src/sync/hooks.rs
MCord: done.

extrawurst:
Really cool! Thanks for all the effort! ❤️

PR #64, opened 2020-05-17
Better scrolling

### LINKED ISSUE #52: better scrolling in diff/log
now that the diff and log component know their render height, we can make scrolling/selecting more natural by only scrolling once the selection would leave the view instead of always centering the view over the selection.

Stoope:
Fixes #52

### DIFF
--- a/src/components/diff.rs
+++ b/src/components/diff.rs
@@ -140,9 +140,9 @@ impl DiffComponent {
 
     fn get_text(&self, width: u16, height: u16) -> Vec<Text> {
         let selection = self.scroll;
-        let height_d2 = height / 2;
-        let min = self.scroll.saturating_sub(height_d2);
-        let max = min + height;
+        let min =
+            self.current_height * (self.scroll / self.current_height);
+        let max = min + self.current_height;
 
         let mut res = Vec::new();
         let mut line_cursor = 0_u16;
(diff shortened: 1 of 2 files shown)


extrawurst:
Hey @Stoope thanks for your contribution! 

But I think I messed up in my explanation in the ticket. What I meant was that when the selection for example would leave the view on the bottom (of log or diff) the view should scroll down one line to keep it in view. similar to the behaviour people know from tools like `less` and `vim`. Do you know what I mean?

Stoope:
Yup, I get it

akshay1992kalbhor:
Hey @extrawurst, what about scrolling up? In vim for example, the cursor has to reach the top of the screen for the view to scroll up. So if your cursor is at the very bottom of the screen (maybe you are scrolling down), you cannot scroll up until the cursor moves all the way up to the upper edge of the screen. Is this what you had in mind?

akshay1992kalbhor:
@extrawurst Well scrolling down is pretty easy, but scrolling up seems complicated to me.

Stoope:
I'm going to implement some stuct for handling scroll. It will store some addition data about previous scrolling to handle up scroll case. Does it make sense?

extrawurst:
I just added an example of this new behaviour for the log view (https://github.com/extrawurst/gitui/commit/dd49d68eab25248c495ada2ad7577380cbf38484)

I don't see the need for any data structures. please feel free to use the new `calc_scroll_top` for the other scrollable lists aswell.

Screenflick Movie 17

extrawurst:
#52 was closed with 112b5c888cd6bb3f8352f21474f5907bb405b749

PR #65, opened 2020-05-17
[WIP] Implement color themes - closes #28

### LINKED ISSUE #28: Add inverted colors mode for light backgrounds

Currently the main text color is yellow. Which doesn't read well on a white terminal background.

A CLI flag to switch to a dark color theme for better readability.

- I don't know if it's a good idea, but simply inverting all colors might be enough
- Full blown (custom) theme support
- An environment variable or a configuration file are viable alternatives to CLI flags

**Additional context**

image

MCord:
This PR refactors all color related code and creates an style that can be modified to create different color themes for the program. it also provides serialization to and from a config file. changes in this PR

1. created `Theme` struct to represent different colors and some helper functions.
1. replaced all hard coded colors with references to the global `THEME` variable.
1. add `serde` and `ron` crates for serialization. and `lazy_static` for static init.
1. on start the program tries to load `~/.gitui/theme.ron` to load the theme or creates one with default values if not found.

This PR alone at the current state is not enough to provide theming since many components don't specify any style and rely on defaults to draw elements. I have not yet looked into how to change these defaults.


src/components/diff.rs
```diff
@@ -227,16 +216,11 @@ impl DiffComponent {
...
-            .fg(Color::Red)
-            .bg(if selected { select_color } else { Color::Reset });
-        let style_add = Style::default()
-            .fg(Color::Green)
-            .bg(if selected { select_color } else { Color::Reset });
-        let style_header = Style::default()
-            .fg(Color::White)
-            .bg(if selected { select_color } else { Color::Reset })
-            .modifier(Modifier::BOLD);
+        let style_delete = THEME.diff_delete.select(selected);
```
extrawurst, line 219: unfortunately these colours are not the right ones. `diff_delete` as defined now is the color of the entry in the file list in `changes.rs` the lines in the diff view use the more saturated colours `non-light`

src/ui/style.rs
```diff
@@ -0,0 +1,296 @@
...
+            modifier: Modifier::empty(),
+        },
+        text_highlight: Style {
+            fg: Color::Yellow,
+            bg: Color::Reset,
+            modifier: Modifier::empty(),
+        },
+        text_selected: Style {
+            fg: Color::Rgb(0, 0, 100),
+            bg: Color::Magenta,
```
extrawurst, line 99: I believe something was lost in conversion here. `fg` should be the blue and fg simply default `Reset`

Cargo.toml
```diff
@@ -31,8 +31,11 @@ scopeguard = "1.1"
 bitflags = "1.2"
 chrono = "0.4"
 backtrace = { version = "0.3" }
+ron = "0.5.1"
 scopetime = { path = "./scopetime", version = "0.1" }
 asyncgit = { path = "./asyncgit", version = "0.2" }
+serde = "1.0.110"
+lazy_static = "1.4.0"
```
extrawurst, line 38: I am still grinding on this one. I am trying to figure out a way that we can do this without resorting to using a static. Afterall we are not even using altering the `Theme` anymore once it is loaded. It should be a perfect candidate to borrow it everywhere since everyone can only read from it

extrawurst:
Looking at the `theme.ron` I am worried that we went too far here. afterall in most cases we only use one of the colors and no modifier. maybe we can take another angle at it and only load colours and pass a color scheme down to any component that uses styling. this component then could generate the styles on the fly as before but based on the color schema. 

this way me might end up with less content in the `theme.ron` and less todo for anyone getting their hands on it. 

The big benefit for us is aswell that it is easier to reason about what colors we already use in the `scheme`.

What do you think?

Cargo.toml
MCord: yes I don't like that either. I tried passing an instance of `Theme` down the stack but type annotations just exploded everywhere so I decided to keep a private static instance so that I can expose it as `&'static` down the stack and it worked great. the only downside is the reference to `lazy_static`. what do you think ?

MCord:
@extrawurst can you provide an example `theme.ron` file so that I better understand the model you have in mind ?

extrawurst:
pretty sure I missed some:
```
(
    selection_bg: Rgb(0,0,100), // also used for commands background
    selected_tab: Yellow,
    command_disabled: LightGrey,
    diff_line_add: Green,
    diff_line_delete: Red,
    diff_file_added: LightGreen,
    diff_file_removed: LightRed,
    diff_file_moved: LightMagenta,
    commit_hash: Magenta,
    commit_time: Blue,
    commit_author: Green,
    commit_tag: Yellow,
)
```

the trick is then to consider every new entry and reuse any color that is already in there (of course that will drive appropriate naming virtually impossible. But this way looking at gitui someone right aways know: "ok this yellow I see here is bad in my term color scheme, so I change yellow in the file and it is fixed in all places at once"

MCord:
I think having specific names for colors like 'selected_tab' can be confusing once we have used this color in other places but I agree that we don't have enough cases to try to generalize now. I will implement specific case for the moment as you specified.

src/components/diff.rs
MCord: I have completely redone colors. might be better now.

src/ui/style.rs
MCord: fixed

Cargo.toml
MCord: removed lazy static but I am `copy`ing theme instance in a few places to make borrow checker happy.

src/components/diff.rs
```diff
@@ -9,10 +10,10 @@ use asyncgit::{hash, DiffLine, DiffLineType, FileDiff};
 use crossterm::event::Event;
 use std::{borrow::Cow, cmp, convert::TryFrom};
 use strings::commands;
+use tui::style::{Modifier, Style};
```
extrawurst, line 13: this is very nitpick of me, but I don't understand why this guy was moved out of the other collapsed `use`

MCord: side effect of multiple refactorings i guess. fixed

src/app.rs
```diff
@@ -37,24 +39,32 @@ pub struct App {
     revlog: Revlog,
     status_tab: Status,
     queue: Queue,
+    pub theme: Theme,
```
extrawurst, line 42: this can be private

MCord:
fixed

src/spinner.rs
```diff
@@ -16,6 +14,10 @@ impl Spinner {
         self.idx %= SPINNER_CHARS.len();
     }
 
+    pub fn new() -> Self {
```
extrawurst, line 17: I think this whole change can be reverted to using the derive

Cargo.toml
extrawurst: that ok, if we ever see this as an issue we can pass the Theme in some `Rc` or whatever

src/spinner.rs
MCord: done

src/ui/style.rs
```diff
@@ -0,0 +1,275 @@
...
+            };
+
+            default_theme.save(mode).unwrap_or_default();
+            default_theme
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Copy, Clone)]
+enum ColorDef {
```
extrawurst, line 205: why do we have to replicate those?

src/ui/style.rs
```diff
@@ -0,0 +1,275 @@
...
+
+#[derive(Copy, Clone, Debug)]
+pub enum Mode {
+    Dark,
+    Light,
+}
+
+impl Theme {
+    #[cfg(test)]
+    pub const fn empty() -> Self {
```
extrawurst, line 33: lets put this into the test `mod`

src/ui/style.rs
MCord: because serde can not serialize remote types (Color enum is defined in `tui` crate)

src/ui/style.rs
```diff
@@ -0,0 +1,275 @@
...
+
+    fn apply_select(&self, style: Style, selected: bool) -> Style {
+        if selected {
+            style.bg(self.command_background.into())
+        } else {
+            style
+        }
+    }
+
+    pub fn diff_line(
```
extrawurst, line 108: this sort of logic used to be in the component itself. unless we refactor all of that into here we should stay consistent

src/ui/style.rs
```diff
@@ -0,0 +1,275 @@
...
+            Err(std::io::Error::from_raw_os_error(100))
+        }
+    }
+
+    pub fn init(mode: Mode) -> Theme {
+        if let Ok(x) = Theme::read_file(Theme::get_theme_file(mode)) {
+            x
+        } else {
+            let default_theme: Theme = match &mode {
+                Mode::Dark => from_bytes(include_bytes!(
```
extrawurst, line 188: I dont see the benefit of embedding a file into the binary compared to having a const static `THEME_DARK`/`THEME_LIGHT` private to this module and use it here

src/ui/style.rs
MCord: if I do that I need to make `ColorDef` public or maybe I am wrong ?

src/ui/style.rs
extrawurst: ok never mind, this is too minor, it is time to roll this in and sort these things out afterwards

src/ui/style.rs
extrawurst: actually we can get rid of it once we don't embed the themes via file anymore but use a static variable for it

extrawurst:
some styles seem to be lost in translation:

before:
image

after:
image

src/ui/style.rs
extrawurst: I opened an issue for this in `tui-rs`, can you link it in the comment so we can remove this once it is supported: https://github.com/fdehau/tui-rs/issues/292

MCord:
strange, the difference is not visible on windows. I try to reproduce on Linux.

MCord:
i have tried to reproduce it by looking at the diff. can you try again ? (feel free to fix if you know what's wrong)

extrawurst:
looks good. I checked how the default looks on a light terminal:
image
that is not bad as a default actually.

my only remaining concern is that we now rely on an ENV variable to set when one wants to switch the theme. 

what if we only put the dark theme in the binary and dont parse the ENV but document in the readme what to paste into the generated `ron` file?

MCord:
The idea was to include two themes and a way to auto detect which to use given terminals background color but I failed to find a way to auto detect so there is not much value in keeping both unless we find a way to auto detect.

extrawurst:
Yeah I reached out to the `crossterm` guys on discord, there is not good way for auto detecting. I would love to have that. but I think for now this is out of scope of this PR

MCord:
gone !

src/components/help.rs
```diff
@@ -187,7 +199,7 @@ impl HelpComponent {
         {
             txt.push(Text::Styled(
                 Cow::from(format!(" {}\n", key)),
-                Style::default().fg(Color::Black).bg(Color::Gray),
+                Style::default().modifier(Modifier::REVERSED),
```
extrawurst, line 202: nice idea!

src/components/msg.rs
```diff
@@ -2,18 +2,18 @@ use super::{
...
 use tui::{
     backend::Backend,
-    layout::{Alignment, Rect},
-    widgets::{Block, Borders, Clear, Paragraph, Text},
+    layout::Rect,
+    widgets::{Clear, Text},
     Frame,
 };
 
-#[derive(Default)]
```
extrawurst, line 16: can this be `Default` again?

src/spinner.rs
```diff
@@ -2,8 +2,6 @@ use std::io;
 use tui::{backend::Backend, buffer::Cell, Terminal};
 
 static SPINNER_CHARS: &[char] = &['|', '/', '-', '\\'];
-
-///
```
extrawurst, line 6: can be reverted I guess?

src/tabs/revlog/mod.rs
```diff
@@ -299,6 +265,24 @@ impl DrawableComponent for Revlog {
 }
 
 impl Component for Revlog {
+    fn commands(
```
extrawurst, line 268: changed the order? I am confused 🙈

MCord: caused by my editor (Clion) auto fix I think. the order of methods in implementation is recommended to be the same as in  the trait it seems apparently 🤷‍♀️

MCord: fixed

extrawurst:
It is in! Thank you so much for all the effort!🚀

Issue #77, opened 2020-05-20
app error handling using anyhow


extrawurst:
now that we have nice error coverage in `asyncgit` we could take the next step and add error support to the main application using `anyhow` based on the great summary on error handling in rust: 

my idea would be to show an error popup similar to this one where ever possible: https://github.com/extrawurst/gitui/issues/74#issuecomment-631359272

if we come across unrecoverable errors we might need to look into a second path aswell which will close the application.

extrawurst:
one problem is going to be external libraries that do not go full *no-unwrap* too. Currently we have one of those issue with `tui-rs`: #72

MCord:
I am working on this.

PR #84, opened 2020-05-21
#77 anyhow integration

MCord:
This PR contains code refactoring that ensure non-test code does not use `unwrap()` and functions that might fail return a `Result<T>`.   when it was not necessary to forward the errors I used `expect()` instead of `unwrap()` to align with the guidelines. I also took the liberty of denying the use of unwrap by adding clippy annotations to all modules.


src/main.rs
```diff
@@ -132,7 +134,11 @@ fn draw<B: Backend>(
     terminal: &mut Terminal<B>,
     app: &mut App,
 ) -> io::Result<()> {
-    terminal.draw(|mut f| app.draw(&mut f))
+    terminal.draw(|mut f| {
+        if let Err(e) = app.draw(&mut f) {
+            log::warn!("failed to draw: {:?}", e)
```
extrawurst, line 139: * do we really want to keep going with the application when this happens?
* I think this can safely be logged as `error`

src/tabs/stashing.rs
```diff
@@ -187,12 +195,12 @@ impl Component for Stashing {
         }
     }
 
-    fn event(&mut self, ev: crossterm::event::Event) -> bool {
+    fn event(&mut self, ev: crossterm::event::Event) -> Result<bool> {
         if self.visible {
-            let conusmed = self.index.event(ev);
+            let conusmed = self.index.event(ev)?;
```
extrawurst, line 200: I just now saw this typo, but I like that you left it like this  XD

src/ui/style.rs
```diff
@@ -180,44 +182,41 @@ impl Theme {
...
-            })?)
+    fn init_internal() -> Result<Theme> {
+        let file = Theme::get_theme_file()?;
+        if file.exists() {
+            Ok(Theme::read_file(file)?)
         } else {
-            Err(std::io::Error::from_raw_os_error(100))
+            let def = Theme::default();
+            if def.save().is_err() {
+                warn!("failed to store default theme to disk.")
```
extrawurst, line 212: we should make usage of the log functions consistent. the best practice seems to be using free funcs with explicit module like `log::warn`. but I see I also did not do that yet

asyncgit/src/lib.rs
```diff
@@ -53,6 +54,6 @@ pub fn hash<T: Hash + ?Sized>(v: &T) -> u64 {
 pub fn current_tick() -> u64 {
     SystemTime::now()
         .duration_since(UNIX_EPOCH)
-        .unwrap()
+        .expect("failed to fetch time.")
```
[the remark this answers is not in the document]
MCord: fixed

src/main.rs
MCord: changed the log level but not sure if we have to continue or crash here. one of the reasons (I assume) this can fail is the terminal buffer being full for whatever reason, in that case I prefer to continue than to crash the screen.

src/tabs/stashing.rs
MCord: I did not even noticed the typo :)

src/ui/style.rs
MCord: fixed

extrawurst:
Really great stuff! Thank you! 🙏

PR #93, opened 2020-05-27
replaces GITUI_LOGGING env variable by `-l` or `--gitui-logging`

### LINKED ISSUE #88: introduce program arguments
* convert `GITUI_LOGGING` from env to program argument using `clap`
* add version cmd to check version of binary without running for better diagnosis

this is also a prerequisite for #73

shenek:
It uses `clap` crate to do it.

closes #88

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,6 +10,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ansi_term"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


extrawurst:
Hi @shenek thanks for taking this and working out a PR. Hope you don't mind if I rename the long version of the param. I don't think `gitui` needs to be in the param name :)

shenek:
> Hi @shenek thanks for taking this and working out a PR. Hope you don't mind if I rename the long version of the param. I don't think `gitui` needs to be in the param name :)

No problem. Cheers.

PR #94, opened 2020-05-28
chore: Add LICENSE file into the sub-crates

ignatenkobrain:
Those are released separately on crates.io and they do not contain
LICENSE file now that is required when packaging them in Linux
distributions.

Signed-off-by: Igor Raits <i.gnatenko.brain@gmail.com>

### DIFF
--- a/asyncgit/LICENSE.md
+++ b/asyncgit/LICENSE.md
@@ -1,0 +1,1 @@
+../LICENSE.md
--- a/scopetime/LICENSE.md
+++ b/scopetime/LICENSE.md
@@ -1,0 +1,1 @@
+../LICENSE.md


extrawurst:
Good point! Thanks for your contribution @ignatenkobrain but I am curious wouldn't it be better to add them as actual files instead of symlinks? I just realise that crates.io takes a snapshot of the crate on publish and I don't even know what it does when there is a symlink in there: https://docs.rs/crate/asyncgit/0.4.0/source/

ignatenkobrain:
@extrawurst as long as you don't try to publish crate from windows machine, it will be fine (many projects are using it this way).

ignatenkobrain:
it will be converted to the real file since cargo will resolve any symlinks when publishing.

extrawurst:
Perfect! Thanks four your input, I did not know that!

extrawurst:
@ignatenkobrain did this come up in adding `gitui` to any particular distro? I am no linux user, I would love to list it in the README if it is part of any linux distro.

ignatenkobrain:
@extrawurst I have a package under review now in Fedora. Once I build it for stable releases, I'll send update to the README :)

ignatenkobrain:
#95 here you go @extrawurst :)

Thanks for working on this project! I found it quite nice and will open few RFEs tomorrow :)

extrawurst:
@ignatenkobrain what is an RFE? 🙈

ignatenkobrain:
Request For Enhancement :)

Basically I miss opening a commit from commit list that can show diff in some nice view. Aka git show, but better. Also searching capabilities in commits are quite limited now.

PR #95, opened 2020-05-28
README: Add Fedora instructions

ignatenkobrain:
P.S> It will be available only in F32+ and should be pushed there within a few days: https://bodhi.fedoraproject.org/updates/FEDORA-2020-d3770ba5a9


extrawurst:
This is so cool! F32 I guess is a recent Fedora Version? I have now idea how to distribute on linux - so yay! Thank you!

ignatenkobrain:
Fedora 32 Is the latest version of Fedora.

PR #99, opened 2020-05-30
Note AUR package in README

wezm:
I've created [an AUR package](https://aur.archlinux.org/packages/gitui/) for `gitui`. This PR notes it in the README along with the other platforms/distributions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,16 @@ Over the last 2 years my go-to GUI tool for this was [fork](https://git-fork.com
 
 For the time being this product is considered alpha and **not** production ready.
 
+## Arch Linux
+
+There is an [AUR package for `gitui`](https://aur.archlinux.org/packages/gitui/):
+
+```sh
+git clone https://aur.archlinux.org/gitui.git
+cd gitui
+makepkg -si
+```
+
 ## Fedora
 
 ```sh


extrawurst:
Thank you @wezm for doing this. I have no experience with linux packages, what kind of maintenance do they require ? Whats the process when a new `gitui` release lands?

wezm:
> what kind of maintenance do they require

As package maintainer I mostly just have to update the [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=gitui) with the new version and checksum when there's a new release.

> Whats the process when a new gitui release lands?

I subscribe to the [Atom feed](https://github.com/extrawurst/gitui/releases.atom) for new releases in my feed reader and have it set up to notify me on my phone of new entries in the feed. So I will notice any new releases and update the AUR package.

I've actually had this in place for a [number of releases](https://aur.archlinux.org/cgit/aur.git/log/?h=gitui) already.

extrawurst:
Pretty cool @wezm! 
Is there a way to automate this?

wezm:
In theory yes but it’s typically done manually so that a human is involved in checking that the new package works, any newly added or removed dependencies are tracked in the PKGBUILD etc.

extrawurst:
I see! wow. Today I learned.
Thanks for doing this❤️

extrawurst:
@wezm one question that came up:
What does this conflict line mean?
```
Conflicts: gitui-git
```

See https://aur.archlinux.org/packages/gitui/

wezm:
Sometimes people create a package for the master version of software in addition to the released version. In that case the convention is to name the package with a -git suffix. Since one of these packages and a package for the release version would likely contain very similar files they can’t coexist since they will try to install files into the Sam paths. The conflicts line says that if someone were to install a hypothetical -git package for gitui it would conflict with this package. In that case pacman will prompt the user and offer to remove this package so the -git one can be installed. Basically it’s just a good practice even though such a package doesn’t, or may never exist. 

Hope that makes sense.

PR #104, opened 2020-06-02
add --directory option (closes #73)

### LINKED ISSUE #73: Open gitui in a given directory (gitui --cd=xyzdir/)

I'd like to set the tool up so that when I right click on a directory (in windows) it launches gitui in the clicked directory. To do this I'd need a way to open the program in a different directory.

`gitui.exe --cd=projectdir` to open the tool in the directory called projectdir

The context menu works fine from inside the directory, but with git bash I'm used to clicking on a directory to open that tool, so it would be nice if this tool mirrored that behaviour. This is a low priority thing.

**Additional context**

image

alistaircarscadden:


extrawurst:
Hey @alistaircarscadden thanks for looking into this. I love the simplicity of the solution. It did not even cross my mind to do it like that. I was expecting to have to search and change very place I am using `CWD`.

Does it have any downsides to switch the current working directory in your opinion?

alistaircarscadden:
@extrawurst 
From my systems experience this is the usual way to do this. I think a shell actually does the same thing when you start a process, between fork(2) and exec(3), using chdir(2).

The notable effects are:
- the gitui process will create child processes in it's working directory, which would (after this change) be wherever the user set it with -d.
- relative paths in the process (`../`, `./somedir`) are relative to the working directory, so they'd be relative to the new working directory too.
- the change here does not effect the parent shell of gitui, so when you quit gitui you'd be back in the original directory.

I think these are all behaviours we want gitui to have.

extrawurst:
Perfect thanks for clarifying! Any shorter alternative to `directory`? maybe `path`,`folder`? What do you think?

Can you also add this to the changelog?

jugglerchris:
The common short option (used by make and git) would be -C.  Make has --directory as a long option. 

I quite like --in but I don't know of a precedent. :-)

alistaircarscadden:
@extrawurst 
> Can you also add this to the changelog?

Yes.

> Any shorter alternative to directory

`--dir` ?

@jugglerchris 
> would be -C

At first glance -C/--directory is a bit dissonant to me, but I'm up for whatever.

extrawurst:
> Yes.

Cool!

> At first glance -C/--directory is a bit dissonant to me, but I'm up for whatever.

I agree.
Lets stick to `-d` and `--directory` 👍

alistaircarscadden:
> Can you also add this to the changelog?

Finished, now.

extrawurst:
Thanks so much!❤️

Issue #26, opened 2020-04-21
revert workaround for crossterm key modifier issue


extrawurst:
revert 72a1f94483283a022b64c59ecbafaa00070793c8 once Crossterm issue is fixed: https://github.com/crossterm-rs/crossterm/issues/421

PR #110, opened 2020-06-04
fixed an issue with tab reverse on windows.

MCord:
This PR fixes a small but annoying bug on windows that prevents `SHIFT + TAB` to select the tabs in revers order. The bug is caused by the difference in interpretation of modifiers either in `crossterm` or lower levels. 

for non windows platforms the `SHIFT + TAB` key generates `KeyCode::BackTab` without any modifiers while on windows the same produces `KeyCode::BackTab` + Shift modifier. 

I am not sure the issue should be addressed in `gitui` but here is the fix.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -129,7 +129,8 @@ impl App {
                     self.toggle_tabs(false)?;
                     NeedsUpdate::COMMANDS
                 }
-                keys::TAB_TOGGLE_REVERSE => {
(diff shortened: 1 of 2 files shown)


extrawurst:
Oh damn, we had such an issue before: #26
Can you do me a favour and tackle this upstream like I did in the previous one, the guys are more than happy to accept such fixes in crossterm.

MCord:
Sure, I will look into it.

Issue #46, opened 2020-05-09
better commit msg support


extrawurst:
* [x] allow moving cursor (see #117)
* [ ] show chars used (>50 red)
* [x] support delegating commit msg editing to a terminal editor for longer messages

extrawurst:
Thanks @alistaircarscadden for implementing the cursor support:
![Screenflick Movie 32](https://user-images.githubusercontent.com/776816/84589257-aac4b180-ae2d-11ea-88a4-c02ccf808ef6.gif)

extrawurst:
see 477fdd303d0c4e3c533cc49fb715506588196eb4

extrawurst:
@jonstodle thanks again, the external editor feature is released in `0.8` now: https://github.com/extrawurst/gitui/releases/tag/v0.8.0

PR #114, opened 2020-06-09
Multiline commit message

jonstodle:
This intends to implement a way to enter multi-line commit messages. Ref #46 

This adds support for opening a file in an external editor, writing the commit message and saving it to the file, and the use the entered message as the commit message. This works in much the same way git does by default.

This PR at, in it's current state is not ready for merging. There are a few issues which will need to be resolved first and I hope to get some feedback on how to solve these. They are as follows:

## gitui will capture key stroke events and prevent them from being passed on to the editor

> This is only an issue when the editor is opened inside the terminal (such as _vim_), and not as an external GUI application (such as _VS Code_).

The background thread spawned by gitui will continue to run and capture key strokes while the external editor is in use. I see mainly three ways of solving this:

1. Add a way to make the background thread skip polling. This is how it's currently solved, but it's not working completely. When launching the editor, the "poller" will run another time, poll for 2 seconds and then stop polling for as long as the editor is open. If you type inside of those 2 first seconds, the key stroke is going to be captured by gitui and will not be passed on to the editor.
1. Lower the polling duration from 2 seconds. It's not desirable, but it could work.
1. Add a way to kill the spawned thread when launching the editor and then spawning it again once the editor exits.

## gitui needs to redraw everything when returning from editor

Because both gitui and vim uses the _alternative screen_ feature, gitui will need to redraw it's whole UI when returning from the editor as the incremental updates tui usually does means some parts of the UI never gets set to it's correct state. If a part of gitui never changes, the "regular" shell underneath will "show through".

It seems the only way to make tui redraw the whole screen is to both clear the current screen, but also reset the stored previous buffer. Resetting all the stored buffers is something tui doesn't allow consumers to do (as far as I can see), except for one instance: `resize`. This function will reset all the internal buffers and force a redraw of the whole screen. `resize` is currently used to enforce a redraw. (Might be worth considering opening an issue upstream to enable forcing a complete redraw of the UI).

A big downside with current implementation of this is that it introduces a mutable static variable, which means it also introduces `unsafe` (the _forbid unsafe_ directive has been commented out). This might be worked around by using `lazy_static`, but I haven't investigated that in any way.

The reason for the mutable static is the introduction of a new channel, global to the app, which let's different parts of the app send command requests to the main loop. In this case it's a request for a full redraw of the UI, but this could be useful for other commands later (but with a better implementation of how to share the _sender_).

---

This PR is not based on a very recent commit because _master_ would not build on stable for me, so I kept it at 0.5.0-ish. I will rebase it at a later point.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,3 +1,4 @@
+use crate::input::InputState;
 use crate::{
     accessors,
     cmdbar::CommandBar,
@@ -18,6 +19,7 @@ use anyhow::{anyhow, Result};
 use asyncgit::{sync, AsyncNotification, CWD};
 use crossbeam_channel::Sender;
 use crossterm::event::{Event, KeyEvent};
+use std::cell::Cell;
 use std::{cell::RefCell, rc::Rc};
 use strings::{commands, order};
 use tui::{
@@ -28,6 +30,7 @@ use tui::{
     widgets::{Block, Borders, Tabs},
     Frame,
 };
+
 ///
 pub struct App {
     do_quit: bool,
@@ -45,6 +48,10 @@ pub struct App {
     stashlist_tab: StashList,
     queue: Queue,
     theme: SharedTheme,
+
+    // "Flags"
+    requires_redraw: Cell<bool>,
+    set_polling: bool,
 }
 
 // public interface
@@ -85,6 +92,8 @@ impl App {
             stashlist_tab: StashList::new(&queue, theme.clone()),
             queue,
             theme,
+            requires_redraw: Cell::new(false),
+            set_polling: true,
         }
     }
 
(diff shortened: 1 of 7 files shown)


jonstodle:
The summary of this PR pretty much boils down to the commit message of the commit:

> Adds support for editing commit messages in an external editor. It read the GIT_EDITOR, VISUAL, EDITOR environment variables in turn (in the same order git does natively) and tries to launch the specified editor. If no editor is found, it falls back to "vi" (same as git). If vi is not available, it will fail with a message.

extrawurst:
wooohhooooooo!! Thank you @jonstodle 🙏

extrawurst:
closes #46

PR #117, opened 2020-06-12
Add cursor to TextInputComponent for better commit message support (#46)

alistaircarscadden:
This is one of the changes for #46 .

I've added a movable cursor to the TextInputComponent that is used to enter commit messages. This allows the user to use the arrow keys to move the cursor left and right, insert characters at the cursor. The cursor behaves naturally, moving with the text.

This change does not address the part of #46 that says characters after 50 should be red, it is very possible to make that edit to the code but will make the branching quite messy, and I will consider for a while how to best handle that logic.

This change does not address the part about multi line commit messages, and opening another editor.

I tested my changes on the commit window, I'm not sure what other functions use that TextInputComponent, and they have not been tested.


alistaircarscadden:
Example of the cursor as you type a commit message:

image

Example of inserting text in to the middle of your selection:

image

I've just thought of an edge case I somehow missed, backspacing with the cursor in the middle. Committing that fix soon.

a.file.txt
```diff
@@ -0,0 +1 @@
+asdasd
```
extrawurst, line 1: what's up with this? :)

src/components/textinput.rs
```diff
@@ -51,6 +52,23 @@ impl TextInputComponent {
...
+    fn incr_cursor(&mut self, amt: usize) {
+        self.cursor_position =
+            if self.cursor_position + amt > self.msg.len() {
+                self.msg.len()
+            } else {
+                self.cursor_position + amt
+            }
+    }
+
+    fn decr_cursor(&mut self, amt: usize) {
```
extrawurst, line 65: this can be a one-liner using `saturating_sub`

src/components/textinput.rs
```diff
@@ -110,13 +158,25 @@ impl Component for TextInputComponent {
...
-                        self.msg.pop();
+                        if 0 < self.cursor_position
+                            && self.cursor_position <= self.msg.len()
+                        {
+                            self.msg.remove(self.cursor_position - 1);
+                        }
+                        self.decr_cursor(1);
                         return Ok(true);
                     }
+                    KeyCode::Left => {
```
extrawurst, line 174: KeyEvents that are consumed should return `return Ok(true);`

extrawurst:
Cool @alistaircarscadden thanks for tackling this! I just had a few minor comments, would be cool if u could look into that and then it will be merged!

Also please add a changelog entry

src/components/textinput.rs
```diff
@@ -60,16 +78,46 @@ impl DrawableComponent for TextInputComponent {
...
-                    Cow::from(self.msg.clone()),
-                    Style::default(),
-                )]
+                let len = self.msg.len();
+
+                // the portion of the text before the cursor is added
+                // if the cursor is not at the first character
+                if self.cursor_position > 0 {
+                    txt.push(Text::styled(
+                        &self.msg[..self.cursor_position],
```
extrawurst, line 94: these indexing panics when the message contains unicode like `ä` and the index tries to split in between a multibyte unicode character

src/components/textinput.rs
```diff
@@ -110,13 +158,25 @@ impl Component for TextInputComponent {
                         return Ok(true);
                     }
                     KeyCode::Char(c) => {
-                        self.msg.push(c);
+                        self.msg.insert(self.cursor_position, c);
+                        self.incr_cursor(1);
                         return Ok(true);
                     }
                     KeyCode::Backspace => {
```
extrawurst, line 165: Would also cool to support `KeyCode::Delete`

extrawurst:
I gave it another round of review and the biggest issue is that right now unicode is not handled and crashes

a.file.txt
alistaircarscadden: Haha, oops. Guess I wanted something to test committing.

src/components/textinput.rs
alistaircarscadden: Thanks for pointing this function out, it's a good one.

alistaircarscadden:
@extrawurst Ok, fixed the behaviour when accents etc are used. The new operations walk thru the string using the &str.chars() and &str.char_indices() functions to get the next char boundary. In order to move the cursor backwards I'm walking thru the string until the previous position, which since these strings are very short takes no time at all, but if this TextInputComponent was used for some 20,000 word essay we'd want to be caching the values or something! 😀

I didn't add the changelog item because I'm currently working on a previous release. I merged the master into my branch, but can't compile anymore due to
```rust
error[E0723]: heap allocations are not allowed in const fn
   --> src\components\help.rs:163:19
    |
163 |             cmds: vec![],
    |                   ^^^^^^
    |
    = note: see issue #57563 <https://github.com/rust-lang/rust/issues/57563> for more information
    = help: add `#![feature(const_fn)]` to the crate attributes to enable
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)```

extrawurst:
looks like you are not using the latest rust (1.44) then this should be fine.

extrawurst:
Awesome stuff man! Thanks for this ❤️

extrawurst:
@alistaircarscadden I found `is_char_boundary` to be what we are looking for to remove the O(n) loop to find the previous index: https://doc.rust-lang.org/std/primitive.str.html#method.is_char_boundary

alistaircarscadden:
> I found `is_char_boundary`

Oh I actually saw that, and for some reason it didn't click. That would be super easy to use for this.

I'll update my rust, thanks.

PR #123, opened 2020-06-14
wrap TextInputComponent text

alistaircarscadden:
image


extrawurst:
Thanks! I wonder how I missed that for so long :)

PR #126, opened 2020-06-14
Move config directory

### LINKED ISSUE #98: Suggest moving theme.ron from XDG_CACHE_HOME to XDG_CONFIG_HOME
Related to #27

[The spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) notes:

> $XDG_CACHE_HOME defines the base directory relative to which user specific non-essential data files should be stored.

The non-essential bit there means this data can be purged at any point without impacting the user. `theme.ron` being user data probably should be preserved, which suggests it should be in `"$XDG_CONFIG_HOME/gitui/theme.ron"`

jonstodle:
This changes the config directory from the `dirs::cache_dir` to `dirs::config_dir`. It does however keep the cache directory as the default logging directory, as it seems a better fit.

It also adds a function, `migrate_config`, which is called at startup to move directory entries inside the "old" config directory to the "new" one (but it skips moving log files). The intention is that this function can be removed after a few releases when the likelihood of someone upgrading from 0.6.0 or earlier is fairly small.

Fixes #98

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -66,6 +66,9 @@ fn main() -> Result<()> {
         return Ok(());
     }
 
+    // TODO: To be removed in a future version, when upgrading from 0.6.x or earlier is unlikely
+    migrate_config()?;
+
     setup_terminal()?;
     defer! {
         shutdown_terminal().expect("shutdown failed");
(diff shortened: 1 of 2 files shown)


extrawurst:
Really cool! Thanks for tackling this!❤️

src/main.rs
```diff
@@ -202,7 +205,26 @@ fn start_terminal<W: Write>(
...
+
+    let entries = cache_path
+        .read_dir()?
+        .flat_map(|dir_entry| dir_entry)
+        .filter(|entry| {
+            !entry.file_name().to_string_lossy().ends_with(".log")
+        });
+
+    for entry in entries {
+        let mut config_path: PathBuf = get_app_config_path()?;
```
extrawurst, line 219: one minor thing: we can move out the call to `get_app_config_path` outside the loop and keep using a clone of it. just pushed that: c6eb066

PR #136, opened 2020-06-16
update README.md

alistaircarscadden:
Changed a lot of text in the README for consistency, correctness, readability, etc.

Notably:

- a handful of capitalization 😀
- transposed this table and added emojis to highlight the best values as well as the original bolding
image
- fixed some spelling errors
- changed many sentence structures to flow more naturally
- moved the order of the motivation section to prioritize it over limitations
- combined Limitations and Expectations into what I feel is more readable (sort of a "Here are the limitations and here is why" format)
- cleaned up some links and added consistency to the "See xyz" segments
- put the word "Download" in the readme so people can Ctrl+F to it


extrawurst:
@alistaircarscadden cool thanks!!

PR #138, opened 2020-06-16
Add CD action to update Homebrew formula

vladimyr:
Uses https://github.com/mislav/bump-homebrew-formula-action for updating Homebrew core formula.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-  
+
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v2
@@ -39,7 +39,7 @@ jobs:
       run: |
         rustup target add x86_64-unknown-linux-musl
         sudo apt-get -qq install musl-tools
-      
+
     - name: Build Release Mac
       if: matrix.os == 'macos-latest'
       run: make release-mac
@@ -71,9 +71,9 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-    - name: Bump Brew
+    - name: Bump extrawurst/homebrew-tap formula
       if: matrix.os == 'macos-latest'
-      env: 
+      env:
         HOMEBREW_GITHUB_API_TOKEN: ${{ secrets.BREW_TOKEN }}
       run: |
         brew tap extrawurst/tap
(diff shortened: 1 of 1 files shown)


extrawurst:
Wow this is magic, I am wondering, is building it from source then? how does it know how to build it? I am such a brew-noob :(

vladimyr:
Yeah it really looks magical but it is really just a large pile of tools/abstractions 😁 

> I am wondering, is building it from source then?

That is exactly how it works. `brew` fetches source tarball and executes `cargo install`:
https://github.com/Homebrew/homebrew-core/blob/616af97/Formula/gitui.rb#L16-L18 All that happens on Homebrew CI servers and resulting binary gets saved or as they say, [bottled](https://github.com/Homebrew/homebrew-core/blob/616af97/Formula/gitui.rb#L9-L11) so end users just download and unpack build matching their platform.
If you ever used (BSD) ports it is pretty much the same pattern except compiling is almost always done on a remote machine and users get precompiled binaries.

> how does it know how to build it?

Well, it just executes build step from the `install` method. The formula is pretty much nothing more than a runnable Ruby class. And this `cargo install` line is Homebrew's convention available as a template: https://github.com/Homebrew/brew/blob/6b36830/Library/Homebrew/formula_creator.rb#L177-L179 You can specify it by using `brew create --rust ...`.

> I am such a brew-noob :(

You should see me doing my first formula. Clueless is a mild word for describing that disaster 😁 
I still do not consider myself a Homebrew expert but I learned a trick or two and if you have any additional questions feel free to ask and I hopefully help you find the answers.

extrawurst:
Ok thanks for the clarifications. Any idea how to test this? Last time I played around with CD changes it was a mess cause I had to tag multiple releases just because stuff went wrong. People got annoyed because of the spamming this produced 🙈

vladimyr:
I'm afraid I can only point you to create another one 🙈

vladimyr:
FWIW, this should be as good as Github CLI's CD workflow: https://github.com/cli/cli/blob/7b225bf1c9ce8cb1ba5c1e7c422a91da743acbc4/.github/workflows/releases.yml#L30-L36

extrawurst:
@vladimyr ok I am getting closer to a new release, so I will pretty soon merge this. one thing I would like to do is seep the 'old'  private tap formulae update code around at first. could you adjust this?

vladimyr:
Sure, I'll add that. 👍

vladimyr:
@extrawurst I guess that's it but please double-check token scopes: [docs](https://github.com/mislav/bump-homebrew-formula-action#readme:~:text=You%20should%20enable%20GITHUB_TOKEN%20only%20if,the%20downloadable%20archive%20for%20this%20release.)

extrawurst:
@vladimyr thanks for looking into this, but I meant the old CI step, not updating the private formula with the new action. I would like to keep the system that worked before inlace and just `add` the new action that *hopefully* works for the core-tap

vladimyr:
Is this what you want? 🙃

extrawurst:
@vladimyr really cool! thank you! ❤️

PR #139, opened 2020-06-17
Do not mix number and letters here

matthiasbeyer:
Fixes the README to not use a mix of numbers and letters,
but either "Ninehundretthousand" or "900,000" here,
with this patch the latter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@
 
 # Benchmarks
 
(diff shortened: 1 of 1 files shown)


extrawurst:
hm problem with this is that it does not localise the same globally. in Germany this would read 900 with some zeros after the decimimal. maybe `900k` then?

matthiasbeyer:
The rest of the README wouldn't localize either, because it is in english. IMO it is safe to assume that readers from other countries are familiar with how numbers are written in english.

matthiasbeyer:
If you think so...

matthiasbeyer:
Of course not! You're in charge here and I do nothing more than propose something! Having the patch possibly rejected is part of the deal, right? :smile:

PR #154, opened 2020-06-26
Stash commit details should include untracked files (fixes #130)

### LINKED ISSUE #130: untracked files in stash diff not shown
unfortunately git stores the untracked files of a stash in the stash's [third parent](https://stackoverflow.com/a/51276389/2458533) so we need some super specific process to make sure the `inspect` works correctly here.

fun fact, the same applies to fork:
Screenshot 2020-06-15 at 21 08 44

MCord:
Stash commits treat untracked files specially in a parent commit of the actual stash commit. this causes the details of the commit to not include these untracked files.

The fix is to treat commit messages differently when showing the details of a commit message. here we detect that the shown commit is actually a commit messages and add any files inside the third parent to the list of files included in the commit. I have also modified the diff function to calculate the diff from the correct commit when calculating the diff on an untracked file inside a stash commit.

### DIFF
--- a/asyncgit/src/sync/commit_files.rs
+++ b/asyncgit/src/sync/commit_files.rs
@@ -1,5 +1,5 @@
-use super::{utils::repo, CommitId};
-use crate::{error::Result, StatusItem, StatusItemType};
+use super::{stash::is_stash_commit, utils::repo, CommitId};
+use crate::{error::Result, StatusItem, StatusItemType, CWD};
 use git2::{Diff, DiffDelta, DiffOptions, Repository};
 use scopetime::scope_time;
 
@@ -33,6 +33,22 @@ pub fn get_commit_files(
         None,
     )?;
 
+    // stash commits have parent commits containing untracked files and if we want to show
+    // these files as if they were actually in the stash commit  we have to have some specific
+    // handling regarding these special stash commits.
+    // more info can be found at https://stackoverflow.com/questions/51275777/why-does-git-stash-pop-say-that-it-could-not-restore-untracked-files-from-stash/51276389#51276389
+    if is_stash_commit(repo_path, &id)? {
+        let commit = repo.find_commit(id.into())?;
+        let untracked_commit = commit.parent_id(2)?;
+
+        let mut untracked_files = get_commit_files(
+            repo_path,
+            CommitId::new(untracked_commit),
+        )?;
+
+        res.append(&mut untracked_files);
+    }
+
     Ok(res)
 }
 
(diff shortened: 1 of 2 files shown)


asyncgit/src/sync/commit_files.rs
```diff
@@ -65,6 +81,16 @@ pub(crate) fn get_commit_diff(
         opt.as_mut(),
     )?;
 
+    if diff.deltas().len() == 0 && is_stash_commit(CWD, &id)? {
```
extrawurst, line 84: sure `CWD` is what we want here? will this work when running gitui inside a sub dir of the repo?

asyncgit/src/sync/commit_files.rs
```diff
@@ -52,7 +68,7 @@ pub(crate) fn get_commit_diff(
         None
     };
 
-    let mut opt = pathspec.map(|p| {
+    let mut opt = pathspec.clone().map(|p| {
```
extrawurst, line 71: are two new `clone`s (this and the one below) really necessary🤔 can we `as_ref()` before mapping and maybe only clone inside?

asyncgit/src/sync/commit_files.rs
```diff
@@ -74,40 +100,39 @@ mod tests {
     use crate::{
         error::Result,
         sync::{
-            commit, stage_add_file, stash_save, tests::repo_init,
-            CommitId,
+            commit, stage_add_file, stash, stash_save,
+            tests::repo_init, CommitId,
         },
-        StatusItemType,
+        StatusItemType, CWD,
```
extrawurst, line 106: where is `CWD` used here?

asyncgit/src/sync/commit_files.rs
MCord: This still works even if we are in a subfolder since we are still in the same git repo.

extrawurst: here we are already inside a method that made the lookup to get a `&Repository` so either:
1) we should make use of it here and make `is_stash_commit` crate private
2) or we leave it public and need to feed it the `asyngit::sync::work_dir()` of `repo`

otherwise unittests will fail using it, because they do not switch the `cwd`

asyncgit/src/sync/commit_files.rs
MCord: fixed

asyncgit/src/sync/commit_files.rs
MCord: fixed

asyncgit/src/sync/commit_files.rs
extrawurst: @MCord there is still a unnecessary `clone` which can be `as_ref`

PR #173, opened 2020-07-06
Add command to edit selected file in editor

### LINKED ISSUE #166: open file in $EDITOR from tigui
When you watching at diffs you sometimes may want to edit file just a bit, and for that in `tig` there is `e` key to open selected file in `$EDITOR`, would be super cool to have same functionality here as well

jonstodle:
Pressing `e` while looking at a file in the _Status_ view will launch an external editor with the current file opened. The editor chosen is determined by the default logic introduced in #114.

An improvement to this in the future could be launching at the specific line at which the _Diff_ view is focused, but that seems to require a change in `FileDiff` which is a change bigger than this PR.

Fixes #166

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -18,6 +18,7 @@ use anyhow::{anyhow, Result};
 use asyncgit::{sync, AsyncNotification, CWD};
 use crossbeam_channel::Sender;
 use crossterm::event::{Event, KeyEvent};
+use std::path::PathBuf;
 use std::{cell::Cell, cell::RefCell, rc::Rc};
 use tui::{
     backend::Backend,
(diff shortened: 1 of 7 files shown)


extrawurst:
@jonstodle good work! ❤️

beet:
This is awesome, thanks!

Looks like on Mac it opens `~/Library/Application Support/gitui/COMMITMSG_EDITOR` in VIM to pre-populate the commit message with some boilerplate text, so I was able to get word wrap at 80 characters working by adding this to my .vimrc:

```
filetype plugin on
au FileType gitcommit set textwidth=80 wrap
autocmd BufNewFile,BufRead COMMITMSG_EDITOR set filetype=gitcommit
```

PR #187, opened 2020-07-08
Add command for tagging commit

### LINKED ISSUE #103: tag commit
In the commit log, I want to be able to tag the selected commit.

* [ ] new command
* [ ] text input popup

cruessler:
Opened as a draft for getting early feedback.

Open questions:

- [x] Do you want to validate the tag before calling `repo.tag` or do you want
  to rely on git refusing invalid tags?
- [x] Should the dialog have just a single line?

Will close #103.

### DIFF
--- a/asyncgit/src/sync/commit.rs
+++ b/asyncgit/src/sync/commit.rs
@@ -1,6 +1,6 @@
 use super::{get_head, utils::repo, CommitId};
 use crate::error::Result;
-use git2::{ErrorCode, Repository, Signature};
+use git2::{ErrorCode, ObjectType, Repository, Signature};
 use scopetime::scope_time;
 
 ///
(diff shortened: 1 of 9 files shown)


extrawurst:
@cruessler thanks for tackling this. 

> Do you want to validate the tag before calling repo.tag or do you want
to rely on git refusing invalid tags?
I would rely on git refusing an invalid tag. we have a reusable error message box for that already.

> Should the dialog have just a single line?

Not exactly sure what you mean? I would expect the same dialog we already use for the stash-name.

asyncgit/src/sync/commit.rs
```diff
@@ -80,6 +80,23 @@ pub fn commit(repo_path: &str, msg: &str) -> Result<CommitId> {
         .into())
 }
 
+pub fn tag(
```
extrawurst, line 83: Please add a doc comment here and also a unites

src/tabs/revlog.rs
```diff
@@ -28,6 +28,7 @@ const SLICE_SIZE: usize = 1200;
 ///
 pub struct Revlog {
     commit_details: CommitDetailsComponent,
+    tag_input: TextInputComponent,
```
extrawurst, line 31: directly integrating this into the revlog is not the right approach. please check the `StashMsgComponent` that is register in `App` as a Popup.

cruessler:
@extrawurst Thanks for the feedback! I have incorporated it and am now able to locally tag commits. The revlog does not update, though, after a commit is tagged and I have found no way to change that. `InternalEvent::Update(NeedsUpdate::ALL)` is pushed to the queue after the commit is tagged. Am I missing something?

extrawurst:
@cruessler I could help you more when you would push those changes you made :)

extrawurst:
@cruessler as of #195 new tags will be refetched every 2s.

when you add a tag you can also *force* a refetch (because we know a new one is gonna be there) in `Revlog`:
```
self.git_tags.request(Duration::from_secs(0), true)?;
```

cruessler:
@extrawurst I updated the branch with your feedback. I did not need to explicitly ask the `Revlog` to re-load the tags, the showed up immediately.

extrawurst:
@cruessler nice work! thank you! ❤️

PR #197, opened 2020-07-13
Greater 50 chars commit highlighting

ciaranmooney:
This PR adds a red highlight to characters that exceed the limit of 50 in a commit message (see second point of #46).

image

In order to accomplish this I have renamed the `get_draw_text` function to `apply_cursor` as this seemed more apt and allowed me to add in the `apply_highlighting` function.

Along the way I corrected some minor typographical errors in comments.


src/components/textinput.rs
```diff
@@ -151,6 +152,25 @@ impl TextInputComponent {
 
         txt
     }
+
+    fn apply_highlighting(&self) -> Vec<Text> {
+        let style = self.theme.text(true, false);
+        let mut txt = Vec::new();
+        let char_limit_warning = 50;
```
extrawurst, line 159: for constants lets put them up on top outside the type

src/components/textinput.rs
```diff
@@ -151,6 +152,25 @@ impl TextInputComponent {
...
+        if self.msg.len() > char_limit_warning {
+            txt.push(Text::styled(
+                &self.msg[..char_limit_warning],
```
extrawurst, line 162: this will likely fail if the limit cuts in the middle of a utf8 boundary.

src/components/textinput.rs
```diff
@@ -151,6 +152,25 @@ impl TextInputComponent {
...
+                style,
+            ));
+            txt.push(Text::styled(
+                &self.msg[char_limit_warning..],
+                style.fg(Red),
```
extrawurst, line 167: colors should be looked up using the theme so people can customise it. in the style we have `text_danger` already which likely is what u are looking for

src/components/textinput.rs
```diff
@@ -166,7 +186,8 @@ impl DrawableComponent for TextInputComponent {
                     self.theme.text(false, false),
                 )]
             } else {
-                self.get_draw_text()
+                self.apply_cursor();
```
extrawurst, line 189: how come we don't need to use the result of `apply_cursor`?

src/components/textinput.rs
```diff
@@ -104,13 +105,13 @@ impl TextInputComponent {
         self.title = t;
     }
 
-    fn get_draw_text(&self) -> Vec<Text> {
```
extrawurst, line 107: I think `get_draw_text` is more appropriate for what we want the method to do, if we wanna split the logic it does because it would be too much for one method lets subdivide it. other components already use a similar notation of moving this logic into a sort of `get_text` method.

ciaranmooney:
Hi, Thanks for taking the time to review this PR.  I must admit this was the first bit of Rust I have written. I was overcome with my own inflated sense of achievement that I decided to offer it up. 

- [ ] Move constant 
- [ ] Handle UTF8 
- [ ] Extensible/Generalised solution
- [ ] Keep old method name (`get_draw_text`)

src/components/textinput.rs
ciaranmooney: Didn't even consider utf8 - embarrassing considering I have one (á) in my name.

extrawurst: no worries, happens to me all the time 🙈

extrawurst:
> Extensible/Generalised solution

lets just worry about the use cases we have: commit message (limit 50), stash name (no limit), upcoming tag title (no limit), so I guess a `limit: Option<u8>` would suffice and can be defined at compile time in the `new`.

I am just saying cause those words could mean many more things otherwise :)

extrawurst:
@ciaranmooney are you still planning on pursuing this?

ciaranmooney:
Hi, Yes. I have recently come back to it. I'm happy to work on it privately on a my own repo for the time being if you wish to close this PR down.

ciaranmooney:
After making 3 or 4 attempts at implementing highlighting properly I'm going to have to admit defeat, for now.  It seems that adding styling based on two separate conditions (cursor and character limit) is not quite as simple as I'd like it to be.

I can't seem to find a way that does not end in a untidy mess of if-statements and magnitude checks (ie limit < length, limit < cursor_position etc). I believe that this component may need some refactoring. Unless there is another area of the application which handles a similar scenario?

I have extended the unit tests to cover existing behaviour so I will endeavour to get them merged in a separate PR.

extrawurst:
closed in favour of https://github.com/extrawurst/gitui/pull/264

PR #208, opened 2020-07-20
Add scrolling to long commit messages

cruessler:
- [x] Manually wrap commit messages
- [x] Add commands for navigation: focus commit message
- [x] Add scrolling to component
- [x] Add tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -171,7 +171,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bdfa80d47f954d53a35a64987ca1422f495b8d6483c0fe9f7117b36c2a792129"
 dependencies = [
  "bitflags",
- "textwrap",
+ "textwrap 0.11.0",
  "unicode-width",
 ]
 
@@ -422,6 +422,7 @@ dependencies = [
  "scopetime",
  "serde",
  "simplelog",
+ "textwrap 0.12.1",
  "tui",
  "unicode-width",
 ]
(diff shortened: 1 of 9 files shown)


extrawurst:
@cruessler the windows build was broken due to #210 - once you rebase onto current master CI should be fine again

cruessler:
@extrawurst I updated the PR. Scrolling is not yet included, because I first wanted to ask whether the PR goes into the right direction. If so, adding the remaining parts should hopefully not be too difficult.

extrawurst:
@cruessler from a first glance this looks like the way to go 👍

some nitpicks: let's not keep highlighting the top entry in the tree while the commit message is focused. I am also not sure about highlighting a line in the commit message🤔

cruessler:
@extrawurst I have incorporated your feedback, so the PR should be near feature-complete. The only thing missing are a few tests.

extrawurst:
@cruessler really cool, a few tests would be good 👍

cruessler:
@extrawurst I just added a few tests. Let me know if you want more!

extrawurst:
unfortunately it looks not right to me, something is off here:
Screenflick Movie 56

please checkout the amount of up arrows needed until it actually starts scrolling up and how it takes many down-arrow before it starts scrolling down again.

also the clip shows how the entire text is bold instead of only the first line (like it was before).

Can you look into that? You can reproduce it with 1c3295a in the `gitui` repo.

cruessler:
@extrawurst Thanks for the feedback! My first solution was needlessly complicated because it tracked `selection` and `scroll_top` separately. I changed it so that it now only uses `scroll_top` which fixes the issues.

My terminal was not configured to show bold text, so I did not notice the other issue before. I fixed that, too.

extrawurst:
@cruessler really cool! thanks for having pulled through with this! great job!❤️

PR #216, opened 2020-07-29
Switch focus when working directory or stage becomes empty

### LINKED ISSUE #215: Move to (un)staged when the current selection is empty
When using the _Stage all_ command (or just when all items have been (un)staged. The focus should switch to the other view. There's isn't really a reason for the focus to remain in _Unstaged changes_, when all the files have been moved to _Staged changes_ (and vice versa).

When using _Stage all_ or when all files have been moved from or to the "stage", the focus should change to the relevant area where there are files present.

None

**Additional context**
...

jonstodle:
If the focus is on the working directory and it becomes empty (no files are listed) by either staging the last file or using the _Stage all_ command, the focus will be moved to the stage automatically. It works the same for when the stage becomes empty, by moving the focus to the working directory.

If the focus is on the diff area, the focus will not be moved when either the working directory or the stage becomes empty.

Fixes #215

### DIFF
--- a/src/tabs/status.rs
+++ b/src/tabs/status.rs
@@ -46,6 +46,7 @@ pub struct Status {
     git_status_workdir: AsyncStatus,
     git_status_stage: AsyncStatus,
     queue: Queue,
+    git_action_executed: bool,
 }
 
 impl DrawableComponent for Status {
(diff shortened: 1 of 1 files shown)


extrawurst:
@jonstodle thanks ❤️

PR #217, opened 2020-07-29
Bump `dirs` version and add config migration

### LINKED ISSUE #175: dirs upgrade (migration of config)
see https://github.com/dirs-dev/dirs-rs/commit/887e2ae55eef326bd3c0b5080e28acd9b9073490

(related #174)

jonstodle:
Bumps `dirs` to version 3.0 and add migration logic to move the config files to the new location. The migration only happens on _macOS_, as it's the only OS affected.

This also fixes a bug where file paths with spaces would not open correctly in external editors. As part of the fix, the code for launching the editor switches to using `OsStr` instead of `str` to prevent gitui from messing file paths up should they contain invalid UTF-8.

Fixes #175 
Closes #174

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -323,11 +323,10 @@ dependencies = [
 
 [[package]]
 name = "dirs"
-version = "2.0.2"
+version = "3.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


extrawurst:
@jonstodle wow thanks for doing this! ❤️

PR #231, opened 2020-08-11
Add multiple selection in diff view

### LINKED ISSUE #229: Selection in diff to copy lines to clipboard
I regularly want to copy stuff out of diffs.

This requires supporting selecting multiple lines in the diff view. 

Then we can add a new command to copy the selected block into the clipboard. 

This selection function would also be the foundation for `per line staging` (#59)

cruessler:
This is a first attempt at implementing multiple selection in the diff view.
You can use `Shift+ArrowUp` as well as `Shift+ArrowDown` to extend the
selection.

Known issue: If the selection spans more than one hunk, only the hunk that
contains the selection’s start is selected. I don’t know a good way to address
that. Preventing the selection from spanning more than hunk seems like an
obvious choice, but then, of course, you could only copy lines out of one hunk
at a time.

Still left to do is the actual copying to clipboard, but that should be fairly
easy once the above issue is resolved.

Closes #229.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,11 @@ jobs:
         profile: minimal
         components: clippy
 
+    - name: Install dependencies for clipboard access
+      if: matrix.os == 'ubuntu-latest'
+      run: |
+        sudo apt-get -qq install libxcb-shape0-dev libxcb-xfixes0-dev
+
     - name: Build Debug
       run: |
         rustc --version
@@ -53,6 +58,10 @@ jobs:
         toolchain: stable
         profile: minimal
         target: x86_64-unknown-linux-musl
+
+    - name: Install dependencies for clipboard access
+      run: |
+        sudo apt-get -qq install libxcb-shape0-dev libxcb-xfixes0-dev
 
     - name: Setup MUSL
       run: |
--- a/src/components/diff.rs
+++ b/src/components/diff.rs
@@ -1,4 +1,6 @@
-use super::{CommandBlocking, DrawableComponent, ScrollType};
+use super::{
+    CommandBlocking, DrawableComponent, ExtendType, ScrollType,
+};
 use crate::{
     components::{CommandInfo, Component},
     keys,
@@ -28,10 +30,40 @@ struct Current {
 }
 
 ///
+#[derive(Clone, Copy)]
+enum Selection {
+    Single(usize),
+    Multiple(usize, usize),
+}
+
+impl Selection {
+    fn get_start(&self) -> usize {
+        match self {
+            Self::Single(start) | Self::Multiple(start, _) => *start,
+        }
+    }
+
+    fn get_end(&self) -> usize {
+        match self {
+            Self::Single(end) | Self::Multiple(_, end) => *end,
+        }
+    }
+
+    fn contains(&self, index: usize) -> bool {
+        match self {
+            Self::Single(start) => index == *start,
+            Self::Multiple(start, end) => {
+                *start <= index && index <= *end
+            }
+        }
+    }
+}
+
+///
 pub struct DiffComponent {
     diff: Option<FileDiff>,
     pending: bool,
-    selection: usize,
+    selection: Selection,
     selected_hunk: Option<usize>,
     current_size: Cell<(u16, u16)>,
     focused: bool,
@@ -52,7 +84,7 @@ impl DiffComponent {
             selected_hunk: None,
             diff: None,
             current_size: Cell::new((0, 0)),
-            selection: 0,
+            selection: Selection::Single(0),
             scroll_top: Cell::new(0),
             theme,
         }
@@ -73,7 +105,7 @@ impl DiffComponent {
         self.current = Current::default();
         self.diff = None;
         self.scroll_top.set(0);
-        self.selection = 0;
+        self.selection = Selection::Single(0);
         self.selected_hunk = None;
         self.pending = pending;
 
@@ -97,12 +129,14 @@ impl DiffComponent {
                 hash,
             };
 
-            self.selected_hunk =
-                Self::find_selected_hunk(&diff, self.selection)?;
+            self.selected_hunk = Self::find_selected_hunk(
+                &diff,
+                self.selection.get_start(),
+            )?;
 
             self.diff = Some(diff);
             self.scroll_top.set(0);
-            self.selection = 0;
+            self.selection = Selection::Single(0);
         }
 
         Ok(())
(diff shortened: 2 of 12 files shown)


extrawurst:
looks already like a great start. only two minor things:

1) is it intended that when u do both `up` and `down` that it simply grows the selection in both directions instead of moving the bottom like usual?
2) similar to 1) I think the scrolling should still work and go with the movable part of the selection
3) I think after stopping to do selection `up`/`down` without shift it should land where the movable part of the selection was

cruessler:
I have updated the PR.

ad 1) I found the initial behavior unexpected, too. I changed it so that `Shift+Arrow` now only moves the selection’s end which is closer to what you would expect if you come from vim.
ad 2), 3) I have not yet addressed that, but I’ll keep it in mind.

Doing the actual copying seems simple. At least on my machine (Ubuntu 20.04, X11), it worked immediately. Coming from vim, I have bound copy to `y`, but that can of course easily be changed.

extrawurst:
@cruessler I checked out the changes. works better now, but I would like to be able to also scroll up beyond the start selection line (basically inverting start/end selection index).
also I can confirm the copy to cb works on Mac too👍

cruessler:
@extrawurst I just updated the PR with another round of changes!

src/components/diff.rs
```diff
@@ -113,37 +167,80 @@ impl DiffComponent {
...
+                                .as_ref(),
+                        )
+                    } else {
+                        None
+                    }
+                })
+                .collect();
+
+            let mut ctx: ClipboardContext = ClipboardProvider::new()
+                .expect("failed to get access to clipboard");
```
extrawurst, line 236: when could this fail? are there reasonable cases? if so we could add a new error type and actually propagate it up, after all we are returning a result here.

src/components/diff.rs
```diff
@@ -553,6 +657,10 @@ impl Component for DiffComponent {
                         }
                         Ok(true)
                     }
+                    keys::COPY => {
```
extrawurst, line 660: can u add the appropriate keys to the command (so that they show up on the bottom of the screen)?

src/components/diff.rs
```diff
@@ -113,37 +167,80 @@ impl DiffComponent {
...
+            if new_start != old_start {
                 self.selected_hunk =
-                    Self::find_selected_hunk(diff, self.selection)?;
+                    Self::find_selected_hunk(diff, new_start)?;
             }
         }
         Ok(())
     }
 
+    fn modify_selection(
```
extrawurst, line 200: one last nit-pick: if u look how editors like vscode handle this there is one more special case: when u select and scroll down and then let go of `shift` it depends on whether ur next input is `up` or `down` where `up will jump back to the top of the selection and `down` will simply move the current bottom selection one more down. scrolling up from the start of the selection of course works the opposite way around.

.github/workflows/ci.yml
```diff
@@ -31,6 +31,11 @@ jobs:
         profile: minimal
         components: clippy
 
+    - name: Install dependencies for clipboard access
+      if: matrix.os == 'ubuntu-latest'
+      run: |
+        sudo apt-get -qq install libxcb-shape0-dev libxcb-xfixes0-dev
```
extrawurst, line 37: this is new third party dependency, people have to have installed right? how does this play with the release-linux-binary distribution? or are these libs that are present on every linux machine anyway?

src/components/diff.rs
cruessler: I just had a look at the [source], and it looks like it can fail on Linux when there is no X11 session. I’ll add a new error type.

[source]: https://docs.rs/x11-clipboard/0.5.1/src/x11_clipboard/lib.rs.html#54

.github/workflows/ci.yml
cruessler: `libxcb-shape0-dev` and `libxcb-xfixes0-dev` are the development files of `libxcb-shape0` and `libxcb-xfixes0`, and they need to be installed at compile time. I would assume that `libxcb-shape0` and `libxcb-xfixes0` which are required at runtime are present on any system that has X or Wayland installed (see `apt-cache rdepends libxcb-shape0 libxcb-xfixes0`).

If gitui is supposed to run on systems without X, we could put the parts related to clipboard access behind a flag and only conditionally compile them. What do you think?

src/components/diff.rs
cruessler: I changed it so that now the error is shown to the user. I tested that by having `copy_string` return a dummy error, but this is not included in the committed code.

src/components/diff.rs
cruessler: Done.

.github/workflows/ci.yml
extrawurst: I am no expert on linux, does this mean we have to provide different built binaries for people who want to use gitui on server they just ssh into? (sorry if this is a stupid question)
as long as we can assume x/wayland to be installed for most people I am fine with this. I don't want to further complicate the deployment pipeline

extrawurst:
@cruessler aside the linux question, can you rebase once more onto master, so we see if the windows pipeline (that got fixed now) runs through?

src/components/diff.rs
extrawurst: works great now!🥳

.github/workflows/ci.yml
cruessler: No worries! I just realized that the binaries you are releasing are linked statically which makes the problem disappear because those binaries include all the necessary libraries and will run on almost any Linux system, regardless of the libraries installed on that system.

cruessler: People will need those libraries only if they want to compile gitui themselves.

extrawurst: Ok great! Then we are ready to go🥳

extrawurst:
Thanks so much @cruessler - great work!

cruessler:
@extrawurst One addition: I just saw that there are packages for Fedora and Arch. It is possible that they might need to be adapted to support clipboard access, but I am not familiar with building packages, so I don’t know how they work.

extrawurst:
@ignatenkobrain maintains fedora and @wezm did AUR maybe you guys can chip in before we release this and ruin your day? :)

wezm:
Looks like there's a couple of new runtime dependencies (`xcb` and friends). That's no big deal. 👍

ignatenkobrain:
I think it is fine for Fedora, but right now we are not able to build gitui because there is some bug that makes rustc to segfault on ppc64le.

PR #236, opened 2020-08-16
Adjusted hook tests to work on windows

### LINKED ISSUE #235: Windows commit hooks without wsl
Current ci seems broken due to lack of wsl support and it turns out that without wsl our commit hook unit tests won't work.

It would be great if someone with access to a windows machine looks at this and makes them run able on vanilla windows 🤞

dr-BEat:
This PR fixes #235.
The extra newline at the beginning of the hook caused the following error on windows:
`error: cannot spawn .git/hooks/commit-msg: No such file or directory`

### DIFF
--- a/asyncgit/src/sync/hooks.rs
+++ b/asyncgit/src/sync/hooks.rs
@@ -105,6 +105,7 @@ fn run_hook(
     let output = Command::new("bash")
         .args(bash_args)
         .current_dir(path)
+        .env("DUMMYENV", "FixPathHandlingOnWindows") // This call forces Command to handle the Path environment correctly on windows, the specific env set here does not matter
         .output();
 
     let output = output.expect("general hook error");
--- a/asyncgit/src/sync/hooks.rs
+++ b/asyncgit/src/sync/hooks.rs
@@ -183,8 +183,7 @@ mod tests {
         let root = repo.path().parent().unwrap();
         let repo_path = root.as_os_str().to_str().unwrap();
 
-        let hook = b"
-#!/bin/sh
+        let hook = b"#!/bin/sh
 exit 0
         ";
 
@@ -204,8 +203,7 @@ exit 0
         let root = repo.path().parent().unwrap();
         let repo_path = root.as_os_str().to_str().unwrap();
 
-        let hook = b"
-#!/bin/sh
+        let hook = b"#!/bin/sh
 echo 'msg' > $1
 echo 'rejected'
 exit 1
@@ -230,8 +228,7 @@ exit 1
         let root = repo.path().parent().unwrap();
         // let repo_path = root.as_os_str().to_str().unwrap();
 
-        let hook = b"
-#!/bin/sh
+        let hook = b"#!/bin/sh
 echo 'msg' > $1
 echo 'rejected'
 exit 1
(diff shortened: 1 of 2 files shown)


dr-BEat:
Apparently my local setup is different from the CI system.
It still returns the same error: `Windows Subsystem for Linux has no installed distributions.\r\r\nDistributions can be installed by visiting the Microsoft Store:\r\r\nhttps://aka.ms/wslstore\r\r\n`

extrawurst:
@dr-BEat thanks for looking into it. Is the wsl default installed on every Windows nowadays? Or do we have to simulate it by using a vanilla windows without wsl?

dr-BEat:
@extrawurst no problem. Its not installed by default, however I think they added a `bash.exe` by default that gives the error above.
Usually any git installation on Windows brings a real bash, so it works out normally.
The `run_hook` functions in hooks.rs tries to call `bash` directly and it gets the fake Windows one.

extrawurst:
Any idea how to fix this? Maybe just install some git using chocolaty in the ci?

dr-BEat:
Yeah installing git might fix it.

dr-BEat:
Gits already installed in the CI image. (:
https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md

dr-BEat:
According to this blog by Jon Skeet https://codeblog.jonskeet.uk/2019/10/12/using-git-bash-from-appveyor/ the order of directories in the Path environment can break stuff. 
I tried to add the git path as the first entry, but it does not seem to work.

extrawurst:
@dr-BEat any more ideas?

dr-BEat:
@extrawurst nothing concrete but I will try some tricks, probably the path is not correct still

dr-BEat:
@extrawurst the path environment is correct, the problem is that `Command::new("bash")` in hooks.rs line 105 somehow does not follow normal windows rules in finding which bash to execute.
If I hardcode the full path to the git bash with `Command::new(r"C:\Program Files\Git\bin\bash.exe")` it works perfectly.

extrawurst:
@dr-BEat interesting. maybe because we run the `cargo test` in the git shell, did you try setting the `shell: cmd` in the ci to make him use the windows stuff?

dr-BEat:
@extrawurst I do not think its related to the shell. I can reproduce the issue locally when just running cargo test from Powershell

extrawurst:
@dr-BEat nice a repro is what we needed!!🥳 I wish I could add any windows knowledge whatsoever :(

dr-BEat:
@extrawurst found it! Calling `env` on the command fixed it. This is probably related to the discussion here https://github.com/rust-lang/rust/issues/37519. The behaviour of `Command::spawn()` changes because the environment is passed differently to Windows if `env` is called.

extrawurst:
OMFG!!! you are the man!!

asyncgit/src/sync/hooks.rs
```diff
@@ -105,6 +105,7 @@ fn run_hook(
     let output = Command::new("bash")
         .args(bash_args)
         .current_dir(path)
+        .env("DUMMYENV", "FixPathHandlingOnWindows") // This call forces Command to handle the Path environment correctly on windows, the specific env set here does not matter
```
extrawurst, line 108: wtf... this is crazy! 🙈

Issue #256, opened 2020-08-29
RFE: Switch from clipboard to copypasta


ignatenkobrain:
The `clipboard` is using some really old version of `x11-clipboard` on Linux. `copypasta` is fork of `clipboard` which supports Wayland and is generally maintained these days.

Sorry that I did not spot this last time you mentioned me on the issue about new dependency.

ignatenkobrain:
Or even https://crates.io/crates/cli-clipboard which is probably better in terms of wayland support, though I did not try it yet.

ignatenkobrain:
@extrawurst let me know if you won't have time to play with this in near future in which case I'll try to write patch myself :)

PR #258, opened 2020-08-29
Switch to cli-clipboard

ignatenkobrain:
It has proper support for Wayland clipboard in CLI applications and is
automatically falling back to X11 when that fails.

Closes: https://github.com/extrawurst/gitui/issues/256
Signed-off-by: Igor Raits <i.gnatenko.brain@gmail.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -25,6 +25,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ansi_term"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ee49baf6cb617b853aa8d93bf420db2383fab46d314482ca2803b40d5fde979b"
+dependencies = [
+ "winapi",
+]
+
+[[package]]
 name = "anyhow"
 version = "1.0.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


ignatenkobrain:
I did not test macos/windows, but on Linux (Fedora with GNOME, both X11 and Wayland sessions) works just fine.

extrawurst:
@ignatenkobrain cool thanks for going at this. Does this change anything about the new dependencies clipboard on Linux required I the CI? See https://github.com/extrawurst/gitui/blob/master/.github/workflows/ci.yml#L36

ignatenkobrain:
@extrawurst  I don't think it changes anything here because it still uses `x11-clipboard` for X11 (which `clipboard` also uses, though older version) and wayland one uses by default Rust implementation of wayland stuff so no new dependencies are needed.

I am not familiar with Ubuntu split of X11, but on Fedora is was enough to install `libxcb-devel`. Probably you could even change it like this:

```diff
-        sudo apt-get -qq install libxcb-shape0-dev libxcb-xfixes0-dev
+        sudo apt-get -qq install libxcb1-dev libxcb-xfixes0-dev
```

extrawurst:
@ignatenkobrain this seems to work on macOS, but currently I am reluctant to merge this before we have the linux distribution problem solved: https://github.com/extrawurst/gitui/issues/259

ignatenkobrain:
Fine with me, no rush.

extrawurst:
@ignatenkobrain change of plans, since we obviously cannot use x11 and the clipboard feature in static binaries I removed this feature from the linux version for now. on discord we are considering the options here now. maybe you want to chime in:
1) is there a way to build against this stuff statically (musl)
2) going the `tig`/`neovim` way and use a shell call to `xclip` or `xsel` to fill the clipboard - how probable is such a tool to be installed on a regular linux distro?

wezm:
> how probable is such a tool to be installed on a regular linux distro?

I think it's unlikely these would be installed by default on a regular distro/BSD. I'm not sure that's a huge issue though. If the `gitui` was unable to spawn the process it could tell the user that they need to install it.

extrawurst:
closes in favour of https://github.com/extrawurst/gitui/issues/262

PR #261, opened 2020-08-31
revert clipboard feature on linux to fix static linux binary build

### LINKED ISSUE #259: Failed to execute gitui binary file downloaded from the release page
**Describe the bug**

I downloaded gitui from https://github.com/extrawurst/gitui/releases/download/v0.10.0/gitui-linux-musl.tar.gz
extract and execute `./gitui`, failed with:
~~~
$ ./gitui
Failed to execute process './gitui'. Reason:
The file './gitui' does not exist or could not be executed.
~~~
And the ./gitui has the right permission.

FYI:
~~~
$ ldd ./gitui
linux-vdso.so.1 (0x00007ffe637e1000)
libxcb.so.1 => /usr/lib/libxcb.so.1 (0x00007f706f942000)
libXau.so.6 => /usr/lib/libXau.so.6 (0x00007f706f93d000)
libXdmcp.so.6 => /usr/lib/libXdmcp.so.6 (0x00007f706f935000)
libc.so.6 => /usr/lib/libc.so.6 (0x00007f706f76c000)
/lib/ld64.so.1 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007f706f9b0000)
~~~

**To Reproduce**
1. download https://github.com/extrawurst/gitui/releases/download/v0.10.0/gitui-linux-musl.tar.gz
2. extract it
3. execute it

**Expected behavior**

Successfully working.

**Screenshots**

 - OS/Distro + Version: [5.7.17-2-MANJARO]
 - GitUI Version [v0.10.0]
 - Rust version: [rustc 1.45.0 (5c1f21c3b 2020-07-13)]

**Additional context**

extrawurst:
this reverts the new dependency on clipboard and x11 in turn so that static linux binaries can be build again. 
we need to look into a different solution later.

closes #259

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,6 +72,8 @@ jobs:
     - name: Build Debug
       run: |
         cargo build --target=x86_64-unknown-linux-musl
+        ls -lisa ./target/x86_64-unknown-linux-musl/debug
+        ldd ./target/x86_64-unknown-linux-musl/debug/gitui
         ./target/x86_64-unknown-linux-musl/debug/gitui --version
     - name: Build Release
       run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,14 +71,11 @@ jobs:
         sudo apt-get -qq install musl-tools
     - name: Build Debug
       run: |
-        cargo build --target=x86_64-unknown-linux-musl
-        ls -lisa ./target/x86_64-unknown-linux-musl/debug
-        ldd ./target/x86_64-unknown-linux-musl/debug/gitui
+        make build-linux-musl-debug
         ./target/x86_64-unknown-linux-musl/debug/gitui --version
     - name: Build Release
       run: |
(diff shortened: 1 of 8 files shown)


wezm:
One possibility for X11 would be to shell out to `xclip` or `xsel`. This is the approach taken by [pass](https://www.passwordstore.org/) and [Neovim](https://neovim.io/).

extrawurst:
@wezm actually considering this now. I guess there is no way to build against a static version of x11 using musl, right?

extrawurst:
@wezm also see here: https://github.com/extrawurst/gitui/pull/258#issuecomment-684108120

wezm:
> I guess there is no way to build against a static version of x11 using musl, right?

It's possible but probably not with the `x11-clipboard` crate.The [x11 crate](https://github.com/erlepereira/x11-rs) (as opposed to the x11-dl variant) statically links by default. So, if this was used instead then it would be possible. However, I think the `xsel`/`xclip` option is best if you don't want to have a dependency on X11 directly.

PR #263, opened 2020-09-01
Added additional unit tests for current behaviour

ciaranmooney:
This PR adds additional unit tests to the `textinput` component covering existing behaviour. 

The Style component of the Vec<Text> output of `get_draw_text()` is now also checked which is necessary to check that the cursor has been rendered in the correct position.


ciaranmooney:
Merged upstream

PR #264, opened 2020-09-01
Additional Unit Tests for TextInput Component

ciaranmooney:
This PR adds additional unit tests to the textinput component covering existing behaviour.

The Style component of the Vec output of get_draw_text() is now also checked which is necessary to check that the cursor has been rendered in the correct position.


src/components/textinput.rs
```diff
@@ -312,6 +351,10 @@ mod tests {
             "",
         );
 
+        let underlined = Style::new();
+        let underlined = underlined.fg(Color::DarkGray);
```
extrawurst, line 355: the only problem I can see is here that this particular color is theme dependant. It would be cooler if u could grab the default theme and compare against the correct call we use to set the color for this one

extrawurst (commented):
cool stuff, only one minor change requiered

src/components/textinput.rs
ciaranmooney: Okay, the unit tests now derives it's data from the default theme.

```Rust
let theme = SharedTheme::default();
let underlined = theme.text(false, false);
let underlined = underlined.modifier(Modifier::UNDERLINED);
```

extrawurst:
@ciaranmooney can you fix the formatting, then we are good to go

ciaranmooney:
Thanks. I'll try and get it done tonight.

I noticed the build failure after pushing up my changes which was rather
frustrating. That's the third or fourth time `rustfmt` has burnt my
fingers. I need to figure out how to do rub the CI build locally.

extrawurst:
just run `make check` in the root of the project, it will check formatting and run the unit tests

Issue #228, opened 2020-08-09
git error:get_string called on a live config object when trying to commit


c02y:
**Describe the bug**

**To Reproduce**
1. open gitui inside a git repo
2. stage a modified file
3. write a commit message
4. try to commit
5. commit failed, "git error:get_string called on a live config object"

I tried gitui multiple times, failed to commit, but this commit is done successfully using tig.

**Expected behavior**
commit the commit message successfully

**Screenshots**

**Context (please complete the following information):**
 - OS/Distro + Version: [5.7.14-1-MANJARO]
 - GitUI Version [0.9.1]
 - Rust version: [1.45.2]

**Additional context**

extrawurst:
seems related to https://github.com/extrawurst/gitui/issues/79
can you do me a favour and share the config for this repo for the `user.name` and `user.email` ?

c02y:
`user.name`  == empty
`user.email` == "myemail@email.com"

extrawurst:
it is exactly the opposite of #79
<img width="554" alt="image" src="https://user-images.githubusercontent.com/776816/89728762-fb244e00-da2f-11ea-91a3-1aa45c467cac.png">

so it fails if u have a name but no email defined.
```
[user]
         name = Stephan Dilly
```

c02y:
Yes, I get it now, but I used tig to commit it successfully, which is weird, since `user.name` can be undefined in the commit step.

I think we can treat this issue an bug of gitui

cruessler:
If this is not taken yet, I’d have a look.

extrawurst:
@cruessler sure! 👍

extrawurst:
@cruessler any progress on this? I am thinking about cutting a release to get the `vim` key bindings out the door

cruessler:
@extrawurst Sorry for the delay. I’ll now start working on it.

cruessler:
@c02y Running `git commit` with `user.email` not set fails on my machine both using `git` and `tig` with the exact same error message: `*** Please tell me who you are.` If I am not overlooking anything, `gitui` could improve on its error message, but does not do anything unusual.

```
❯ git --version
git version 2.25.1

testrepo on master [+]
❯ tig -v
tig version 2.4.1
ncursesw version 6.2.20200212
readline version 8.0
```

dgrine:
+1

extrawurst:
#269 actually fixes this, but having reliable tests for this is a pickle, but hey lets just bite that bullet

Issue #79, opened 2020-05-20
The commit function does not work when you have user.email but no user.name defined


alistaircarscadden:
**Describe the bug**
In gitui v0.3.0 the commit function does not allow a commit to be made when the user has a user.email defined and no user.name defined.

It took me a while to discover that there is a difference between user.name being defined as empty (`git config user.name ""`) and it being undefined (as in, `git config --list` does not show the entry at all).

In the empty case, interestingly enough, git commit does not allow the commit to be made, likewise gitui does not allow it.

In the undefined case, git commit does allow the commit to be made (and I think it shows `unknown` in that field in later `git log` or `git show` calls). This is where the gitui behaviour is different, as gitui does not allow a commit to be made with this message:
```
 Status   |   Log
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
┌Unstaged Changes [1]──────┌Error──────────────────────────────────────────────────────────┐───────────────────────────┐
│                          │commit failed:                                                 │                           │
│                          │git error:config value 'user.name' was not found; class=Config │                           │
│                          │(7); code=NotFound (-3)                                        │                           │
│                          │                                                               │                           │
```

**Expected behavior**
gitui would allow a commit to be made and fill in `unknown` as git commit does.

extrawurst:
followup to #74

extrawurst:
unfortunately #155 had to be rolled back

extrawurst:
removed `good first issue` since #155 showed that writing an appropriate unittest for this is by far more complicated than expected

extrawurst:
still active: https://github.com/extrawurst/gitui/pull/269

extrawurst:
@cruessler is this fixed by your PR?

extrawurst:
@alistaircarscadden @cruessler closing this hoping that #269 actually fixed this. please reopen if this is not the case

PR #269, opened 2020-09-02
Use get_entry instead of get_str

cruessler:
Related to #79 and #228.

Beware that the additional tests most likely will fail locally because the local git config interferes with them. This PR is intended for testing whether there are issues on CI.

### DIFF
--- a/asyncgit/src/sync/commit.rs
+++ b/asyncgit/src/sync/commit.rs
@@ -36,17 +36,24 @@ pub fn amend(
 fn signature_allow_undefined_name(
     repo: &Repository,
 ) -> std::result::Result<Signature<'_>, git2::Error> {
-    match repo.signature() {
-        Err(e) if e.code() == ErrorCode::NotFound => {
-            let config = repo.config()?;
-            Signature::now(
-                config.get_str("user.name").unwrap_or("unknown"),
-                config.get_str("user.email")?,
-            )
-        }
+    let signature = repo.signature();
 
-        v => v,
+    if let Err(ref e) = signature {
+        if e.code() == ErrorCode::NotFound {
+            let config = repo.config()?;
+
+            if let (Err(_), Ok(email_entry)) = (
+                config.get_entry("user.name"),
+                config.get_entry("user.email"),
+            ) {
+                if let Some(email) = email_entry.value() {
+                    return Signature::now("unknown", email);
+                }
+            };
+        }
     }
+
+    signature
 }
 
 /// this does not run any git hooks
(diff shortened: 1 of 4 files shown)


extrawurst:
> Beware that the additional tests most likely will fail locally

how are you planning to work around this?

>  This PR is intended for testing whether there are issues on CI.

looks good doesn't it?

cruessler:
> > Beware that the additional tests most likely will fail locally
> 
> how are you planning to work around this?

I don’t have a solution yet, so I’ll do a little research to see what other projects are doing.

> > This PR is intended for testing whether there are issues on CI.
> 
> looks good doesn't it?

Yes, the git config used on CI seems to not interfere with the new tests.

cruessler:
@extrawurst I might have found a solution. Setting `$HOME` to an empty directory in `repo_init_empty` prevents `git` from falling back to `$HOME/.gitconfig` because it that is now guaranteed to be empty. `git` itself does not provide more granular env variables as far as I could tell. I’m not 100 % sure whether undesired side-effects can occur if we just change `$HOME` in the middle of a test, though.

extrawurst:
Sorry for the delay, this would break tests that depend on `$HOME` to be something useful, right? and since we run tests in parallel it would be non-deterministic.

cruessler:
I did a little research to see how other tools tackle this issue, and this is what I found:

- `tig` explicitly sets `$HOME` to a temporary directory for its tests.

  https://github.com/jonas/tig/blob/b878abf5a0da2efbdd29b8e0215d4df0c7a8109f/test/tools/libtest.sh#L44

- `git` also uses this technique.

  https://github.com/git/git/blob/7814e8a05a59c0cf5fb186661d1551c75d1299b5/t/test-lib.sh#L1372

- `libgit2` has tests that modify `$HOME` to make sure everything works.

  https://github.com/libgit2/libgit2/blob/2376cd26226771dcf8ef5dfd04c83a50c50bf3d4/tests/core/env.c#L120

I share your concerns, though, and, to be honest, I was quite surprised to find that this seems to be a common practice.

To make the tests independent of their order, we could set `$HOME` at the beginning of the test suite so that all tests share the same environment.

extrawurst:
@cruessler thanks for researching this. Ok lets go ahead with this approach, but let's do one more thing: Let's limit the impact of this change to the test that requires this instead of all using `init_repo_empty`.

We should make a utility that helps us set the `env` and protect it using a mutex guard that we can use to have multiple tests run in serial that depend on this behaviour, also this utility can revert the changes to `env` when dropped at the end of the test. Does that make sense to you?

can you make those changes and mark the PR as ready to review?

cruessler:
@extrawurst Sounds good to me! I'll update the PR in the next few days.

cruessler:
It looks like changing `$HOME` just for one test is not trivial. This is due to the following facts:

-   `libgit2` has some global state that is initialized only once. This happens e. g. on the first call to `Repository::init`.
-   The initialization calls `git_libgit2_init` which sets up some global state.
-   In particular, [the current value of `$HOME` is read and cached][home].

That means that whatever value `$HOME` had when `Repository::init` was first called will be used for further calls to `repo.config()`, among others. Effectively, all tests within the same thread run with the same value for `$HOME`.

I briefly tried [rusty-fork], but could not get it to work quickly.

[home]: https://github.com/libgit2/libgit2/blob/e9b0cfc006ceebda563d1084b1cb6604ac13a15f/src/sysdir.c#L117
[rusty-fork]: https://crates.io/crates/rusty-fork

extrawurst:
@cruessler omg why do they make it so hard on us? :) 
Look like we need to call `git_libgit2_shutdown` (https://github.com/rust-lang/git2-rs/blob/master/libgit2-sys/lib.rs#L1829) before we run a test that is `HOME` sensitive cause it is supposed to wipe clean those caches, right?

cruessler:
@extrawurst Looks like that is not an option, either. :-) We would have to depend on `libgit2-sys` in `asyncgit` which produces the following error message:

    error: multiple packages link to native library `git2`, but a native library can be linked only once

I searched far and wide, but `git_libgit2_shutdown` does not seem to be exposed by `git2` itself, so there’s no way to call it as far as I can see.

extrawurst:
looks like u have an opportunity to become a `git2-rs` contributor :D

cruessler:
@extrawurst There is already an issue at `git2-rs` regarding environment variables. They’re considering mentioning this behaviour in the docs, so I don’t have the impression that `git2-rs` will be exposing `git_libgit2_shutdown` anytime soon. :-)

extrawurst:
@cruessler as far as I can see no one ever raised the idea of exposing `shutdown` so I did now: https://github.com/rust-lang/git2-rs/issues/550#issuecomment-706549734

asyncgit/src/sync/commit.rs
```diff
@@ -245,4 +252,91 @@ mod tests {
...
         Ok(())
     }
+
+    /// Beware: this test has to be run with a `$HOME/.gitconfig` that has
+    /// `user.email` not set. Otherwise, `git` falls back to the value of
+    /// `user.email` in `$HOME/.gitconfig` and this test fails.
+    ///
+    /// As of September 2020, `repo_init_empty` sets `$HOME` to an empty
+    /// temporary directory, so this constraint is met.
+    #[test]
```
extrawurst, line 262: @cruessler coming back to this now, I am wondering:
1) can we utilise https://github.com/dotenv-rs/dotenv to have control over what is used as `$HOME` in `cargo test`?
2) is it possible to exclude this test locally and only run in CI?

cruessler: @extrawurst I’ll have a look!

extrawurst:
@cruessler also maybe this change is relevant for us: https://github.com/rust-lang/git2-rs/pull/656 ?

cruessler:
> @cruessler also maybe this change is relevant for us: [rust-lang/git2-rs#656](https://github.com/rust-lang/git2-rs/pull/656) ?

Thanks for the hint! I just checked locally, and on my machine the issue was solved and the new tests were green. I’ll update the branch, so we can see what CI says.

cruessler:
There is only one downside: `set_search_path` is `unsafe` and `asyncgit` forbids unsafe code.

extrawurst:
@cruessler well that's a concession I can make. can you quickly add that to the changelog aswell, then I can merge it 👍

cruessler:
@extrawurst Great, done!

extrawurst:
@cruessler Thanks❤️

Issue #262, opened 2020-08-31
enable clipboard support on linux again


extrawurst:
#261 reverted this feature to fix #259 

so now we have two options:
1) figure out how to use `clipboard` in a static build binary using musl
2) use a shell call against something like: `xclip` or `xsel`

also see #258

extrawurst:
looks like we can use `pbcopy` on mac and `clip` on windows then we can rip out the `clipboard` dependency completely🥳

PR #273, opened 2020-09-05
Added XCB dependencies to README

VictorKoenders:
I tried installing gitui through cargo on my new ubuntu machine and ran into some dependencies it was missing.
I figured I'd add them to the README as I couldn't find any information on it yet

This only works for ubuntu, I don't know what other linux distros would need (arch, fedora, etc). If anyone wants to look that up I'll add them to the PR

[Rendered](https://github.com/VictorKoenders/gitui/blob/ab6e5190ff0cc0e867b43a4dffceacf628e7cefc/README.md#build)

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,8 @@ Binaries available for:
 
 - Latest `rust` and `cargo`
   - See [Install Rust](https://www.rust-lang.org/tools/install)
+- xcb
+  - Ubuntu: `sudo apt install libxcb-render0-dev libxcb-shape0-dev libxcb-xfixes0-dev`
 
 ### Cargo Install
 


extrawurst:
we disabled the clipboard support in the linux release binaries because of this in `0.10.1` and plan to implement it in a dependency-less way later: https://github.com/extrawurst/gitui/issues/262

extrawurst:
In the meantime I think it makes more sense to have a PR that disables the `clipboard` feature in cargo.toml *just* for linux builds but that is beyond my cargo-foo and I did not have time to look into it since this is a temporary limitation

extrawurst:
dependency removed: #262

PR #288, opened 2020-09-25
Fold folders up into parent if alone in directory

### LINKED ISSUE #192: Compact Treeview
Thanks for building this tool.

When files changed are down in the tree hierarchy, having branching of every folder in the way of the file is unnecessary and a little annoying.

For example (android project),
Screen Shot 2020-07-10 at 7 50 40 PM
in the example above, `app/src/main` can be combined together in one line instead of having `app`, `src` and `main` in 3 different lines.

WizardOhio24:
Closes #192.
Folds folders up into their parent if they are the only item it their directory.  For example:
Before:
before_screenshot
After:
after_screenshot

### DIFF
--- a/src/components/filetree.rs
+++ b/src/components/filetree.rs
@@ -251,13 +251,11 @@ impl FileTreeComponent {
                     && tree_items[idx_temp + 1].info.indent
                         < tree_items[idx_temp + 2].info.indent
                 {
-                    // fold up the folder/file
-                    // because there is only one in the directory
+                    // fold up the folder/file because there is only one in the directory
                     idx_temp += 1;
                     should_skip_over += 1;
 
-                    // check if there is another folder or file at the
-                    // same level, if there is, don't fold up
+                    // check if there is another folder or file at the same level, if there is, don't fold up
                     let mut idx_temp_inner;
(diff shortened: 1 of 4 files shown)


extrawurst:
@WizardOhio24 thanks for tackling this! Very appreciated. But on a first glance I do not see any **tests**. can you cove this new implementation with a few tests?

WizardOhio24:
Will do

WizardOhio24:
Tests added, let me know if there's anything else that should be added/changed

extrawurst:
I think we need more unittests for the selection moving code, there is still some strange bug with collapsing a folded up path:

Screenflick Movie 70

WizardOhio24:
I didn't consider the right and left arrow keys for navigation (only up/down), thanks for pointing that out.  It should be an easy fix.

WizardOhio24:
Bug fixed and right/left navigation tests added.
It was left navigation which was going to the previous index, rather than using the new method (available_selections) in up/down navigation.

extrawurst:
Really cool thanks!

extrawurst:
@WizardOhio24 damn just found another bug: `left` used to jump up to the current parent directory. now it seems to only move up one

PR #289, opened 2020-09-26
Bump tui: 0.9 -> 0.12

### LINKED ISSUE #205: migrate to tui-rs 0.10
[v0.10 release](https://github.com/fdehau/tui-rs/blob/master/CHANGELOG.md#v0100---2020-07-17) contains breaking changes.

afterwards we can finally fix: https://github.com/extrawurst/gitui/issues/149

WizardOhio24:
Closes #205, Closes #285  

Notable changes:
 - `Text` -> `Span`
 - `Paragraph` was changed such that '\n' no longer creates a new line, rather `Spans` (`Vec<Span>`) has to be used where each `Spans` would be a new line

(I've tried to fix all the paragraphs but if there is a paragraph which shows its elements horizontally rather than vertically, this PR is most likely the cause)

tui::widgets::Paragraph.title_style() is deprecated and command::CommandInfo.print() is no longer used, should these be fixed in this PR?  (They're easy fixes)
Also, the test for line endings in DiffComponent was removed as it no longer applies.

### DIFF
--- a/src/components/command.rs
+++ b/src/components/command.rs
@@ -79,11 +79,6 @@ impl CommandInfo {
     }
 
     ///
-    pub fn print(&self, out: &mut String) {
-        out.push_str(&self.text.name);
-    }
-
-    ///
     pub const fn show_in_quickbar(&self) -> bool {
         self.quick_bar && self.available
     }
--- a/src/components/command.rs
+++ b/src/components/command.rs
@@ -1,6 +1,6 @@
 ///
 #[derive(Clone, PartialEq, PartialOrd, Ord, Eq)]
-pub struct CommandText {
+pub struct CommandSpan {
     ///
     pub name: String,
     ///
@@ -11,7 +11,7 @@ pub struct CommandText {
     pub hide_help: bool,
 }
 
-impl CommandText {
+impl CommandSpan {
     ///
     pub const fn new(
         name: String,
@@ -36,7 +36,7 @@ impl CommandText {
 ///
 pub struct CommandInfo {
     ///
-    pub text: CommandText,
+    pub text: CommandSpan,
     /// available but not active in the context
     pub enabled: bool,
     /// will show up in the quick bar
@@ -51,7 +51,7 @@ pub struct CommandInfo {
 impl CommandInfo {
     ///
     pub const fn new(
-        text: CommandText,
+        text: CommandSpan,
         enabled: bool,
         available: bool,
     ) -> Self {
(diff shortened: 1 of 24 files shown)


WizardOhio24:
There's now no more clippy warnings caused by upgrading tui to 0.11.  The deprecations warnings have been fixed and command.print() has been removed.

extrawurst:
@WizardOhio24 could you rebase, then nightly should work again

WizardOhio24:
tui 0.12 has been released, Great!  This allows `Text` to contain \n characters again.  For 0.12, would you want to go back to using `Text` or keep the `Span`/`Spans` as in this PR?  The library seems to suggest using `Span`/`Spans`: https://docs.rs/tui/0.12.0/tui/text/index.html

> These types form a hierarchy: Spans is a collection of Span and each line of Text is a Spans.

WizardOhio24:
tui recently updated to 0.12 which allows `Text` to contain '\n' characters again.  I can bump this to 0.12 but due to how tui 0.11 worked, all the `Text`s have been changed to `Span`/`Spans`.  Although, the docs for tui do suggest using `Span`/`Spans` for multiple lines, https://docs.rs/tui/0.12.0/tui/text/index.html:
>These types form a hierarchy: Spans is a collection of Span and each line of Text is a Spans.

extrawurst:
@WizardOhio24 ok then lets stick to `Span`/`Spans` then. I guess that's the cleaner way to go. I did not check this PR yet, since I want to have the folder-folding stuff sorted before opening the next can :)

src/components/command.rs
```diff
@@ -1,6 +1,6 @@
 ///
 #[derive(Clone, PartialEq, PartialOrd, Ord, Eq)]
-pub struct CommandText {
+pub struct CommandSpan {
```
extrawurst, line 3: I think this is not necessary, it has nothing to do with those `tui-rs` naming

src/components/commit.rs
```diff
@@ -1,5 +1,5 @@
 use super::{
-    textinput::TextInputComponent, visibility_blocking,
+    textinput::SpanInputComponent, visibility_blocking,
```
extrawurst, line 2: Same as above, we do not have to rename everything that has `Text` in the name just because `tui-rs` made that change

extrawurst:
the line breaks in the commit msg window seem broken:

image

extrawurst:
@WizardOhio24 thanks so much!

PR #294, opened 2020-09-28
Fix left key in tree broken

### LINKED ISSUE #291: left key in tree broken
before https://github.com/extrawurst/gitui/pull/288 the `left` key *jumped* up to the current parent dir, now it simply moves one up:
Screenflick Movie 71

WizardOhio24:
Fixes #291

Pressing left now causes the selection to jump to the parent rather than just moving up one.


extrawurst:
@WizardOhio24 thanks man!

Issue #91, opened 2020-05-27
Manage and checkout branches


alistaircarscadden:
**Is your feature request related to a problem? Please describe.**
I would like to be able to do functionality such as:

`git checkout -b newbranch`
`git branch -d oldbranch`

**Describe the solution you'd like**
Functionality similar to commit in the Status menu, that would perform these functions.

extrawurst:
Hi @alistaircarscadden I was thinking about this before, I don't know a good place to put it in right now. In what tab would you expect this functionality?

alistaircarscadden:
Here's my brainstorming. Obviously there's a lot of room for improvement with my design!

Perhaps from Status you'd press B and be taken to another page that isn't quite a tab. I'd imagine doing something like (from Status)

`Branch [b]`

Would initially show

```
 Status   |   Log   |
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
┌Branch────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
|   dev       7a9684a add: new stuff
|   dev_0.0.2 eccf7b4 clean up some stuff
|   dev_27    efe89c0 Merge branch 'master' of https://github.com/stuff/stuff into dev_27
| * dev_41    50be3ac add: good ideas and stuff
└──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
Switch branches [b] Create new branch [c] Rename branch [m] Set upstream [u] Detach from [d] Merge branch [m]```

The contents would be generated similar to `git branch --list --verbose --verbose (two verboses prints the upstream branches)`.

Some popups would look like:

```
┌Upstream branch───────────────────────────────────────────────────────┐
│type branch name..                                                    │
└──────────────────────────────────────────────────────────────────────┘
Set upstream branch [Enter] Remove usptream branch [Backspace] Close [esc]```

Which would perform `git branch <local_branch> --set-upstream-to=<remote_branch>`

```
┌New branch─────────────────────────────────From───────────────────────┐
│type branch name..                        |(optional)                 │
└──────────────────────────────────────────────────────────────────────┘
Create new branch [Enter] Create new branch and checkout [Shift+Enter] Close [esc]```

Which would perform `git branch <new_branch>` and `git branch <new_branch> <old_branch>` and `git checkout -b <new_branch_to_checkout_to>`.

```
┌Rename branch──────────────────────────────From───────────────────────┐
│type branch name..                        |(optional)                 │
[...]

PR #303, opened 2020-10-03
Branch popup

WizardOhio24:
Implements a branch popup which can be accessed by pressing 'b' in the status tab.
branch_popup
A new branch can be created by pressing 'c' while the branch popup is open and a branch can be checked out by selecting the branch and pressing enter.

This does currently work but it is still a work in progress.

Linked to #91.


WizardOhio24:
This is now complete, bar that getting the branches is currently sync and should probably be async or cached (so the calls are currently inefficient).  I'm not sure about the async stuff so it might take a bit of time to work out.
Also, the styling will be added after #289  is merged as the text will have to be changed to Span and Spans.

asyncgit/src/sync/branch.rs
```diff
@@ -28,6 +28,86 @@ pub(crate) fn get_branch_name(repo_path: &str) -> Result<String> {
...
+/// Used to return only the nessessary information for displaying a branch
+/// rather than an iterator over the actual branches
+pub fn get_branches_to_display(
+    repo_path: &str,
+) -> Vec<BranchForDisplay> {
+    if let Ok(cur_repo) = utils::repo(repo_path) {
+        cur_repo
+            .branches(None)
+            .map_err(super::super::error::Error::Git)
+            .expect("")
```
extrawurst, line 55: We should return `Result` here. every `expect` can then be made `?` or contain an actual meaningful text

asyncgit/src/sync/branch.rs
```diff
@@ -28,6 +28,86 @@ pub(crate) fn get_branch_name(repo_path: &str) -> Result<String> {
...
+    // hasn't been committed or stashed, in this case it will Err
+    if let Ok(repo) = utils::repo(repo_path) {
+        if repo.set_head(branch_ref).is_ok()
+            && repo.checkout_head(None).is_ok()
+        {
+            return Ok(());
+        }
+    }
+
+    // TODO change this
```
extrawurst, line 107: change to what?

src/app.rs
```diff
@@ -269,6 +276,7 @@ impl App {
         self.revlog.update()?;
         self.stashing_tab.update()?;
         self.stashlist_tab.update()?;
+        self.select_branch_popup.update_branches();
```
extrawurst, line 279: why not use the same `update` schema like we have in the other components?

src/components/select_branch.rs
```diff
@@ -0,0 +1,269 @@
...
+impl DrawableComponent for SelectBranchComponent {
+    fn draw<B: Backend>(
+        &self,
+        f: &mut Frame<B>,
+        _rect: Rect,
+    ) -> Result<()> {
+        // Render a scrolllist of branches inside a box
+
+        if self.visible {
+            const SIZE: (u16, u16) = (50, 20);
```
extrawurst, line 47: how does this behave when resizing the terminal to a smaller size?

src/components/select_branch.rs
```diff
@@ -0,0 +1,269 @@
...
+            .map(|displaybranch| displaybranch.name.len())
+            .max()
+            .expect("Failed to find max branch length");
+
+        for (i, displaybranch) in self.branch_names.iter().enumerate()
+        {
+            let mut commit_message =
+                displaybranch.top_commit_message.clone();
+            if commit_message.len() > 30 {
+                commit_message.truncate(30);
```
extrawurst, line 224: does this work in between unicode grapheme boundaries or happen on byte boundaries? if the latter then it will panic when slicing inside a unicode cluster

src/components/select_branch.rs
```diff
@@ -0,0 +1,269 @@
...
+        txt
+    }
+
+    ///
+    fn switch_to_selected_branch(&self) {
+        checkout_branch(
+            "./",
+            &self.branch_names[self.selection as usize].reference,
+        )
+        .expect("Failed to checkout branch, does the branch exist?");
```
extrawurst, line 264: this should send an error message box instead of crashing

src/components/select_branch.rs
WizardOhio24: This is the same as help, if the screen is impractically small it takes up the whole screen
smallScreen

src/app.rs
WizardOhio24: I wasn't sure if this fits here, should it not be a git async command?

src/components/select_branch.rs
WizardOhio24: This shouldn't happen because the branches update with git (i.e if a branch is changed outwith gitui, it will be reflected in gitui), so gitui would have to fall completely out of sync with the underlying git repo for this error to occur and I imagine there will be other errors occurring before that.  But I'll make an error box for it anyway.

src/components/select_branch.rs
WizardOhio24: This will not, good point.

src/components/select_branch.rs
extrawurst: ok fine by me

extrawurst:
btw this is how it looks for me:
image

there seems to be something wrong 🤔

src/components/select_branch.rs
extrawurst: also true

src/app.rs
extrawurst: yeah if we make it async this would go away.

but before we make it async it would be good to figure out how expensive it is in giant repos and if there is a way to figure out cheaply if there are new local branches or if the last commit (the commit msg previewed in the list) changed.

asyncgit/src/sync/branch.rs
```diff
@@ -28,6 +29,85 @@ pub(crate) fn get_branch_name(repo_path: &str) -> Result<String> {
...
+    pub is_head: bool,
+}
+
+/// TODO make this cached
+/// Used to return only the nessessary information for displaying a branch
+/// rather than an iterator over the actual branches
+pub fn get_branches_to_display(
+    repo_path: &str,
+) -> Vec<BranchForDisplay> {
+    if let Ok(cur_repo) = utils::repo(repo_path) {
```
extrawurst, line 52: please add `scope_time!("get_branches_to_display");` here to be able to log the runtime of this to be able to check how it performs in giant repos

extrawurst:
@WizardOhio24 where are we on this one? since you are already working on the next PR :D

WizardOhio24:
I'm still working on this, I was just waiting for tui to be bumped so the styles could be fixed/implemented.

src/components/select_branch.rs
WizardOhio24: On further inspection, this can handle unicode boundaries because each element is a utf8 character, so the truncate just truncates the number of characters, not the number of bytes.

WizardOhio24:
This currently has a thing(bug?) where it automatically stages all files which have been changed, this isn't a bug, but I don't think it's what the user would expect.  Otherwise this is complete, I think.

extrawurst:
> This currently has a thing(bug?) where it automatically stages all files which have been changed, this isn't a bug, but I don't think it's what the user would expect. Otherwise this is complete, I think.

How do regular git GUIs handle it when you switch branches while having file changes in the working dir?

extrawurst:
How about refusing to switch branches if you have uncommitted changes?

WizardOhio24:
I think it's a bug, it stages and unstages changes which do and undo each other, I'll try to do it the same way git does a checkout.

extrawurst:
@WizardOhio24 libgit2 is very mature, I would be surprised if switching-branches has such a significant bug. if so we should create an issue

WizardOhio24:
It's not a bug with libgit2, I meant the way you would do it in git (e.g `git checkout master`) is not the same as the way it is done in libgit2, or, at least, the equivalent commands don't reach the same result.  I was unsure if there was a bug with my implementation or if this is just what happens when you checkout branches, but I think my implementation is incorrect.

extrawurst:
@WizardOhio24 did you remove the *last commit msg* column?
image

WizardOhio24:
Everything should work now.  The checkout displays an error message if there is uncommitted changes on the current branch, as you suggested (and the way git does it (I think)).  I'm not sure how long the branch names need to be so I've made them cutoff at 15 characters, the commit hash and message take up the remaining space.

extrawurst:
I have reverted my changes and still I cannot checkout:
image

WizardOhio24:
> I have reverted my changes and still I cannot checkout:
> image

What happens if you exit gitui and run `git status`.  Are there 4 changes shown?

extrawurst:
@WizardOhio24 nope, tried that

extrawurst:
looks like we need more unit tests

asyncgit/src/sync/branch.rs
```diff
@@ -2,8 +2,10 @@
 
 use crate::{
     error::{Error, Result},
-    sync::utils,
+    sync::{reset_workdir, utils},
```
extrawurst, line 5: this throws a unused import warning

asyncgit/src/sync/branch.rs
```diff
@@ -28,6 +30,72 @@ pub(crate) fn get_branch_name(repo_path: &str) -> Result<String> {
...
+pub fn checkout_branch(
+    repo_path: &str,
+    branch_ref: &str,
+) -> Result<()> {
+    // This defaults to a safe checkout, so don't delete anything that
+    // hasn't been committed or stashed, in this case it will Err
+    let repo = utils::repo(repo_path)?;
+    if repo.statuses(None)?.is_empty() {
+        repo.set_head(branch_ref)?;
+        repo.checkout_head(Some(
```
extrawurst, line 88: what happens if we fail here ? do we undo the `set_head` change?

src/components/select_branch.rs
```diff
@@ -0,0 +1,321 @@
...
+    pub fn open(&mut self) -> Result<()> {
+        self.update_branches()?;
+        self.show()?;
+
+        Ok(())
+    }
+
+    ////
+    pub fn update_branches(&mut self) -> Result<()> {
+        self.branch_names = Self::get_branch_names()?;
```
extrawurst, line 199: since in this version of the PR we are calling this as *sync* lets call it only when opening the dialog and not on every update. 
if you want add another command to explicitly refresh the list. your call.

asyncgit/src/sync/branch.rs
```diff
@@ -28,6 +30,72 @@ pub(crate) fn get_branch_name(repo_path: &str) -> Result<String> {
...
+    }
+    Ok(branches_for_display)
+}
+
+/// Modify HEAD to point to a branch then checkout head, does not work if there are uncommitted changes
+pub fn checkout_branch(
+    repo_path: &str,
+    branch_ref: &str,
+) -> Result<()> {
+    // This defaults to a safe checkout, so don't delete anything that
```
extrawurst, line 83: `scope_time` missing here aswell

src/app.rs
extrawurst: see my other comment. we should not make it update

src/components/select_branch.rs
WizardOhio24: What if a branch is created while viewing the branch list?  Then it would have to be closed and reopened to update which is just an inconvenience.

extrawurst: first, I consider this an edge case, you can create a branch in here now and we can make sure to refresh then.
second, this stops being necessary once we can run it async.

WizardOhio24:
> I have reverted my changes and still I cannot checkout:
> image

I can't reproduce this, I don't even understand how this error occurs because the code is running the equivalent of `git status` so whatever git status shows should be the same as in the code.  What is the output of `git status` for this?

extrawurst:
So I checkout your branch and then try to switch to master using the new panel:
Screenshot 2020-10-09 at 23 24 49

then I get the error:
Screenshot 2020-10-09 at 23 24 55

`git status` is regular:
Screenshot 2020-10-09 at 23 25 31

WizardOhio24:
Fixed, it was not ignoring files in .gitignore

asyncgit/src/sync/branch.rs
```diff
@@ -97,7 +102,7 @@ pub fn checkout_branch(
         Ok(())
     } else {
         Err(Error::Generic(
-            format!("Cannot change branch. There are unstaged/staged changes which have not been committed/stashed. There is {:?} changes preventing checking out a different branch.",  repo.statuses(None)?.len() ),
+            format!("Cannot change branch. There are unstaged/staged changes which have not been committed/stashed. There is {:?} changes preventing checking out a different branch.",  repo.statuses(None)?.get(0).unwrap().path().unwrap() ),
```
extrawurst, line 105: this looks wrong now

WizardOhio24: I was trying to figure out what was going wrong

asyncgit/src/sync/branch.rs
```diff
@@ -84,7 +84,12 @@ pub fn checkout_branch(
     // hasn't been committed or stashed, in this case it will Err
     let repo = utils::repo(repo_path)?;
     let cur_ref = repo.head()?;
-    if repo.statuses(None)?.is_empty() {
+    if repo
+        .statuses(Some(
+            git2::StatusOptions::new().include_ignored(false),
```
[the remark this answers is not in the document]
WizardOhio24: This does include them by default, see https://docs.rs/git2/0.13.11/git2/enum.StatusShow.html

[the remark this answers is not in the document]
WizardOhio24: > IndexAndWorkdir
>The default, this roughly matches git status --porcelain regarding which files are included and in what order.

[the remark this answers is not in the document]
extrawurst: ok cool 👍

extrawurst:
@WizardOhio24 any idea why the last commits do net get run in CI anymore?

src/components/select_branch.rs
extrawurst: perfect

WizardOhio24:
See https://www.githubstatus.com/, I think.

asyncgit/src/sync/branch.rs
extrawurst: @WizardOhio24 do you want to revert this to `len`?

extrawurst:
@WizardOhio24 are you going to write some tests for `checkout_branch` and `get_branches_to_display` ?

extrawurst:
@WizardOhio24 lets create a followup PR with some tests, I want to get this out for other to play around with. thanks for all the work you put into this❤️

Issue #304, opened 2020-10-04
Push fails with username not found.


edfloreshz:
**Pushing to a repository fails**
I made a small modification to a file to test gitui, after I made a commit and tried to push I got a backtrace error.

**To Reproduce**
1. Modify a file
2. Commit (or don't, happens either way)
3. Push
4. See the error

**Expected behavior**

**Screenshots**
![Screenshot_20201004_032915](https://user-images.githubusercontent.com/22224438/95012974-da244780-05f1-11eb-8c30-0b778055c6a3.png)

**Context (please complete the following information):**
 - Arch Linux 
 - GitUI Version [0.10.1]
 - Rust version: [1.46]

**Additional context**
I don't know if this is an in-progress feature, if it is, feel free to close the issue.

extrawurst:
Yeah *pushing* is indeed in-progress but either way an easy way to keep it from crashing is to use the methods we already use to show errors in other places (example: https://github.com/extrawurst/gitui/issues/108)

but the question is: do you actually have not username defined ?

edfloreshz:
If by having a username defined you mean having it on git config, then yes. In case there's an internal git credentials keychain in `gitui`, then no. 

Do you mind if I try to handle the UB in this issue? Might be a good way to start contributing to your project for the first time 🧑🏻‍💻.

extrawurst:
Sure take it :)
Yeah I mean the gitconfig, gitui does not handle user credentials directly.

BTW. this is no *undefined behaviour* - its a very well defined *panic* :P

edfloreshz:
Oh okay, sorry, handle the crash then :+1:  
Thank you!

edfloreshz:
@extrawurst Unrelated to this issue, I have a question.
Is the `push` option supposed to be available even when no commit has been made?

extrawurst:
> Is the `push` option supposed to be available even when no commit has been made?

actually no, it could be shown in a `disabled` state. but let's create another issue for that

extrawurst:
fixed in #305

extrawurst:
@edfloreshz do you mind looking into #353 to see if this allows you to push to the repo in question?

edfloreshz:
Sure! 👍🏻

PR #305, opened 2020-10-04
Handled crash when git2 fails to obtain username from URL

edfloreshz:
I modified `remote_callbacks( )` to return a `Result<RemoteCallbacks<'a>>` with an error in case the `credentials` callback fails, I also matched every occurrence of the `remote_callbacks( )` function to handle the possible scenarios.

An error is now correctly displayed when this occurs.
Screenshot_20201004_131058

Linked to [#304](https://github.com/extrawurst/gitui/issues/304)

### DIFF
--- a/asyncgit/src/sync/remotes.rs
+++ b/asyncgit/src/sync/remotes.rs
@@ -1,13 +1,13 @@
 //!
 
+use super::CommitId;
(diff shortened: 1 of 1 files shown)


extrawurst:
Thank 🙏

Issue #226, opened 2020-08-07
Mouse support in GITUI


Node0:
**Add TUI mouse support (like exists in Midnight Commander)**
It would make browsing through commit history, and other searching operations convenient if the mouse cursor could access them.

**There is no alternative git client implemented as a TUI, let alone any with mouse support**

extrawurst:
mouse support is not really a focus for me, gitui's philosophy is to make keyboard control super easy and fast for everyone to grasp. albeit if someone comes forward and provides a PR I wouldn't block it

aTable:
@Node0  For what it's worth, [lazygit](https://www.youtube.com/watch?v=F7TCWUO9vt4) has mouse support and might be worth a try?

pm100:
been investigating this, for full support we need tui mods. I raised an issue with them but no response.

https://github.com/fdehau/tui-rs/issues/397

PR #306, opened 2020-10-04
Support mouse scrolling

WizardOhio24:
Implements mouse scrolling for any lists.

Linked to #226


WizardOhio24:
I've set clippy to allow that the function has too many lines, I can split it up if you want but it seems that all the event stuff should be handled in event.

extrawurst:
> I've set clippy to allow that the function has too many lines, I can split it up if you want but it seems that all the event stuff should be handled in event.

that's fine for now, a potential refactor of this should go into a different issue anyway

extrawurst:
I am not sure about this yet. if we do this we should do it right and check the bounds of the mouse and scroll in whatever element the mouse currently is. right now it scrolls whatever element is currently in the key-based focus

WizardOhio24:
That would require a lot of work.  TUI does not have mouse support so that would have to be implemented first then when the scroll button is pressed/scrolled it can switch focus to the one under the mouse then scroll.  
But I don't think it needs to focus to what's under the mouse.  On windows, I think you can click on a window then no matter where the mouse is, when the scroll is used, the window which is currently selected is scrolled (I'm on manjaro, so I'm not sure).  So, from a windows perspective, I think the PR implementation would be the expected implementation.

extrawurst:
> TUI does not have mouse support

not sure what you mean. 
lets consider the `Diff` for now: you already capture the mouse now we only need to use the `col` and `row` to check if we are in bounds of eg. the `Diff`. we can easily do this by not only saving the `current_size` on draw but the entire `Rect`. is `col`/`row` in bounds of this rect the scroll happened on Diff and we can scroll the `scroll_top` without having to even focus this component.

WizardOhio24:
Ah, Ok.  I guess you could implement it like that.  What I meant was that you cannot use the mouse to select a list item in TUI, but I guess that it probably wouldn't be that hard just to find out if the mouse is in a box.  
I think still that there should be an option, so the user can have this behavior you're describing (which I prefer) or so the user must click on the box then scroll to scroll it (and if they move their mouse out of the box, it will still scroll the clicked box when they scroll).

WizardOhio24:
With this PR, I think I will come back to it after more mouse handling is implemented.  This adds scrolling on the box selected by the keys but not the mouse(mouse selection is not implemented) and at the moment mouse selection is not a priority for me(This is after all a TUI), I did like being able to select with the keys and scroll with the mouse though. 

But before closing this PR, currently there is glitchy scrolling in GitUI, this would fix it, then further down the line an option could be implemented to turn on or off mouse scrolling.  I am happy to make any edits to the PR for mouse scrolling, but I don't think I'm going to go as far as mouse selection.

WizardOhio24:
I think I'll continue with Mouse Selection, but not in this PR.  This PR will only deal with Mouse Scrolling on the selected box, but I might put in a PR in the future on Full Mouse Selection.

extrawurst:
@WizardOhio24 can you look into updating the branch to current master and fix the format?

WizardOhio24:
I think this is now ready to be reviewed/merged.

src/main.rs
```diff
@@ -167,6 +168,7 @@ fn main() -> Result<()> {
 fn setup_terminal() -> Result<()> {
     enable_raw_mode()?;
     io::stdout().execute(EnterAlternateScreen)?;
+    io::stdout().execute(EnableMouseCapture)?;
```
extrawurst, line 171: I think we need to do the opposite using `DisableMouseCapture` too

WizardOhio24: ? That line allows capturing the scrolling properly, it's needed otherwise the scrolling won't work (or might work in a glitchly way), the default is DisableMouseCapture.  Doing the opposite would mean all the scrolling code wouldn't work, right?

PR #309, opened 2020-10-06
Draw scrollbar for long commit messages

### LINKED ISSUE #308: scrollbar in commit msg
we have scrollbar in diff now (see #204), so it would be nice to use it here too:
Screenflick Movie 78

timaliberdov:
Closes #308 

Implements drawing a scrollbar when the commit message is too long. Also, I changed the minimum area height for the scrollbar to 4 so that the scrollbar would be drawn in the default terminal window (80x25):

scrollbar_in_commit_msg


extrawurst:
@timaliberdov thanks man!❤️

Issue #32, opened 2020-04-23
Interactive rebase


Geobert:
**Is your feature request related to a problem? Please describe.**
I'm also using Fork, and the sudden change of policy doesn't really suit me so I was looking for an alternative. The feature I use the most was interactive rebase so here is my feature request :)

**Describe the solution you'd like**
Ability to have a list of commit that I can act upon (squash, reword etc…)

**Describe alternatives you've considered**
I'm not really good at git CLI, always used a UI so if gitui could become a fully featured Git UI, that would be awesome.

extrawurst:
Hi @Geobert thanks for your interest in this project, Interactive Rebase is absolutely on my list of features for the commit log view (#2)

naturallymitchell:
related:  [Git Interactive Rebase Tool](https://github.com/MitMaro/git-interactive-rebase-tool)

extrawurst:
A great example how to make it a nice UX

zQueal:
Is there any planned movement on this at all? Really loving gitui, but I need to keep lazygit around as well because it makes rebasing and fixup commits very easy. Would love to just be able to use gitui for it.

extrawurst:
It is planned, yes

extrawurst:
depends on upstream lack of rebase support: https://github.com/libgit2/libgit2/issues/6332

extrawurst:
inspiration: https://github.com/MitMaro/git-interactive-rebase-tool

hyiltiz:
Would it be possible to workaround the limitations of `libgit2` by relying on https://github.com/MitMaro/git-interactive-rebase-tool as a 3rd party dependency, if it exposes the key functionality as somewhat stable API?

tomasol:
To keep things simple, `gitui` could simply invoke `git rebase <commit> --interactive` . Depending on the user's git configuration, it would invoke a text editor or a gui like `git-interactive-rebase` or perhaps even `gitui` once it gains this feature.

tkr-sh:
I'm probably going to work on it in the coming weeks. This doesn't seems that hard to do something similar as https://github.com/MitMaro/git-interactive-rebase-tool

zac1wz:
It has been over 6 years already, any updates?

PR #312, opened 2020-10-07
Support reword in log tab & refactor external editor calls

WizardOhio24:
No more do we have to go though an interactive rebase in which the exact commit reference must be known, where we then must choose 'reword' on the correct commit.

This PR provides the reword functionality on the log(commit list) tab, simply press 'r' on the commit, type in the new message and press enter to change it.

This does not do an interactive rebase in git2rs, rather a slightly more complex rebase operation which achieves the same thing (there is no such thing as an interactive rebase in git2rs/libgit2)
This does not work on the first commit because it has no parent.

Linked to #32


asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,143 @@
...
+
+/// Changes the commit message of a commit with a specified oid
+///
+/// While this function is most commonly associated with doing a
+/// reword opperation in an interactive rebase, that is not how it
+/// is implimented in git2rs
+///
+/// This is dangerous if this errors, as the head will be detached so this should
+/// always be wrapped by another function which aborts the rebase if something goes worng
+pub fn reword(
```
extrawurst, line 53: I don't think we should expose this as `pub`

asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,143 @@
...
+    };
+
+    let commit_to_change = if let Some(pc_oid) = parent_commit_oid {
+        // Need to start at one previous to the commit, so
+        // next point to the actual commit we want to change
+        repo.find_annotated_commit(pc_oid)?
+    } else {
+        return Err(Error::NoParent);
+        // Would the below work?
+        // repo.find_annotated_commit(commit_oid)?
```
extrawurst, line 76: are you still investigating this?

asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    // Find the head branch
+    for b in repo.branches(None)? {
+        let branch = b?.0;
+        if branch.is_head() {
+            cur_branch_ref = Some(String::from(
+                branch
+                    .get()
+                    .name()
+                    .expect("Branch name is not valid utf8"),
```
extrawurst, line 90: why panic instead of returning the error using `?`? We have `Utf8Error` for this in error

asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,143 @@
...
+            Some(&mut RebaseOptions::default()),
+        )?;
+
+        let mut target;
+
+        rebase.next();
+        if parent_commit_oid.is_none() {
+            return Err(Error::NoParent);
+        } else {
+            target = rebase.commit(None, &sig, Some(message))?; //Some(message))?;
```
extrawurst, line 119: what's with the commented code?

asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,143 @@
...
+        // Set target to top commit, don't know when the rebase will end
+        // so have to loop till end
+        while rebase.next().is_some() {
+            target = rebase.commit(None, &sig, None)?;
+        }
+        rebase.finish(None)?;
+
+        // Now override the current branch
+        repo.branch(
+            &cur_branch_name.expect("Couldn't unwrap branch name"),
```
extrawurst, line 131: see above

asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,143 @@
...
+            &repo.find_commit(target)?,
+            true,
+        )?;
+
+        // Reset the head back to the branch then checkout head
+        repo.set_head(
+            &cur_branch_ref.expect("Couldn't unwrap branch name"),
```
extrawurst, line 138: see above

src/components/reword.rs
```diff
@@ -0,0 +1,151 @@
...
+            commit_id: None,
+            key_config,
+        }
+    }
+
+    ///
+    pub fn open(&mut self, id: CommitId) -> Result<()> {
+        self.commit_id = Some(id);
+        let commit = sync::get_commit_details(CWD, id)?;
+        self.input.set_text(commit.message.expect("").combine());
```
extrawurst, line 115: handle as `Result`

extrawurst (changes requested):
Thanks for looking into this!❤️

asyncgit/src/sync/rebase.rs
WizardOhio24: For now, I don't think this needs to handle the first commit.  It is very rare that someone would want to go back to the very start and change the very first commit.

extrawurst: ok

WizardOhio24:
This does work, but it doesn't keep it's commit selection prior to the change, it just goes to the top.

WizardOhio24:
Fixed

asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    // Find the head branch
+    for b in repo.branches(None)? {
+        let branch = b?.0;
+        if branch.is_head() {
+            cur_branch_ref = Some(String::from(
+                branch
+                    .get()
+                    .name()
+                    .expect("Branch name is not valid utf8"),
```
extrawurst, line 24: we have `Utf8Error` for this

asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,146 @@
+//!
+
+use super::commit::signature_allow_undefined_name;
+use crate::{error::Error, error::Result, sync::utils};
+use git2::{Oid, RebaseOptions};
+
+/// This is the same as reword, but will abort and fix the repo if something goes wrong
+pub fn reword_safe(
```
extrawurst, line 8: we should cover this with unittests

src/components/reword.rs
```diff
@@ -0,0 +1,154 @@
...
+        visibility_blocking(self)
+    }
+
+    fn event(&mut self, ev: Event) -> Result<bool> {
+        if self.is_visible() {
+            if self.input.event(ev)? {
+                return Ok(true);
+            }
+
+            if let Event::Key(e) = ev {
```
extrawurst, line 65: the current commit msg popup allows opening an external editor (`vi`) to edit the msg, we should have the same option here

extrawurst (changes requested):

asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,188 @@
...
+            ))?);
+            top_branch_commit = Some(repo.find_annotated_commit(
+                branch.get().peel_to_commit()?.id(),
+            )?);
+            break;
+        }
+    }
+
+    if let (
+        Some(top_branch_commit),
```
extrawurst, line 100: let's make this a single `Option<(commit,ref,name)>`

asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,188 @@
...
+}
+
+/// Changes the commit message of a commit with a specified oid
+///
+/// While this function is most commonly associated with doing a
+/// reword opperation in an interactive rebase, that is not how it
+/// is implimented in git2rs
+///
+/// This is dangerous if this errors, as the head will be detached so this should
+/// always be wrapped by another function which aborts the rebase if something goes worng
```
extrawurst, line 53: typo: "wrong"

asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,188 @@
+//!
+
+use super::commit::signature_allow_undefined_name;
+use crate::{error::Error, error::Result, sync::utils};
+use git2::{Oid, RebaseOptions};
+
+/// This is the same as reword, but will abort and fix the repo if something goes wrong
+pub fn reword_safe(
```
extrawurst, line 8: lets call this `reword` and the internal one `reword_internal` or something

asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,188 @@
...
+    repo_path: &str,
+    commit_oid: Oid,
+    message: &str,
+) -> Result<()> {
+    let repo = utils::repo(repo_path)?;
+    let mut cur_branch_ref = None;
+
+    // Find the head branch
+    for b in repo.branches(None)? {
```
extrawurst, line 17: we already do *finding head branch* in `get_branch_name` lets extract an internal `get_head_branch` for this

asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,188 @@
...
+            &repo.find_commit(target)?,
+            true,
+        )?;
+
+        // Reset the head back to the branch then checkout head
+        repo.set_head(&cur_branch_ref)?;
+        repo.checkout_head(None)?;
+        return Ok(());
+    }
+    // Repo is not on a branch, possibly detached head
```
extrawurst, line 141: would be nice to early out instead of moving into the long if above

src/app.rs
```diff
@@ -58,6 +66,7 @@ pub struct App {
     theme: SharedTheme,
     key_config: SharedKeyConfig,
     input: Input,
+    cur_editor_visible: EditorVisible,
```
extrawurst, line 69: I think we should put this into one variable together with `file_to_open`. meaning that we request the external editor to open using `InternalEvent::OpenExternalEditor` and passing the optional path to the file we also pass the editor that requested it so that we know where to return to afterwards.

so that `file_to_open` becomes something like `external_editor: Option<{Path: Option<String>, source: EditorSource}>`

src/app.rs
```diff
@@ -45,6 +52,7 @@ pub struct App {
     external_editor_popup: ExternalEditorComponent,
     push_popup: PushComponent,
     tag_commit_popup: TagCommitComponent,
+    reword_popup: RewordComponent,
```
extrawurst, line 55: let's rather extend the current existing `CommitComponent`

extrawurst (changes requested):

asyncgit/src/sync/rebase.rs
WizardOhio24: Each of these can fail and return None, so this would just add a line where each are checked for if they are `some` and put in that format, `Option<commit, ref, name>`, which means it just moves the checks.  What would adding this achieve/How would this be added?

asyncgit/src/sync/rebase.rs
WizardOhio24: I think `reword` and `reword_safe` are better names because they better represent what they do, but I'm happy to change it. With `reword_internal` and `reword`, I understand what you are trying to say but I don't think it's that kind of relationship, `reword` does exactly that, but it doesn't do anything else(if it errored the user has to face the consequences), while `reword_safe` does `reword` but catches any errors and fixes the repo for the user.  This is usually what the user wants, but this is not just rewording, this is rewording and fixing if there are errors.  Do you still want this to be changed?

asyncgit/src/sync/rebase.rs
WizardOhio24: What do you mean?

src/app.rs
WizardOhio24: How would this work?  This would be like merging two components with separate commands, titles, everything,  and use a lot of if's to check which the user is doing.  Why would this be added to commit?  Plus this is rebasing, not like `commit amend`, so I think these should be kept separate, so as to reflect best the way git works.

src/app.rs
WizardOhio24: In this PR?  I can make another PR for that after this one, "Refactor file_to_open".  Should be a simple fix.

asyncgit/src/sync/rebase.rs
WizardOhio24: That's what all the if let Some()...s were for, they would be None if the user was not on a branch, so this does already early out, is that what you're talking about?

src/app.rs
extrawurst: makes sense in this PR

asyncgit/src/sync/branch.rs
```diff
@@ -31,6 +31,32 @@ pub(crate) fn get_branch_name(repo_path: &str) -> Result<String> {
     Err(Error::NoHead)
 }
 
+/// Gets the current branch the user is on.
+/// Returns none if they are not on a branch and Err
+/// if there was a problem finding the branch
+pub fn get_cur_branch(
```
extrawurst, line 37: can be `pub(crate)` since we do not expect the user to use `git2-rs` directly

asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,178 @@
...
+            Err(e)
+        }
+    }
+}
+
+/// Changes the commit message of a commit with a specified oid
+///
+/// While this function is most commonly associated with doing a
+/// reword opperation in an interactive rebase, that is not how it
+/// is implimented in git2rs
```
extrawurst, line 40: typo: `implemented`

asyncgit/src/sync/rebase.rs
extrawurst: my primary concern is to keep a reasonable public API. `reword_safe` makes only sense if you look at `reword` which not even publicly exposed (which I think is the right decision). I am open for better names but my preferences is to have verbose naming for the internal one and clear and short one for the public facing one.

src/app.rs
WizardOhio24: Ok, I'm just trying to keep my PRs smaller and more contained on individual features after the... the branch upstream one(337), and I felt that this is a seperate thing, like a refactor of the external editor, but I can do it in this one, no problem.

asyncgit/src/sync/rebase.rs
extrawurst: a) this is how you can combine `Options:
```
let a  = Some(1);
let b  = Some(2);
let ab = a.zip(b);
```

b) please use the new `bytes2string` helper to make the upper code less verbose for extracting branch name and ref name

asyncgit/src/sync/commit.rs
```diff
@@ -33,7 +33,7 @@ pub fn amend(
 /// Wrap Repository::signature to allow unknown user.name.
 ///
 /// See <https://github.com/extrawurst/gitui/issues/79>.
-fn signature_allow_undefined_name(
+pub fn signature_allow_undefined_name(
```
extrawurst, line 36: why do we need to make this public? again this is of no use for the user of this library since we expect them not to directly use `git2-rs`

asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,178 @@
...
+        repo.find_annotated_commit(pc_oid)?
+    } else {
+        return Err(Error::NoParent);
+    };
+    let mut top_branch_commit = None;
+    let mut cur_branch_ref = None;
+    let mut cur_branch_name = None;
+
+    // Find the head branch
+    for b in repo.branches(None)? {
```
extrawurst, line 72: can we use the new `get_cur_branch` here?

extrawurst (changes requested):

asyncgit/src/sync/rebase.rs
WizardOhio24: Turns out you don't need options

asyncgit/src/sync/rebase.rs
WizardOhio24: I am assuming the most recent code resolves this? In that if the user is not on a branch, it doesn't start a rebase.

asyncgit/src/sync/commit.rs
WizardOhio24: This is async git, and there are other pub functions there, e.g `commit`, `amend`, but ok, it can be pub crate.

src/app.rs
WizardOhio24: I have made a different PR for it, I'm unsure about how you see this implemented, but I'll try to get the implementation right.

WizardOhio24: Continued in #428

WizardOhio24:
With this, should the character count in the reword input box be shown?  e.g, if the user wants to know if the commit message length?

WizardOhio24:
It's true for commit message, I'll set the show character count to true.

asyncgit/src/sync/branch.rs
```diff
@@ -30,6 +30,32 @@ pub(crate) fn get_branch_name(repo_path: &str) -> Result<String> {
...
+        let branch = b?.0;
+        if branch.is_head() {
+            return Ok(Some(branch));
+        }
+    }
+    Ok(None)
+}
+
+/// Convenience function to get the current branch reference
+pub fn get_cur_branch_ref(repo_path: &str) -> Result<Option<String>> {
```
extrawurst, line 49: please use `get_head_refname` instead

src/components/reword.rs
```diff
@@ -0,0 +1,224 @@
...
+
+        Ok(())
+    }
+
+    /// After an external editor has been open,
+    /// this should be called to put the text in the
+    /// right place
+    pub fn show_editor(&mut self) -> Result<()> {
+        const COMMIT_MSG_FILE_NAME: &str = "COMMITMSG_EDITOR";
+        //TODO: use a tmpfile here
```
extrawurst, line 150: lets fix that before we merge, adding a new `TODO` will never be fixed otherwise :)

src/strings.rs
```diff
@@ -726,6 +732,30 @@ pub mod commands {
...
     }
+    pub fn reword_commit_confirm_msg(
+        key_config: &SharedKeyConfig,
+    ) -> CommandText {
+        CommandText::new(
+            format!(
+                "Reword [{}]",
+                key_config.get_hint(key_config.enter),
+            ),
+            "tag commit",
```
extrawurst, line 743: "tag"? looks like a copy/paste mistake

extrawurst (changes requested):

src/components/reword.rs
WizardOhio24: I think the fix was in #428, or related to it, I'll move that commit over to here, can't remember exactly what #428 does.

WizardOhio24: I'll fix the one in commit while I'm at it, that's where this was originally from.

WizardOhio24:
Updated to master.

extrawurst:
superseded by #672

WizardOhio24:
This has nothing to do with #672?  This is about rewording commits in the log tab, nothing to do with searching.  
Perhaps you meant to close #506?  (Feel free, it has been superceded).

extrawurst:
was too late, of course I meant to close #506

src/app.rs
```diff
@@ -35,7 +35,14 @@ use tui::{
     Frame,
 };
 
-/// the main app type
+/// Used to determine where the user should
+/// be put when the external editor is closed
+pub enum EditorSource {
```
extrawurst, line 40: should be put into the `externaleditor.rs`

src/app.rs
```diff
@@ -64,10 +72,10 @@ pub struct App {
     theme: SharedTheme,
     key_config: SharedKeyConfig,
     input: Input,
+    external_editor: Option<(Option<String>, EditorSource)>,
```
extrawurst, line 75: this should be a struct instead of tuple to help express the intent of this. something like `OpenEditorState`

src/app.rs
```diff
@@ -528,14 +553,23 @@ impl App {
                     .insert(NeedsUpdate::ALL | NeedsUpdate::COMMANDS);
             }
             InternalEvent::Update(u) => flags.insert(u),
-            InternalEvent::OpenCommit => self.commit.show()?,
+            InternalEvent::OpenCommit => {
+                self.external_editor =
```
extrawurst, line 557: I am not sure why we have to do this now? 
doesn't this trigger `show_editor` now every time we just open the regular commit popup?

src/components/externaleditor.rs
```diff
@@ -187,3 +193,34 @@ impl Component for ExternalEditorComponent {
         Ok(())
     }
 }
+
+pub fn show_editor(with_text: Option<&String>) -> Result<String> {
+    let temp_file = NamedTempFile::new()?;
```
extrawurst, line 198: by now we intentionally use a specific filename to edit commit msgs in so that vim knows how to highlight it, we should keep doing this (see `COMMIT_EDITMSG`)

src/queue.rs
```diff
@@ -69,7 +72,7 @@ pub enum InternalEvent {
     ///
     SelectBranch,
     ///
-    OpenExternalEditor(Option<String>),
+    OpenExternalEditor(Option<String>, EditorSource),
```
extrawurst, line 75: use a struct like mentioned above

src/app.rs
```diff
@@ -555,10 +589,12 @@ impl App {
                 self.inspect_commit_popup.open(id, tags)?;
                 flags.insert(NeedsUpdate::ALL | NeedsUpdate::COMMANDS)
             }
-            InternalEvent::OpenExternalEditor(path) => {
+            InternalEvent::OpenExternalEditor(
+                _path,
```
extrawurst, line 593: why don't we need those parameters anymore?

asyncgit/src/sync/branch/mod.rs
```diff
@@ -44,6 +44,32 @@ pub(crate) fn get_branch_name_repo(
...
+        let branch = b?.0;
+        if branch.is_head() {
+            return Ok(Some(branch));
+        }
+    }
+    Ok(None)
+}
+
+/// Convenience function to get the current branch reference
+pub fn get_head_refname(repo_path: &str) -> Result<Option<String>> {
```
extrawurst, line 63: `pub(crate)` this seems to only be needed inside `asyncgit`

asyncgit/src/sync/branch/mod.rs
```diff
@@ -44,6 +44,32 @@ pub(crate) fn get_branch_name_repo(
...
+    let repo = utils::repo(repo_path)?;
+    if let Ok(Some(b)) = get_cur_branch(&repo) {
+        return Ok(Some(String::from_utf8(
+            b.get().name_bytes().to_vec(),
```
extrawurst, line 67: we have `bytes2string` for this

asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,163 @@
...
+            Err(e)
+        }
+    }
+}
+
+/// Changes the commit message of a commit with a specified oid
+///
+/// While this function is most commonly associated with doing a
+/// reword opperation in an interactive rebase, that is not how it
+/// is implemented in git2rs
```
extrawurst, line 47: what does that mean? what does it imply?

asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,163 @@
...
+fn reword_internal(
+    repo_path: &str,
+    commit_oid: Oid,
+    message: &str,
+) -> Result<()> {
+    let repo = utils::repo(repo_path)?;
+    let sig = signature_allow_undefined_name(&repo)?;
+
+    let parent_commit_oid = if let Ok(parent_commit) =
+        repo.find_commit(commit_oid)?.parent(0)
```
extrawurst, line 60: what if the commit is a merge commit with multiple parents? please add a unittest for that case

asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,163 @@
...
+    let commit_to_change = if let Some(pc_oid) = parent_commit_oid {
+        // Need to start at one previous to the commit, so
+        // first rebase.next() points to the actual commit we want to change
+        repo.find_annotated_commit(pc_oid)?
+    } else {
+        return Err(Error::NoParent);
+    };
+
+    // If we are on a branch
+    if let Ok(Some(branch)) = get_cur_branch(&repo) {
```
extrawurst, line 76: lets map this into the `Err` we want and `?` it here to early out if not on a branch, this makes the rest indent one less and easier to read

asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,163 @@
...
+    #[test]
+    fn test_reword() -> Result<()> {
+        let file_path = Path::new("foo");
+        let (_td, repo) = repo_init_empty().unwrap();
+        let root = repo.path().parent().unwrap();
+        let repo_path = root.as_os_str().to_str().unwrap();
+
+        File::create(&root.join(file_path))?.write_all(b"a")?;
+        stage_add_file(repo_path, file_path).unwrap();
+        commit(repo_path, "commit1").unwrap();
```
extrawurst, line 138: the new `write_commit_file` makes creating/writing/committing test files more convenient/easier to read now

asyncgit/src/sync/rebase.rs
```diff
@@ -0,0 +1,163 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::sync::{
+        commit, stage_add_file, tests::repo_init_empty,
+    };
+    use std::{fs::File, io::Write, path::Path};
+
+    #[test]
+    fn test_reword() -> Result<()> {
```
extrawurst, line 130: we currently only test the happy path, what if the rename does not go through and the rebase has to be rolled back. we need a test for that aswell

extrawurst (changes requested):
I worked through all of this again and it cannot be merged:

Screenshot 2021-05-27 at 22 19 53

this happens right when I simply open the commit popup, go to external editor and close it again. this is currently broken.

most of the external editor stuff can be most like reverted anyway if you use the `Mode` in the commit popup. actually the entire reword-popup can now be handled inside `CommitComponent`, this would be much easier I think

extrawurst:
going back to draft in the meantime. change it back to review once you think its ready 👍

extrawurst:
closing in favour of https://github.com/extrawurst/gitui/pull/1553

PR #324, opened 2020-10-10
Use helper binary for clipboard access (#262)

### LINKED ISSUE #262: enable clipboard support on linux again

cruessler:
Use `clip` on Windows, `pbcopy` on MacOX, `xclip` on Linux.

Remove dependency on `clipboard`.

Closes #262.

### DIFF
--- a/src/clipboard.rs
+++ b/src/clipboard.rs
@@ -1,33 +1,47 @@
 use anyhow::Result;
-#[cfg(feature = "clipboard")]
-use clipboard::{ClipboardContext, ClipboardProvider};
+use std::io::Write;
+use std::process::{Command, Stdio};
 
-#[cfg(feature = "clipboard")]
-pub fn copy_string(string: String) -> Result<()> {
+fn execute_copy_command(
+    command: &mut Command,
+    string: &str,
+) -> Result<()> {
     use anyhow::anyhow;
 
-    let mut ctx: ClipboardContext = ClipboardProvider::new()
-        .map_err(|e| {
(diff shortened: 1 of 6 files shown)


src/clipboard.rs
```diff
@@ -1,22 +1,51 @@
 use anyhow::Result;
-#[cfg(feature = "clipboard")]
-use clipboard::{ClipboardContext, ClipboardProvider};
+use std::io::Write;
+use std::process::{Command, Stdio};
 
-#[cfg(feature = "clipboard")]
-pub fn copy_string(string: String) -> Result<()> {
+fn execute_copy_command(
+    command: &mut Command,
```
extrawurst, line 6: we can take full ownership of `command` here.

src/clipboard.rs
```diff
@@ -1,22 +1,51 @@
...
+#[cfg(all(feature = "clipboard", target_os = "linux"))]
+pub fn copy_string(string: &str) -> Result<()> {
+    execute_copy_command(
+        Command::new("xclip").arg("-selection").arg("clipboard"),
+        string,
+    )
+}
+
+#[cfg(all(feature = "clipboard", target_os = "macos"))]
+pub fn copy_string(string: String) -> Result<()> {
```
extrawurst, line 40: the function signature need to be the same on each platform, CI currently fails

extrawurst (changes requested):

PR #332, opened 2020-10-11
Delete branch

### LINKED ISSUE #323: delete branch
now that we have the branch list it would be nice to also allow deleting branches.

this is a destructive cmd so we should confirm before executing just like with *reverting files*

WizardOhio24:
Closes #323 

Allows the user to delete a branch by opening the branch list then pressing 'd' on the branch they want to delete.  The user must not be on the branch for the branch to be deleted, if the user is on the branch, an error message is displayed.

### DIFF
--- a/asyncgit/src/sync/branch.rs
+++ b/asyncgit/src/sync/branch.rs
@@ -124,6 +124,8 @@ pub fn delete_branch(
     let mut branch_as_ref = repo.find_reference(branch_ref)?;
     if cur_ref != branch_as_ref {
         branch_as_ref.delete()?;
+    } else {
+        return Err(Error::Generic("You cannot be on the branch you want to delete, switch branch, then delete this branch".to_string()));
     }
     Ok(())
 }
--- a/asyncgit/src/sync/branch.rs
+++ b/asyncgit/src/sync/branch.rs
@@ -121,7 +121,7 @@ pub fn delete_branch(
 
     let repo = utils::repo(repo_path)?;
     let cur_ref = repo.head()?;
-    let branch_as_ref = repo.find_reference(branch_ref)?;
+    let mut branch_as_ref = repo.find_reference(branch_ref)?;
     if cur_ref != branch_as_ref {
         branch_as_ref.delete()?;
     }
--- a/asyncgit/src/sync/branch.rs
+++ b/asyncgit/src/sync/branch.rs
@@ -120,10 +120,10 @@ pub fn delete_branch(
     scope_time!("delete_branch");
 
     let repo = utils::repo(repo_path)?;
-    let cur_ref = repo.head()?;
-    let mut branch_as_ref = repo.find_reference(branch_ref)?;
-    if cur_ref != branch_as_ref {
-        branch_as_ref.delete()?;
+    let branch_as_ref = repo.find_reference(branch_ref)?;
+    let mut branch = git2::Branch::wrap(branch_as_ref);
+    if !branch.is_head() {
+        branch.delete()?;
     } else {
         return Err(Error::Generic("You cannot be on the branch you want to delete, switch branch, then delete this branch".to_string()));
     }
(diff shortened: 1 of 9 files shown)


asyncgit/src/sync/branch.rs
```diff
@@ -107,6 +107,24 @@ pub fn checkout_branch(
     }
 }
 
+/// The user must not be on the branch for the branch to be deleted
+pub fn delete_branch(
```
extrawurst, line 111: please add test

src/components/select_branch.rs
```diff
@@ -113,6 +113,14 @@ impl Component for SelectBranchComponent {
                 true,
                 true,
             ));
+
+            out.push(CommandInfo::new(
+                strings::commands::delete_branch_popup(
+                    &self.key_config,
+                ),
+                true,
```
extrawurst, line 121: we can use this to disable the cmd for the current branch

src/components/select_branch.rs
```diff
@@ -142,6 +150,21 @@ impl Component for SelectBranchComponent {
...
+                        InternalEvent::ConfirmAction(
+                            Action::DeleteBranch(
+                                self.branch_names
+                                    [self.selection as usize]
+                                    .reference
+                                    .clone(),
+                            ),
+                        ),
+                    );
+                    /*self.queue
```
extrawurst, line 164: cleanup?

src/components/select_branch.rs
```diff
@@ -142,6 +150,21 @@ impl Component for SelectBranchComponent {
                         .borrow_mut()
                         .push_back(InternalEvent::CreateBranch);
                     self.hide();
+                } else if e == self.key_config.delete_branch {
+                    self.queue.borrow_mut().push_back(
```
extrawurst, line 154: check to not allow for current branch

src/keys.rs
```diff
@@ -113,6 +114,7 @@ impl Default for KeyConfig {
             copy: KeyEvent { code: KeyCode::Char('y'), modifiers: KeyModifiers::empty()},
             create_branch: KeyEvent { code: KeyCode::Char('c'), modifiers: KeyModifiers::NONE},
             select_branch: KeyEvent { code: KeyCode::Char('b'), modifiers: KeyModifiers::NONE},
+            delete_branch: KeyEvent{code: KeyCode::Char('d'), modifiers: KeyModifiers::NONE},
```
extrawurst, line 117: lets stick to `D` like for *delete stash* and others

src/components/select_branch.rs
WizardOhio24: I'm not sure about this.  If it's implemented like this, pressing delete just seems broken.  I think it's much clearer if an error is displayed saying they cannot delete the current branch because they are on it.  By just stopping the delete button working, the user might not know why the delete button stopped working.

extrawurst: that's why we also disable the cmd in the bottom cmd bar (see other comment above)

src/components/select_branch.rs
```diff
@@ -118,7 +118,25 @@ impl Component for SelectBranchComponent {
                 strings::commands::delete_branch_popup(
                     &self.key_config,
                 ),
-                true,
+                if self
```
extrawurst, line 121: lets make it a helper function and use it also for the `event` part

WizardOhio24: I'm not sure it should be used for the event part, it's fine to disable the button to indicate the user shouldn't do it, but if the button just doesn't do anything (on the current branch) the user might not understand why the delete button just stopped working.

src/components/select_branch.rs
```diff
@@ -113,6 +113,14 @@ impl Component for SelectBranchComponent {
                 true,
                 true,
             ));
+
+            out.push(CommandInfo::new(
+                strings::commands::delete_branch_popup(
+                    &self.key_config,
+                ),
+                self.selection_is_cur_branch(),
```
extrawurst: that seems inverted. the command should be *enabled* when the selection is **not** the current branch, right?

src/components/select_branch.rs
extrawurst: I see your concern, but visualising the cmd disabled and then still allow pressing it is pointless and does not align with other cmds and the `enable` state so far in gitui. 
I am expecting enough general git understanding by the user to grasp that deleting the branch you are standing on is not allowed (quiet literally ^^)

extrawurst:
@WizardOhio24 can you add a test and the last change to `event` then we can get this in

extrawurst:
Nice thank you 🥳

PR #334, opened 2020-10-11
Rename branch

WizardOhio24:
Adds rename branch functionality to the branch list.  The user can now select a branch, press 'r', type in the new name then press enter to rename.  This also works while on the branch being renamed.


extrawurst:
@WizardOhio24 thanks!
how does this behave with the branch renamed being a tracking branch of a remote branch?

WizardOhio24:
It won't rename the remote branch, only the local branch, so if a remote tracking branch had been set up it would still push to that branch.  GitUI doesn't have any control over remote branches, e.g to rename a remote github branch, you would have to go onto github and rename the branch there (then update the remote for the local branch).

extrawurst:
Thanks @WizardOhio24 ❤️

PR #337, opened 2020-10-12
Support upstream branches

WizardOhio24:
This PR implements:
 - [x] Viewing upstreams
 - [x] Viewing which upstream branch is the upstream branch for the selected local branch (if there is one) 
 - [x] Selecting an upstream for a branch
 - [x] Creating a new upstream branch (in memory)(doesn't currently work with push, but that should be a different PR)
 - [x] Adding a new upstream with a name and URL
 - [ ] Removing upstreams

To access the upstreams menu, open the select branch menu, then press 'u' on a branch to set it's upstream or add a remote.
upstreams

I think the upstreams menu should be a number of drop downs for each upstream, so the user can fold all of them up just to see the upstream names, then can unfold them to see the actual branches.  Then to remove an upstream the user can press SHIFT+D on the drop down name.


extrawurst:
@WizardOhio24 can you bring this up to date? would be great to get in for `0.11`

extrawurst:
*removing an upstream* can easily be deferred to a later PR

WizardOhio24:
This can now be merged, I think.  
The create branch function does not work with push(but that's for a different PR), and remove remote does not work(it's not too complicated to implement, the question is how do you make it clear that the user is deleting a remote and not a remote branch(which they cannot do in GitUI)).

WizardOhio24:
Also, it should be made to have relative sizing in the same way as the select branch popup after #363 has been fixed

extrawurst (changes requested):
Thank you for putting the work into such a big feature. I took a first look playing around with the feature. I think it is not easy to understand. I also don't know what primary use case it is catering: create upstream for a local branch? switching the upstream of a local? creating a remote branch? may we should descope this a little bit and reduce the features to what we actually think is a common usecase: 
I think the major one related to upstreams aside from creating a local branch including upstream tracking branch is simply discovering the remote branches and pulling one down and checking it out.

WizardOhio24:
> Thank you for putting the work into such a big feature. I took a first look playing around with the feature. I think it is not easy to understand. I also don't know what primary use case it is catering: create upstream for a local branch? switching the upstream of a local? creating a remote branch? may we should descope this a little bit and reduce the features to what we actually think is a common usecase:
> I think the major one related to upstreams aside from creating a local branch including upstream tracking branch is simply discovering the remote branches and pulling one down and checking it out.

In essence this should be able to replace the `git remote` command.
The use case I was thinking of was when you have multiple upstreams you have to coordinate, e.g, say you are building a web server and there are serval different upstream branches which you'll want to coordinate, like `live-web`, `web-master`, `master`.  When pushing to `master`, you might also want to push to `web-master`, to catch it up (this is equivalent to (fast-forward)merging `web-master` with `master`, then pushing to `web-master`).  Another use case is when you want to change the upstream branch of your local branch (e.g. if the upstream branch has been renamed(e.g renaming a github repo)).  

Adding upstreams is rare(generally only done at the start of development) but should be possible with GitUI.

The create remote branch will save an upstream branch for the next push, so say you have a local branch `local-dev` and you want to push to an upstream branch `develop` which does not exist yet on the git server but will be created once something pushes to it (same as the way github works).  A tracking branch would track the upstream branch `local-dev`, but in this case that is not wanted. 

Discovering remote branches and pulling them down is a use case, perhaps the most common one, but I was trying to cater for all use cases of `git remote`, so GitUI can replace `git remote`.  
Perhaps a command should be implemented to pull a remote branch (in a different PR)? Although, is this not just fetch then checkout, so it doesn't really need any interaction with remotes(rather you only need to interact with local branches), except perhaps switching which upstream to fetch from?

extrawurst:
Yeah my philosophy for adding features to gitui is to focus on relevant use cases first. and each feature needs to be super easy to use and understand, that was the whole idea of creating gitui. this upstream list is not for me. and I also don't want to add features just to have them. gitui does and will for a long time work in tandem with the CLI. it is not a replacement. the ergonomics of whatever we add to gitui need to be justifying the added complexity. I hope you understand this is not personal. I am super anxious to write these lines because I value everyone who takes the time and effort to write a contribution. 

the question is now where do we go from here so that not all is wasted in this PR?

also I think we should take inspiration from git-brunch for how to handle branches: https://github.com/andys8/git-brunch
I like this side-by-side view and I think this is a good way to make it easy to visualise which local branch belongs to which remote one (using some sort of selection sensitive highlighting)

WizardOhio24:
I always thought and will always think that GitUI can one day replace the Git CLI and perhaps become the best and easiest way to interact with Git:  We have different views, fair enough.  

With this other branch view you propose, would you want a different tab for branches then?  With this, the branches could be on the left and upstreams on the right.  The controls should be more intuitive then.  The branch popup in status can stay, but there can also be a branch tab for interacting with branches and (Adding/changing etc.) upstreams.

extrawurst:
> I always thought and will always think that GitUI can one day replace the Git CLI

we certainly have the same vision, no disagreement there. we just have different priorities / philosophies to get there I guess :) every time I am googling how to do something on the CLI or using a GUI like Fork for something I know its not there yet but we will get there

WizardOhio24:
Ok, I think this PR is too big, I seem to have got a bit carried away.  I'll have another go at this, but I'll split it into a number of (smaller) PRs, with the first one probably being to add a new branch tab.  Feel free to close.

WizardOhio24:
Or leave this open, I can continue on this if you want, it's just between one feature rich PR or many small individual feature PRs.

PR #338, opened 2020-10-12
Fix crash on small terminal size when branch list is open

WizardOhio24:
When the terminal is made too small and the branch list is open, the program would crash.  This fixes that so the program just resizes.

### DIFF
--- a/src/components/select_branch.rs
+++ b/src/components/select_branch.rs
@@ -258,12 +258,16 @@ impl SelectBranchComponent {
         width_available: u16,
     ) -> Result<Text> {
         const BRANCH_NAME_LENGTH: usize = 15;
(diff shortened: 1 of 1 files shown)


src/components/select_branch.rs
```diff
@@ -259,22 +259,28 @@ impl SelectBranchComponent {
     ) -> Result<Text> {
         const BRANCH_NAME_LENGTH: usize = 15;
         // total width - commit hash - branch name -"*  " - "..." = remaining width
-        let commit_message_length: usize =
-            width_available as usize - 8 - BRANCH_NAME_LENGTH - 3 - 3;
+        let commit_message_length: usize = (width_available as usize)
+            .saturating_sub(8)
```
extrawurst, line 263: these magic numbers are not self explanatory. can you make named local ones?

PR #349, opened 2020-10-16
feat: add branch selection popup in the log view

### LINKED ISSUE #345: allow opening the new branch popup in log
we have a new cmd in the status view that allows opening the branch popup, I think we should allow it in the log view aswell

alocquet:
fix #345


extrawurst:
thanks ❤️

PR #353, opened 2020-10-19
feat: support https push

### LINKED ISSUE #344: support https push via credentials dialog
right now only ssh push works really smooth.
https-repos will not know what pwd to use in the least and if the checkout url did not contain the username it will also not be found.

we should try to query the username from git-config and open username input if not, afterwards lets open another input for the pwd. in this first iteration let's not worry about storing any of this.

to keep things simple lets reuse `TextInputComponent` we already have even for password in a first iteration and not worry about exchanging everything with `*`. Let's use two popups in a row for *username* and *password* so we can easily adjust for when a username can be found and only ask for the password then.

see also #304

alocquet:
Fix #344

Finally, I use git2 credential help and I retrieve username/password from :
* git credential manager
* url
* git config

I show username textbox only if I can't retrieve it. I clean credentials after push to avoid keeping it in memory.

### DIFF
--- a/asyncgit/src/error.rs
+++ b/asyncgit/src/error.rs
@@ -10,7 +10,7 @@ pub enum Error {
     NoHead,
 
     #[error("git: remote url not found")]
-    NoRemote,
+    UnknownRemote,
 
     #[error("io error:{0}")]
     Io(#[from] std::io::Error),
--- a/asyncgit/src/error.rs
+++ b/asyncgit/src/error.rs
@@ -9,6 +9,9 @@ pub enum Error {
     #[error("git: no head found")]
     NoHead,
 
+    #[error("git: remote url not found")]
+    NoRemote,
+
     #[error("io error:{0}")]
     Io(#[from] std::io::Error),
 
--- a/asyncgit/src/sync/cred.rs
+++ b/asyncgit/src/sync/cred.rs
@@ -20,7 +20,7 @@ pub fn need_username_password(remote: &str) -> Result<bool> {
     let url = repo
         .find_remote(remote)?
         .url()
-        .ok_or_else(|| Error::Generic("No remote URL".to_owned()))?
+        .ok_or_else(|| Error::NoRemote)?
         .to_owned();
     let is_http = url.starts_with("http");
     Ok(is_http)
@@ -34,7 +34,7 @@ pub fn extract_username_password(
     let url = repo
         .find_remote(remote)?
         .url()
-        .ok_or_else(|| Error::Generic("No remote URL".to_owned()))?
+        .ok_or_else(|| Error::NoRemote)?
         .to_owned();
     let mut helper = CredentialHelper::new(&url);
 
--- a/asyncgit/src/sync/cred.rs
+++ b/asyncgit/src/sync/cred.rs
@@ -34,7 +34,7 @@ pub fn need_username_password(remote: &str) -> Result<bool> {
     let url = repo
         .find_remote(remote)?
         .url()
-        .ok_or(Error::NoRemote)?
+        .ok_or(Error::UnknownRemote)?
         .to_owned();
     let is_http = url.starts_with("http");
     Ok(is_http)
@@ -48,7 +48,7 @@ pub fn extract_username_password(
     let url = repo
         .find_remote(remote)?
         .url()
-        .ok_or(Error::NoRemote)?
+        .ok_or(Error::UnknownRemote)?
         .to_owned();
     let mut helper = CredentialHelper::new(&url);
 
--- a/asyncgit/src/sync/cred.rs
+++ b/asyncgit/src/sync/cred.rs
@@ -6,7 +6,7 @@ use crate::error::{Error, Result};
 use crate::CWD;
 
 /// basic Authentication Credentials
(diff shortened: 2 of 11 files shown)


asyncgit/src/sync/cred.rs
```diff
@@ -0,0 +1,70 @@
+//! credentials git helper
+
+use git2::{Config, CredentialHelper};
```
extrawurst, line 3: this new module could be covered with tests, right?

asyncgit/src/sync/remotes.rs
```diff
@@ -165,21 +173,52 @@ fn remote_callbacks<'a>(
...
-        );
 
-        match username_from_url {
-            Some(username) => Cred::ssh_key_from_agent(username),
-            None => Err(GitError::from_str(
-                " Couldn't extract username from url.",
-            )),
-        }
-    });
+    let mut first_call_to_credentials = true;
```
extrawurst, line 177: is there no overlap of options to authenticate and it makes sense to try another one if a try fails? Either way we should reference the docs for this: https://libgit2.org/docs/guides/authentication/

src/components/push.rs
```diff
@@ -47,23 +56,63 @@ impl PushComponent {
...
         }
     }
 
     ///
     pub fn push(&mut self, branch: String) -> Result<()> {
+        self.branch = branch;
+        self.show()?;
+        match need_username_password(DEFAULT_REMOTE_NAME) {
+            Ok(true) => {
+                self.cred = extract_username_password("origin")
```
extrawurst, line 86: shouldn't this also take the const?

extrawurst (changes requested):
Hey @alocquet thanks for working on this❤️! I have made some comments

src/components/push.rs
```diff
@@ -10,8 +11,10 @@ use crate::{
 };
 use anyhow::Result;
 use asyncgit::{
-    AsyncNotification, AsyncPush, PushProgress, PushProgressState,
-    PushRequest,
+    sync::cred::extract_username_password,
+    sync::cred::need_username_password,
```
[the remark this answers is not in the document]
alocquet: I agree

asyncgit/src/sync/cred.rs
alocquet: I agree

asyncgit/src/sync/remotes.rs
alocquet: In fact, I don't see how I can display the cred inputs when I am in this callback, because this callback is in asyncgit module and moreover, I think it needs to be blocking : we need to wait user inputs.
This is why I have extracted username and password before push call, so as to ask creds before.

extrawurst: I know I was not thinking about showing the popup during the callbacks. I just thought that the system looks like we have multiple ways to feed Cred to the callback, right? and the callback let's us know if the server accepted them by calling the callback again. so I was wondering if there might be a case where we would have to try different ways to feed the creds that we have?

alocquet: Ok, I understand. I think that for https auth (with user/password), there are no other way to feed the creds. The callback can only be called to ask the user to fill other creds.
I have checked Cargo implementation which is used as reference in a stackoverflow thread and they use 2nd call to callback to give more precision about error :
https://github.com/rust-lang/cargo/blob/9fb208dddb12a3081230a5fd8f470e01df8faa25/src/cargo/sources/git/utils.rs#L747

extrawurst: ok so it's fine to only allow one callback, return a potential error to the user and on a new try they could change their credentials before the callback start

extrawurst: then let's just link the docs that I mentioned and the SO post in the code as a reference and keep it this way

asyncgit/src/error.rs
```diff
@@ -9,6 +9,9 @@ pub enum Error {
     #[error("git: no head found")]
     NoHead,
 
+    #[error("git: remote url not found")]
+    NoRemote,
```
extrawurst, line 13: `NoRemote` might be misleading. Maybe `UnknownRemote`?

src/components/cred.rs
```diff
@@ -0,0 +1,158 @@
...
+    ui::style::SharedTheme,
+};
+
+///
+pub struct CredComponent {
+    visible: bool,
+    key_config: SharedKeyConfig,
+    input_username: TextInputComponent,
+    input_password: TextInputComponent,
+    pub cred: BasicAuthCredential,
```
extrawurst, line 24: let's make this `private` and use a setter

extrawurst (approved):

extrawurst:
@alocquet sorry it takes me so long to get this merged. I hope to have time tomorrow to give it a spin and then merge 🙈

alocquet:
Don't worry :)

extrawurst:
Awesome! Works for me. Just gave it a spin! Thanks for working on this❤️!
Maybe its a little more ergonomic to not use the huge default TextInputComponent Size but that's a battle for another day. also using `"".to_string()` I suggest using `String::new()` in the future.

PR #354, opened 2020-10-20
feat: dynamic branch popup size

### LINKED ISSUE #346: branch popup should have dynamic width
right now the popup uses a fixed width, it should use a dynamic one to show more information on wider terminal sizes:
branches

alocquet:
Fix #346.

I have used a dynamic size (width=60%, height=25%) and I have kept a minimum size for small screen (original size).

Finally, branch name size is now 30% of popup size.

gitui-window-size

This is very interesting to work on a rust project like this :smiley:


extrawurst:
Very cool. Thank you @alocquet

PR #360, opened 2020-10-25
msi try 2

pm100:
try 2

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -70,6 +70,8 @@ jobs:
         files: |
           ./release/*.tar.gz
           ./release/*.zip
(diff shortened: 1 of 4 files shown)


pm100:
OK, i have modded ci.yml to run a modified Makefile, the makefile build the msi and places it in the release dir

I have changed cd.yml to pick up the msi - but I cant test that.

I will revert the change to ci.yml

extrawurst:
@pm100 thanks! we will do a release candidate before releasing `0.11` then the *CD* will run and we can test it. 
can you create expressive commits in future PRs please? :)

pm100:
i need to revert ci.yml first - it still does the wix build as a test

PR #371, opened 2020-10-26
Add windows MSI

pm100:
Final PR with ci.yml reverted

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -70,6 +70,8 @@ jobs:
         files: |
           ./release/*.tar.gz
           ./release/*.zip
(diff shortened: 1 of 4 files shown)


extrawurst:
I think we can drop this. the major part was merged in #360 and I am fixing up the `CI` now but still run `cargo wix` to verify it works: #372

extrawurst:
#372 is merged but just like in your CI test the pipeline says:
```
WARN: Skipped building the release binary
```

can you add the appropriate script that checks the msi file is generated, print its size and fail if it is not generated. (just to be sure it works and out of curiosity how big such a MSI is)

PR #373, opened 2020-10-26
Windows MSI - add display of file size

pm100:
not tested by me

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,8 @@ jobs:
       run: |
         cargo install cargo-wix
         cargo wix init
-        cargo wix --no-build --nocapture
+        cargo wix --no-build --nocapture --output ./target/wix/gitui.msi
+        ls -l ./target/wix/gitui.msi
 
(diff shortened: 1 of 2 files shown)


extrawurst:
Cool but can you add it to the CI step that I added please? 😉

pm100:
you sure you want to run wix everytime? wix is very slow to build ~ 5 mins

extrawurst:
It looks like it takes ~2.5 min in the CI - Ideen like we can put it in CI until we have one successful release with it and then relax the CI

extrawurst:
@pm100 
do you have an idea why *wix* prints: `WARN: Skipped building the release binary` ?

pm100:
Because we said 'nobuikd' since we just cargo built it. Normally cargo wix would build first

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

extrawurst:
I see so it uses the binary in `target` instead of building itself?

extrawurst:
@pm100 thank you :)

PR #375, opened 2020-10-26
Add branch name in Status and ChangesComponent cleanup

### LINKED ISSUE #365: better 'current branch' visualisation
since #362 we now show the outgoing/incoming commits relative to the current branch in a more spacious place in the `left_chunks` of the status tab. 

it would be great to move the actual name of the current branch next to ity:
* this gives the branch name more room
* keeps it in focus
* allows us to cleanup the weird placement of it in the `ChangesComponent`

see `draw_branch_state` in *status.rs*

remique:
This fixes #365.

Still not sure if lookup in `update` function is done correctly:
```rust
self.git_branch_name.lookup().map(Some).unwrap_or(None)
```
Other than that I think everything is okay now :)

### DIFF
--- a/src/components/changes.rs
+++ b/src/components/changes.rs
@@ -52,10 +52,7 @@ impl ChangesComponent {
 
     pub fn update(&mut self) -> Result<()> {
         if self.is_working_dir {
-            self.files.set_title(format!(
-                    "{}",
-                    &self.title, 
-                    ))
+            self.files.set_title(format!("{}", &self.title,))
         }
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


extrawurst (changes requested):

remique:
Got rid of the `update` function as well as `title` string in ChangesComponent as it was never used anymore.

extrawurst:
@remique looks good to me, will merge once the CI is done - thanks for the contribution!! 🥳

extrawurst (approved):

Issue #370, opened 2020-10-25
panic on small terminal when textinput for stash name is open


extrawurst:
```
panic: PanicInfo { payload: Any, message: Some(Trying to access position outside the buffer: x=9, y=2, area=Rect { x: 0, y: 0, width: 9, height: 9 }), location: Location { file: "/Users/stephan/.cargo/registry/src/github.com-1ecc6299db9ec823/tui-0.12.0/src/buffer.rs", line: 217, col: 9 } }
trace:
   0: backtrace::backtrace::libunwind::trace
             at /Users/stephan/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.53/src/backtrace/libunwind.rs:100:5
      backtrace::backtrace::trace_unsynchronized
             at /Users/stephan/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.53/src/backtrace/mod.rs:66:5
   1: backtrace::backtrace::trace
             at /Users/stephan/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.53/src/backtrace/mod.rs:53:14
   2: backtrace::capture::Backtrace::create
             at /Users/stephan/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.53/src/capture.rs:176:9
   3: backtrace::capture::Backtrace::new
             at /Users/stephan/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.53/src/capture.rs:140:22
   4: gitui::set_panic_handlers::{{closure}}
             at src/main.rs:300:25
   5: std::panicking::rust_panic_with_hook
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panicking.rs:573:17
   6: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panicking.rs:476:9
   7: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/sys_common/backtrace.rs:153:18
   8: rust_begin_unwind
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panicking.rs:475:5
   9: std::panicking::begin_panic_fmt
[...]

PR #376, opened 2020-10-26
feat: password obfuscation

### LINKED ISSUE #361: password obfuscation
now that we have https://github.com/extrawurst/gitui/pull/353 done let's clean this up a bit.

most changes in how we use `TextInputComponent`:

1) let's give it a `password` option that will not show the actual password but just `*`
2) let's allow it to be smaller for the username/password popups (1 line and less width)

alocquet:
fix #361 
I don't know if it is ok for you but I managed this one with an input type :
* TEXTAREA = previous behavior
* TEXT : input on one line
* PASSWORD : input on one line and obfuscated

### DIFF
--- a/src/components/cred.rs
+++ b/src/components/cred.rs
@@ -4,7 +4,7 @@ use tui::{backend::Backend, layout::Rect, Frame};
 
 use asyncgit::sync::cred::BasicAuthCredential;
 
-use crate::components::TextInputComponent;
+use crate::components::{InputType, TextInputComponent};
 use crate::{
     components::{
         visibility_blocking, CommandBlocking, CommandInfo, Component,
@@ -37,13 +37,15 @@ impl CredComponent {
                 key_config.clone(),
                 &strings::username_popup_title(&key_config),
                 &strings::username_popup_msg(&key_config),
-            ),
+            )
+            .with_input_type(InputType::Singleline),
             input_password: TextInputComponent::new(
                 theme,
                 key_config.clone(),
                 &strings::password_popup_title(&key_config),
                 &strings::password_popup_msg(&key_config),
-            ),
+            )
+            .with_input_type(InputType::Password),
             key_config,
             cred: BasicAuthCredential::new(None, None),
         }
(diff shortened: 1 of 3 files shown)


extrawurst:
@alocquet sounds cool but lets call it more specifically
```
pub enum InputType {
     SINGLELINE,
     MULTILINE,
     PASSWORD,
 }
```

src/components/textinput.rs
```diff
@@ -166,8 +197,13 @@ impl DrawableComponent for TextInputComponent {
                 self.get_draw_text()
             };
 
-            let area = ui::centered_rect(60, 20, f.size());
-            let area = ui::rect_min(10, 3, area);
+            let area = match self.input_type {
+                InputType::TEXTAREA => {
+                    let area = ui::centered_rect(60, 20, f.size());
+                    ui::rect_min(10, 3, area)
```
extrawurst, line 203: did you test this with terminals smaller 10?

src/components/cred.rs
```diff
@@ -4,6 +4,7 @@ use tui::{backend::Backend, layout::Rect, Frame};
 
 use asyncgit::sync::cred::BasicAuthCredential;
 
+use crate::components::textinput::InputType;
```
extrawurst, line 7: can we merge those uses?

src/components/textinput.rs
```diff
@@ -9,11 +9,20 @@ use crate::{
...
 use tui::{
     backend::Backend, layout::Rect, style::Modifier, text::Span,
     widgets::Clear, Frame,
 };
 
+#[derive(PartialEq)]
+pub enum InputType {
+    TEXT,
+    TEXTAREA,
+    PASSWORD,
```
extrawurst, line 23: lets go for `Singleline`,`Multiline` and `Password`

extrawurst (changes requested):

src/components/textinput.rs
alocquet: Nope because it was existing code for multiline input; but it fails with a panic. I think this is the same issue as #370.
I can fix it in another PR if you want ?

extrawurst: yeah good point

extrawurst:
Nice! Thank you!

PR #383, opened 2020-10-28
Turn popup functions into macros

WizardOhio24:
Changes the functions which make popups work(that is any_popup_visible() and draw_popups()) into macros, so only one macro needs to be called near the top of App.

@extrawurst, is this the type of thing you meant when you said make these dynamic?


extrawurst:
I would like to focus on getting `0.11` out the door - I will revisit this afterwards

WizardOhio24:
Also, with this, would you want to use proc macros?  Perhaps this is a bit ambitious, but I was thinking we could use a syntax for specifying popups like this:
```
App {
  status_tab--
  commitlist-+
  somepopup^^
}
```
Then use a proc macro and just derive app by that macro.  Then there is no need to worry about putting the right accessors and functions in the right place, the derive does it for you.

lastthyme (approved):

PR #386, opened 2020-10-28
fix for linux hooks

pm100:
pre-commit now works

also improved the layout of error message dialog to preserve the message formatting


pm100:
actually this fix works on windows too, but it requires git 4 windows to be installed (or anything else that puts bash on the devs path)

README.md
```diff
@@ -36,7 +36,7 @@
 - Scalable terminal UI layout
 - Async [input polling](assets/perf_compare.jpg)
 - Async git API for fluid control
-
+a
```
extrawurst, line 39: ?

asyncgit/src/sync/hooks.rs
```diff
@@ -94,13 +110,8 @@ fn run_hook(
     hook_script: &str,
     args: &[&str],
 ) -> HookResult {
-    let mut bash_args = vec![hook_script.to_string()];
```
extrawurst, line 97: why not just prepend the "-c" here?

src/components/commit.rs
```diff
@@ -194,6 +194,16 @@ impl CommitComponent {
     }
 
     fn commit_msg(&mut self, msg: String) -> Result<()> {
+        if let HookResult::NotOk(e) = sync::hooks_pre_commit(CWD)? {
+            log::error!("pre-commit hook error: {}", e);
+            self.queue.borrow_mut().push_back(
+                InternalEvent::ShowErrorMsg(format!(
+                    "commit-msg hook error:\n{}",
```
extrawurst, line 201: I think this needs to be `pre-commit` right?

extrawurst (changes requested):

asyncgit/src/sync/hooks.rs
pm100: because its a 2 item vec not one string. I pass a vec of 2 args "-c" and "'script' arg1 arg2 arg3..."

pm100: the prepend -c is what I first did, didnt work the args didnt get passed to the script right

src/components/msg.rs
```diff
@@ -32,17 +32,26 @@ impl DrawableComponent for MsgComponent {
...
-        );
 
-        let area = ui::centered_rect_absolute(65, 25, f.size());
+        // determine the maximum width of text block
+        let lens = self
+            .msg
+            .split('\n')
+            .map(str::len)
+            .collect::<Vec<usize>>();
+        let max = lens.iter().max().expect("max");
```
extrawurst, line 42: We never make sure that msg is not empty which would panic here, right?

asyncgit/src/sync/hooks.rs
```diff
@@ -45,6 +46,21 @@ pub fn hooks_commit_msg(
     }
 }
 
+/// this hook is documented here https://git-scm.com/docs/githooks#_pre_commit
+///
+pub fn hooks_pre_commit(repo_path: &str) -> Result<HookResult> {
```
extrawurst, line 51: Can you add a test for this ?

src/components/msg.rs
```diff
@@ -32,17 +32,26 @@ impl DrawableComponent for MsgComponent {
...
+        let mut width = u16::try_from(*max + 2).expect("cant fail");
```
extrawurst, line 43: Better `expect(“why it can’t fail”)`

src/components/msg.rs
pm100: no its fine, just displays an empty box. (I tested it)

pm100:
FYI - this code works on both linux and windows on my dev system. But I think thats because I have WSL installed. I am not sure what will happen on the github test systems. So I have tests for both windows and non windows. I will check that in and see what happens. I have resolved all yr other issues.

asyncgit/src/sync/hooks.rs
```diff
@@ -204,6 +215,52 @@ exit 0
...
+sys.exit(0)
+        ";
+
+        create_hook(root, HOOK_PRE_COMMIT, hook);
+        let res = hooks_pre_commit(repo_path).unwrap();
+        assert_eq!(res, HookResult::Ok);
+    }
+
+    #[test]
+    fn test_pre_commit_fail_py() {
```
extrawurst, line 242: can you also add a shell based hook test

src/components/msg.rs
```diff
@@ -32,17 +32,30 @@ impl DrawableComponent for MsgComponent {
...
-        );
 
-        let area = ui::centered_rect_absolute(65, 25, f.size());
+        // determine the maximum width of text block
+        let lens = self
+            .msg
+            .split('\n')
+            .map(str::len)
+            .collect::<Vec<usize>>();
+        let mut max = lens.iter().max().expect("max") + 2;
```
extrawurst, line 42: this will still panic if self.msg is empty

extrawurst:
@pm100 two more things and we can merge this

src/components/msg.rs
pm100: no it wont,  I have actually tested it and it works fine.

PR #387, opened 2020-10-29
Fix scrollbar in commit msg

### LINKED ISSUE #350: scrollbar in commit msg sometimes does not show
Screenflick Movie 85

see #308

remique:
Fixes #350

Well I know that may not be the perfect solution, but a better fix itself would have to be inside `scrollbar.rs` anyway. This has to do with the way of how position of scrollbar is calculated, and changing `max` parameter of `draw_scrollbar` function would change nothing as either the scrollbar would never reach the end, or there would be a lot of empty lines.

This functionality would display scrollbar by default (which, regardless of this fix, is also the case with diff) and a simple check can prevent it from rendering. 

What do you think?

### DIFF
--- a/src/components/commit_details/details.rs
+++ b/src/components/commit_details/details.rs
@@ -355,14 +355,18 @@ impl DrawableComponent for DetailsComponent {
         );
 
         if self.focused {
-            ui::draw_scrollbar(
-                f,
-                chunks[1],
-                &self.theme,
-                self.get_number_of_lines(width as usize)
-                    .saturating_sub(height as usize),
-                self.scroll_top.get(),
-            )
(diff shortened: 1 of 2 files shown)


extrawurst:
@remique thanks for looking into this. it bugged me that this still felt like the wrong solution so I took a try on it: https://github.com/extrawurst/gitui/pull/388 - maybe you can play around with that to see how it works for you?

extrawurst:
@remique sorry I did not want to demotivate you by taking a look myself :(

remique:
@extrawurst No worries, you did not 👍  I had some last minute school stuff to resolve, so sorry for no response! 

Also, this is still great learning experience for me=)

Issue #352, opened 2020-10-18
linux fallback for copy-to-clipboard suing xsel


extrawurst:
thanks to @cruessler we now can use copy-to-clipboard on linux. 

but if your linux does not have `xclip` installed we do not try any fallback like `xsel`. we should do that and then have an error msg listing which ones we try.

denismaxim0v:
I'd like to have a try on this if it's still for takes

extrawurst:
@denismaxim0v absolutely - go for it :)

PR #390, opened 2020-10-30
feat: xclip fallback for linux with xsel #352

denismaxim0v:


denismaxim0v:
Fixes #352

extrawurst:
Thanks @denismaxim0v ❤️Will look into this today

extrawurst:
so this looks ok. could be done a little more readable/scalable by having a list of tools and their arguments and iterate it. but I need a second pair of eyes with a linux machine to test this on either way

denismaxim0v:
> so this looks ok. could be done a little more readable/scalable by having a list of tools and their arguments and iterate it. but I need a second pair of eyes with a linux machine to test this on either way

Should I propose another solution in that case? And if so what kind of tools would we include in the list?

Cargo.toml
```diff
@@ -40,6 +40,7 @@ serde = "1.0"
 anyhow = "1.0.33"
 unicode-width = "0.1"
 textwrap = "0.12"
+which = "4.0.2"
```
1wilkens, line 43: This could be os-gated to linux only.
```suggestion

[target.'cfg(target_os = "linux")'.dependencies]
which = "4.0.2"
```

src/clipboard.rs
```diff
@@ -27,10 +28,37 @@ fn execute_copy_command(command: Command, text: &str) -> Result<()> {
     Ok(())
 }
 
+fn gen_command(
+    path: impl AsRef<OsStr>,
+    xclip_syntax: bool,
```
1wilkens, line 33: This works for now, but is limited to the two X commands. If we want to implement Wayland support later, the second parameter could be an enum or we factor out the syntax differently.

1wilkens (commented):
Overall this looks good. The `gen_command` function could be a bit more generic, but for now this should be fine.

1wilkens:
> 
> 
> so this looks ok. could be done a little more readable/scalable by having a list of tools and their arguments and iterate it. but I need a second pair of eyes with a linux machine to test this on either way

Sorry didn't see this comment before reviewing, but that's basically my only nitpick as well. I for one would like Wayland support eventually, but I can also do that in another PR.

denismaxim0v:
> > so this looks ok. could be done a little more readable/scalable by having a list of tools and their arguments and iterate it. but I need a second pair of eyes with a linux machine to test this on either way
> 
> Sorry didn't see this comment before reviewing, but that's basically my only nitpick as well. I for one would like Wayland support eventually, but I can also do that in another PR.

OK cool, I could participate in another PR too. Anything else needed from me here?

src/clipboard.rs
denismaxim0v: Agree, just didn't have ideas about other commands here and am still new to the project, so if you have any ideas, I'd like to hear.

denismaxim0v:
Let me know if I have any work to do here or should we move to another issue please?

src/clipboard.rs
extrawurst: I have no idea about wayland what would be required for it? Or is it just another cli command?

1wilkens: Exactly just another command. There is `wl-copy` but that can probably land in another PR

extrawurst: @1wilkens yea lets do this in another issue - can you create one?

extrawurst:
@denismaxim0v and @1wilkens thanks for your help❤️

src/clipboard.rs
extrawurst: @1wilkens I created one, please check if I phrased this accurately for I am no wayland user: https://github.com/extrawurst/gitui/issues/397

Issue #382, opened 2020-10-28
Better explanation how to configure external editor


ser-drephs:
I wish there was some kind of multi-line commit message support. 

Inside the commit dialog it should be possible to create a line break using <kbd>CTRL+ENTER</kbd>.

It is possible to use any other external editor, but configuring it is also not documented in any way 😊

extrawurst:
@Ser-Drephs thats what the external editor support is for. What platform are you on? `vi` is usually available and used? Also if you try to open in external editor what error do you get? Cause it should tell you how it works

Ser-Drephs:
Thanks for the fast reply @extrawurst 

I am using Windows. With Windows there is no `vi`. I installed `vim` using chocolatey. Also using the external editor I had to dig through the code to find out that I need to set the environment variable `GIT_EDITOR`.

extrawurst:
Again, what error message did you get before?

Ser-Drephs:
Sorry forgot that 😁

```
Error
failed to launch editor:
"vi": Das System kann die angegebene Datei nicht finden. (os error 2)
```

Ser-Drephs:
After installing `vim` ([Link](https://chocolatey.org/packages/vim)) and adding User Environment Variable `GIT_EDITOR` with path `C:\Tools\vim\vim82\vim.exe` the external editor launches.

extrawurst:
Ok then the issue is that this error should be more helpful. It should tell you what to do.

akshay1992kalbhor:
I would like to help with this one!
Aside from the error message is there anything else that needs to be changed?

extrawurst:
@akshay1992kalbhor no I think the error msg is the important part here

akshay1992kalbhor:
Ok I have a few questions which might or might not be stupid.
1. I am assuming the error message should mention just the 'GIT_EDITOR' environment variable and not the other ones. Or should it?
2. How much info should the message have? Should it include the commands to add environment variables on all the platforms? ( I am be overthinking this one)

Ser-Drephs:
From a user perspective:

1. It should have both. The error message and informing the user about the environment variable `GIT_EDITOR`.
1. Using a tool like this I would assume that the user knows how to add an environment variable 😉

akshay1992kalbhor:
[...]

PR #395, opened 2020-10-31
fix cursor lost at end of line

pm100:


extrawurst:
image

> This looks slightly odd

I guess these double bars is what you meant?

pm100:
on my system it doesnt look that bad (different fonts I guess) I will change it so that the trailing space is not underlined. Give me a sec

extrawurst:
now testing this I realised we have a bigger issue:

image

new lines are not applied anymore. I suspect this is related to out tui-rs upgrade https://github.com/extrawurst/gitui/pull/289 by @WizardOhio24 - can you check if we can fix this in this PR aswell?

WizardOhio24:
You are most likely correct.  With the TUI update, each new line needs to be in a separate Spans, text in the same Spans is rendered on the same line, and so this code renders all characters on the same line without regard for type of character.  https://github.com/extrawurst/gitui/blob/c4fdbf7aba9b6925b941816e84f4459746cbeb89/src/components/textinput.rs#L134-L138
To fix this you just need to split the lines by '\n', inserting a new spans where each is found(just split, map).  I did this somewhere when I was updating TUI I think.  I believe there was a TUI issue to have text be able to display '\n'(i.e insert a newline) if it wasn't in spans but I'm not sure how it was resolved.

pm100:
> You are most likely correct. With the TUI update, each new line needs to be in a separate Spans, text in the same Spans is rendered on the same line, and so this code renders all characters on the same line without regard for type of character.
> 
> https://github.com/extrawurst/gitui/blob/c4fdbf7aba9b6925b941816e84f4459746cbeb89/src/components/textinput.rs#L134-L138
> 
> 
> To fix this you just need to split the lines by '\n', inserting a new spans where each is found(just split, map). I did this somewhere when I was updating TUI I think. I believe there was a TUI issue to have text be able to display '\n'(i.e insert a newline) if it wasn't in spans but I'm not sure how it was resolved.

The failure happens when there is only one line, it is for sure a tui bug. Look at the screen shots I posted in the tui bug report

extrawurst:
> The failure happens when there is only one line, it is for sure a tui
bug. Look at the screen shots I posted in the tui bug report

ehm no, the screenshot i posted is actually two lines

pm100:
interestingly enough using the external editor on windows (I tried 2 different ones) ends up ignoring the new lines completely.

image

extrawurst:
That’s exactly the problem for which wizard described the cause

pm100:
you want me to roll that fix into this PR?

extrawurst:
we can actually tackle this in another PR first if you like to split it up, but this one has to wait then until we have the general line breaking issue is fixed

pm100:
I am happy to fix it, just wanted to be clear

extrawurst:
@pm100 ok we have a misunderstanding here :( 
What this PR was about is fixing two issues with the displaying of text. For adding this new feature of multiline text editing we should have indeed opened a new PR. Can you Split this in two?

pm100:
So I should make the commit msg read only when you return from ^e? 

BTW: The ^Enter idea came from here , I found it when I was trying to work out how to make ^e work.

https://github.com/extrawurst/gitui/issues/382

extrawurst:
@pm100 can you please check the behaviour of gitui `0.10.1` before the `tui-rs` upgrade in 9316863 because it supported this without being readonly

WizardOhio24:
Don't know if this helps now, but I found the text PR in TUI: https://github.com/fdehau/tui-rs/pull/361, which allows for text to display "\n"(i.e insert a new line).  It might be easier to use what was introduced there, as in allow newline characters, then there doesn't really need to be much change to the way it's currently implemented and should remove the read-only issue.

pm100:
it supported what? The current commit box doesnt support multiple lines, It has no code for up or down arrow, it has no concept of current line. Tell me what you want the behavior to be and I will make it so

pm100:
to be clear - this is not a display issue, thats trivial 2 min fix. The issue is - should the user be allowed to edit a multi line commit message? If 'no' then I will make the commit box read only. If 'yes' then we need this PR.  

Also what should happen if the multi line text has more lines than can be displayed, which lines should be displayed?

I will point out that users will be surprised that they can no longer edit when the return from ^e

WizardOhio24:
So, I think what @extrawurst is saying is that this PR was only to fix the missing cursor at the end of the line.  A multiline editor(A great new feature so the user doesn't have to use an external editor), is a different PR.  So, I think the easiest thing to do is to branch from "Implement full multi-line support in textinput" commit(225b972) and put it in a different PR.  (Correct me if I'm wrong)

I checked out 10.1 and in it you could modify multiple lines but you could only do it by moving right, so you had to keep going right until it "wrapped round" onto a new line.  So I think the way you've implemented it is probably better/easier for the user, i.e using the up/down arrow keys.
I think the problem was that the text input was previously not read only, the user could edit multiple lines by going to the end of the line above and moving right.  The user should be able to edit text as much in GitUI as in an external editor(at a basic level) and it shouldn't be read only.
Hope that helps.

pm100:
i will close this PR.

I will resubmit the original PR for this bug

I will open a separate issue for the multi line issues

PR #400, opened 2020-11-03
better explanation how to configure external editor (#382)

akshay1992kalbhor:
Any suggestions will be appreciated. This is kind of a first draft.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -259,9 +259,10 @@ impl App {
                     None => self.commit.show_editor(),
                 };
 
-                if let Err(e) = result {
+                if let Err(_e) = result {
                     let msg =
-                        format!("failed to launch editor:\n{}", e);
+                        format!("Failed to open the external editor. Make sure the\n\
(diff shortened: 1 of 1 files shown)


WizardOhio24:
Thanks for the PR.  
With this, should it not show the user the actual error because it might not be that GIT_EDITOR is not set, it could be something entirely different, I mean, in the original issue it said vi could not be found, but that could be for a number of reasons, not just GIT_EDITOR.  
I'd say as well, it should include "and the editor is installed and valid.", if the user doesn't have micro installed, they can't use micro, no matter if GIT_EDITOR is set.  

So, the error message then tells the user the most common way to fix it, but also shows the user the error.

What do you think?

extrawurst:
@akshay1992kalbhor thanks for your interest in supporting gitui. 
can you check the new behaviour for when some editor env variable is empty: 156776a

for the time being I mark this as draft

extrawurst:
seems unmaintained

PR #403, opened 2020-11-05
Add ability to select multiple commits in commit log

WizardOhio24:
Adds the ability to select multiple commits.  After this, squashing commits should be able to be implemented, but that should be a different PR.
To select multiple commits, use SHIFT+ARROW, so shift-down would select the current commit and the commit below.

Screenshot_20201105_235336

### DIFF
--- a/src/components/commitlist.rs
+++ b/src/components/commitlist.rs
@@ -269,7 +269,7 @@ impl CommitList {
         let selection = self.relative_selection();
 
         let mut txt: Vec<Spans> = Vec::with_capacity(height);
-        //self.scroll_top.get() ==
+
         for (idx, e) in self
             .items
             .iter()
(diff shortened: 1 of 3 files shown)


extrawurst:

extrawurst:
@WizardOhio24 before actually using a selection of multiple commits I do not see a benefit in merging this.

WizardOhio24:
With this, it allows for squashing commits, I can try implementing squash commits instead (built on this).  Shouldn't be too difficult, just rebase and skip selected commits.  Unless something else is higher priority, happy to do something else as well.

WizardOhio24:
Waiting for #312.  Much of the functionality of squashing commits and rewording commits is the same (both rebasing in different ways, squashing ignores x commits and then commits under a different name, rewording makes commit y have a different name).

Issue #247, opened 2020-08-27
whitespace at end of line in commit msg eats cursor


extrawurst:
<img width="291" alt="image" src="https://user-images.githubusercontent.com/776816/91494938-fc6fca80-e8b9-11ea-9bdf-b417b353f1a0.png">

move cursor one left:
<img width="299" alt="image" src="https://user-images.githubusercontent.com/776816/91494969-098cb980-e8ba-11ea-8bf7-5bc70a3cd98f.png">

ciaranmooney:
I think I understand where this might be coming from. I believe the building of the output of a TextInput component mixes the styling and content to a point where the above issue can occur.

As part of #197 I am partway through refactoring how xxx builds up the Text return value. Where the String containing the characters and the styling are built up in stages. See this [branch](https://github.com/ciaranmooney/gitui/commits/feature/refactoring_textinput).

My changes are currently broken with regards to an existing test and UTF-8 (the carriage return symbol). Once that is sorted I'll write another test for the above scenario to see if the issue has been solved as part of my changes.

edfloreshz:
I'm not very familiar with the project but I'd like to try to fix the issue.

extrawurst:
@edfloreshz I think @ciaranmooney is working on this already, please make sure to checkout his branch first, so we don't have two people fixing the same issue :)

ciaranmooney:
@edfloreshz Indeed, I believe my refactor of the TextComponent will remove this issue (see [https://github.com/ciaranmooney/gitui/tree/feature/refactoring_textinput](my branch) Although I haven't tested it yet. I've got my head around how the area is working so I feel well placed to solve this issue.

edfloreshz:
Okay, thanks for letting me know @ciaranmooney @extrawurst! 🙂

extrawurst:
@ciaranmooney is this still coming in at some point? maybe other might want to look into this particular issue until then so we can clear it for `0.11`

ciaranmooney:
I am afraid I will have to concede defeat to this issue and allow others to take a crack at it.

pm100:
can you explain what the bug is, I dont see it

extrawurst:
On the screenshots you can see that the underscore indicating the cursor position disappears when under a space at the end of a line

pm100:
how did you get that new line in there?

And I have no cursor when typing only when backtracking (left / right arrow), is that right?

extrawurst:
[...]

PR #404, opened 2020-11-06
Fix for cursor not displayed in textinput (247)

pm100:
https://github.com/extrawurst/gitui/issues/247

the space at the end of line should be NBSP telling tui not to eat it in text wrap mode but the code in tui is broken too.

Fix posted to tiu

https://github.com/fdehau/tui-rs/pull/405

In the mean time this fix uses '_' to get the underlined space at the end of line

Comments are left in place explaining what to do when tui fixes this issue.

### DIFF
--- a/src/components/textinput.rs
+++ b/src/components/textinput.rs
@@ -17,8 +17,8 @@ use tui::{
     widgets::Clear, Frame,
 };
 
-//const EOL_CHAR: &str = "\u{200b}";
 const EOL_CHAR: &str = "_";
+
 #[derive(PartialEq)]
 pub enum InputType {
     Singleline,
--- a/src/components/textinput.rs
+++ b/src/components/textinput.rs
@@ -17,7 +17,8 @@ use tui::{
     widgets::Clear, Frame,
 };
 
-const EOL_CHAR: &str = "\u{200b}";
+//const EOL_CHAR: &str = "\u{200b}";
+const EOL_CHAR: &str = "_";
 #[derive(PartialEq)]
 pub enum InputType {
     Singleline,
@@ -159,15 +160,10 @@ impl TextInputComponent {
             ));
         }
 
-        // ... and this conditional underline needs to be removed
-        if cursor_str == EOL_CHAR {
-            txt.push(Span::styled(cursor_str, style));
-        } else {
-            txt.push(Span::styled(
-                cursor_str,
-                style.add_modifier(Modifier::UNDERLINED),
-            ));
-        }
+        txt.push(Span::styled(
+            cursor_str,
+            style.add_modifier(Modifier::UNDERLINED),
+        ));
 
         // The final portion of the text is added if there are
         // still remaining characters.
@@ -373,8 +369,7 @@ mod tests {
             "",
         );
         let theme = SharedTheme::default();
-        // retained for when tui trailing NBSP bug fixed
-        let _underlined = theme
+        let underlined = theme
             .text(true, false)
             .add_modifier(Modifier::UNDERLINED);
 
@@ -389,7 +384,7 @@ mod tests {
         assert_eq!(get_text(&txt[0]), Some("a"));
         assert_eq!(get_style(&txt[0]), Some(&not_underlined));
         assert_eq!(get_text(&txt[1]), Some(EOL_CHAR));
-        assert_eq!(get_style(&txt[1]), Some(&not_underlined));
+        assert_eq!(get_style(&txt[1]), Some(&underlined));
     }
 
     #[test]
--- a/src/components/textinput.rs
+++ b/src/components/textinput.rs
@@ -1,3 +1,4 @@
+use crate::ui::Size;
 use crate::{
     components::{
         popup_paragraph, visibility_blocking, CommandBlocking,
@@ -200,7 +201,11 @@ impl DrawableComponent for TextInputComponent {
             let area = match self.input_type {
                 InputType::Multiline => {
                     let area = ui::centered_rect(60, 20, f.size());
-                    ui::rect_min(10, 3, area)
+                    ui::rect_inside(
+                        Size::new(10, 3),
+                        f.size().into(),
+                        area,
+                    )
                 }
                 _ => ui::centered_rect_absolute(32, 3, f.size()),
             };
--- a/src/components/textinput.rs
+++ b/src/components/textinput.rs
@@ -17,6 +17,7 @@ use tui::{
     widgets::Clear, Frame,
 };
 
+const EOL_CHAR: &str = "\u{200b}";
 #[derive(PartialEq)]
 pub enum InputType {
     Singleline,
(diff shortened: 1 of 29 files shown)


pm100:
@extrawurst the failures here are due to clippy complaints about asyncgit/revlog.rs. Not part of this PR. For some reason when I run clippy it doesnt complain, I have the same clippy version and I ran it like the makefile runs it.

src/components/textinput.rs
```diff
@@ -138,11 +138,14 @@ impl TextInputComponent {
             ));
         }
 
+        // this code with _ for the trailing cursor character needs to be revised once tui fixes
```
WizardOhio24, line 141: Why are you retaining all this space stuff?  This fix seems just as good.  Is there any reason using an underlined space is better than an underline character?

src/components/textinput.rs
```diff
@@ -138,11 +138,14 @@ impl TextInputComponent {
...
+        // it should be NBSP => const NBSP: &str = "\u{00a0}";
+        // ...
         let cursor_str = self
             .next_char_position()
             // if the cursor is at the end of the msg
             // a whitespace is used to underline
```
WizardOhio24, line 147: Change "whitespace" to "underline character"?

src/components/textinput.rs
```diff
@@ -155,10 +158,15 @@ impl TextInputComponent {
             ));
         }
 
-        txt.push(Span::styled(
-            cursor_str,
-            style.add_modifier(Modifier::UNDERLINED),
-        ));
+        // ... and this conditional underline needs to be removed
```
WizardOhio24, line 161: Again, see previous comment about if retaining all the space stuff is necessary.

src/components/textinput.rs
```diff
@@ -364,7 +372,8 @@ mod tests {
             "",
         );
         let theme = SharedTheme::default();
-        let underlined = theme
+        // retained for when tui trailing NBSP bug fixed
```
WizardOhio24, line 375: These themes/styles can be removed because the cursor is no longer determined through style, but by using the actual "_" character.  That's all that needs to be tested for.

src/components/textinput.rs
```diff
@@ -378,8 +387,8 @@ mod tests {
         assert_eq!(txt.len(), 2);
         assert_eq!(get_text(&txt[0]), Some("a"));
         assert_eq!(get_style(&txt[0]), Some(&not_underlined));
-        assert_eq!(get_text(&txt[1]), Some(" "));
-        assert_eq!(get_style(&txt[1]), Some(&underlined));
+        assert_eq!(get_text(&txt[1]), Some("_"));
+        assert_eq!(get_style(&txt[1]), Some(&not_underlined));
```
WizardOhio24, line 391: See previous comment, the style checks can be removed.

src/components/textinput.rs
pm100: because a space with an underline places the pixels in the correct place visually.

src/components/textinput.rs
pm100: no the cursor is still an underline for all chars except the CR. So the test should be retained to ensure that the undelining logic is correct

src/components/textinput.rs
WizardOhio24: I didn't notice that the first time, it's a small difference, I've always thought the underline of an underlined piece of text is in-line with the underline character.  Perhaps add that to the comment?  Also, could you mention the TUI issue that needs to be resolved in the comment as well.

src/components/textinput.rs
WizardOhio24: This comment is still incorrect, it is not a whitespace being used(temporarily).  It can be changed back when the TUI issue is fixed.

src/components/textinput.rs
WizardOhio24: Resolved

src/components/textinput.rs
WizardOhio24: You are correct, I misunderstood the purpose of the test.

src/components/textinput.rs
WizardOhio24: Resolved

WizardOhio24:
@pm100 I did a bit of research into this and found unicode 200b, take a look at my PR(https://github.com/extrawurst/gitui/pull/413), (it's a 5 character change so I just put in a PR for it).  What do you think?

WizardOhio24:
The appearance of that one still seems off, like a bold underline, if you wanted to use ZWSP, you have to use " \u{200b}" (that is      "{space}\u{200b}"), the Zero Width space has zero width :) . Is there a reason we shouldn't use ZWSP?  It just tells TUI not to remove the underlined space before it.

pm100:
well it works with just the ZWSP. On a fixed font ZWSP takes a whole character. Doing what you suggest will take 2 chars in tius mind, not sure what will happen when tiu wraps that at then end of the line. 

I am trying to work out what we are trying to do here, this bug is fixed with a workaround pending a fix from tui.

pm100:
sigh, now I end up with 2 underlined characters. the fact that cursor_str is not 2 chars messes underlining up.

image

WizardOhio24:
Lets just use an underline, this is too much trouble for something which has a fix in for TUI and which is a very small problem anyway (probably unnoticeable by most).  I thought there would be a better character that could be used to get round the "underlined underline" problem, but now I see any other characters seem to cause problems.
Remove the comment above EOL_CHAR and lets just go with that.

Thanks for seeing this PR through to the end.  @extrawurst? Merge?

pm100:
hang on I need to revert

pm100:
ok, back to where we started

pm100:
ok @extrawurst  you can merge this

WizardOhio24:
One (small) last thing with this, could you put in a link to the TUI issue which needs to be resolved.  Just so we know in the future.

pm100:
its in the issue this pr is linked to

pm100:
https://github.com/fdehau/tui-rs/issues/404

WizardOhio24:
Yes I know, I was just thinking you could put that link in the code comment, where you say 'this code needs to be revisited once tui fixes', so if someone else comes across it they know why the code is like this and the issue associated with it.  

Also, what is the comment saying 'it should be NBSP => const ...'?  Is it relevant here?

pm100:
since the fix is to use NBSP (not space) I have added a comment to should show what the code for it is. The comment now say EOL_CHAR =

WizardOhio24:
Ah ok, I think I understand what the comment says, I was confused because '=>' is a rust symbol, so I didn't understand what you meant.  You're saying EOL_CHAR should be NBSP, once the TUI issue is resolved?

extrawurst:
@pm100 is this still necessary once the bug in tui-rs is fixed? (see https://github.com/fdehau/tui-rs/issues/433)

pm100:
@extrawurst no, a different change is needed once that fix hits . I have comments in the fix for what needs doing

pm100:
@extrawurst so is that fix merged into gitui now? If so I will pull the PR and post a new one

extrawurst:
@pm100 tui-rs still did not release this but I fixed it in 3ff77e0 - I think we can close this PR

Issue #405, opened 2020-11-06
multiline display in textinput is broken


pm100:
Following recent tui upgrade newlines in a text block are lost so a multi line commit comment (as an example) is displayed all on one line

ctrl+e into external editor

![image](https://user-images.githubusercontent.com/489231/98420728-a26b4e00-203c-11eb-9fa5-3dca51111b5d.png)
::x

return back to gitui

![image](https://user-images.githubusercontent.com/489231/98420797-cd55a200-203c-11eb-98d1-4acaa57d4e7b.png)

pm100:
I will fix this so that it displays 2 lines.

PR #407, opened 2020-11-06
Multi line

pm100:
This is the bare minimum (as requested) of changes to fix this bug https://github.com/extrawurst/gitui/issues/405

 - You can still edit the first line when you return from the external editor.
 -  If you backspace out the whole first line then the second line moves up tp become the first line. I can prevent that if you like.
 -  There is no way to see more lines than will fit in the dialog box

Its my understanding that this is what others wanted.

Note - the display of the hard unicode CR character is gone because I wasnt sure what to do with it. The current code only shows it on the current line (which can only be the first one), should it be shown on every line? In what place? Should it be added if it would make the line overflow the screen width

### DIFF
--- a/src/components/cred.rs
+++ b/src/components/cred.rs
@@ -4,7 +4,7 @@ use tui::{backend::Backend, layout::Rect, Frame};
 
 use asyncgit::sync::cred::BasicAuthCredential;
 
-use crate::components::TextInputComponent;
+use crate::components::{InputType, TextInputComponent};
 use crate::{
     components::{
         visibility_blocking, CommandBlocking, CommandInfo, Component,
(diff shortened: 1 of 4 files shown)


pm100:
correct branch this time :-)

pm100:
these failures are also clippy complaints about revlog.rs

pm100:
reminder - I have another PR that converts the textinput dialog to full multi line capabilities (when in multi-line mode)

WizardOhio24:
With this, it still has the previous fixes for the underline not being displayed, now covered in #404, you might want to revert that commit(680d3d2) so the rebase/merge when merging isn't difficult.

pm100:
closing till other PR resolved

PR #408, opened 2020-11-07
Fix clippy build error

WizardOhio24:
Nightly build is failing because of a clippy error, this fixes it.


extrawurst:
@WizardOhio24 thanks for looking into this! this should fix https://github.com/extrawurst/gitui/pull/404#issuecomment-723328083 aswell

PR #410, opened 2020-11-07
Add cache to GithubActions

### LINKED ISSUE #409: Use cache in Github Action for build process(CI/CD)
Github actions build(CI/CD).  When building, the build can take a long time because it has to install Rust itself, all the dependencies and build all of them, from scratch, even if nothing has changed, it would be much faster if these were cached so it only downloads crates which have been updated/changed.

See https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows.
This explains how caching works for actions.

WizardOhio24:
Closes #409 

Adds the ability for Github Actions to cache the build process so it doesn't download or build anything if nothing is changed.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -16,6 +16,14 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
+    - name: Restore cargo cache
+      uses: actions/cache@v2
+      env:
+        cache-name: cache-cargo-crates
+      with:
+        path: ~/.cargo
+        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('Cargo.lock') }}
+
     - name: Get version
       id: get_version
       run: echo ::set-output name=version::${GITHUB_REF/refs\/tags\//}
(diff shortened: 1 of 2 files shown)


extrawurst:
@WizardOhio24 any reason to only do this for the release pipeline `cd.yml` instead of `ci.yml`? I actually think we should do it the other way around and make sure to build clean without caching in `cd` and use caching in `ci`

WizardOhio24:
You are correct, I modified the wrong file, don't know what I was thinking.

extrawurst:
@WizardOhio24 I looked into the docs and I am wondering shouldn't we also add the rust `stable`/`nightly` part into the cache key? I assume fetching a cache of a previous stable build in a nightly one will not save much time, right?

extrawurst:
@WizardOhio24 in a0ba3d4 i removed an unneeded `cargo clean` which I hope lead to this caching not to have zero benefit. I would like to see this addition to have some speed increase first. 

can you update it with the mentioned change so that we can see if the builds become faster?

extrawurst:
you know what, lets merge it and see how much it improves! thanks

PR #412, opened 2020-11-07
Cache ./cargo/bin

WizardOhio24:
Is there any reason we aren't caching this?  This is where rustup, rustc, cargo... is stored which are all downloaded at the start of the CI.


extrawurst:
lets see how big the cache becomes with that :)

PR #413, opened 2020-11-08
use '<space>\u{200b}' for a '<space>' so tui doesn't remove it

WizardOhio24:
This causes the ending space to be preserved in the text block so it shows as underlined, this can be removed once https://github.com/fdehau/tui-rs/issues/404 is resolved.

### DIFF
--- a/src/components/textinput.rs
+++ b/src/components/textinput.rs
@@ -142,7 +142,7 @@ impl TextInputComponent {
             .next_char_position()
             // if the cursor is at the end of the msg
             // a whitespace is used to underline
-            .map_or(" ".to_owned(), |pos| {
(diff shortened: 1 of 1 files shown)


pm100:
ok so now we have 2 fixes for the same issue.

WizardOhio24:
I just put this here to show you the code, we can merge this or yours, this just has the smallest change and puts the underline in the correct location(visibly).  This way, only the `\u{200b}` needs to be removed when the TUI issue is fixed.

pm100:
well I am not thrilled with this, that character is 'zero width space' so its not clear that its underline occupying a full character width is actually a bug or not. But fine I will change it

pm100:
ok i updated original PR, you want to close this one so extrawaurst knows what to merge

Issue #414, opened 2020-11-09
core.editor is ignored


pm100:
git goes through the following sequence to find the external editor
 - GIT_EDITOR
 - core.editor
 - VISUAL
 - EDITOR
 - "vi"

Currently gitui does not look at core.editor. vscode sets this value when you install the git extension for it so this is a very common situation. I expect other cases set it too.

During investigating this I found 2 other issues
 - gitui calls config.get_str - this always fails, see https://github.com/rust-lang/git2-rs/issues/474
 - gitui when passed a complex editor string doesn't work

vscode set my core.editor to

`core.editor="C:\Users\paulm\AppData\Local\Programs\Microsoft VS Code\Code.exe" --wait`

This doesnt work once I fixed the config lookup.

![image](https://user-images.githubusercontent.com/489231/98504408-c5982800-220b-11eb-8bc2-634f7cf3a34c.png)

The code hints that this would not work.

I think the other calls to config.get_str need to be changed too. I certainly confirmed that get_str on code.editor was silently failing with the error indicated in the git2 bug. I expect that others will silently fail too.

extrawurst:
@pm100 how much of this is fixed via #419 ?

pm100:
all of it. Except the referenced call to get user.name and user.email (fix for bug to do withuser.name not set )  i didnt want to roll to much into on PR. There is another PR that really isnt needed either, was created by back and forht over this one but it is resvolved here https://github.com/extrawurst/gitui/pull/426

extrawurst:
Cool thanks for that, will work my way through the PRs over the next days. I finally want to release push - call defeat on pull for now. it is just too complex right now and the unreleased version grows out of proportion

PR #416, opened 2020-11-10
code.editor config ignored

pm100:
https://github.com/extrawurst/gitui/issues/414

This PR fixes all 3 issues in the original bug

 - gitui does not look at code.editor setting
 - git2 config.get_str always fails
 - gitui doesnt like editors with spaces in their names (vscode sets itself in core.editor)

does not fix the potential error of editors setting with spaces in the args (as noted in the original comments)

does not change the other places where get_str is called - it probably should

I wanted to add a test for trying to read config str with non UTF8 string name (since this is a failure mode for git2) but this project bans unsafe. unsafe is needed to be able to create an illegal utf8 string

Also note that the code working with config.get_entry does not work how the git2 docs say it should (what errors occur when etc) but it actually works.

Finally:  Reviewers, I am still learning my way around rusts idioms and would welcome feedback on how this code can be improved. In particular all the or_else, and_then, if_you_like ,... chaining

### DIFF
--- a/asyncgit/src/sync/utils.rs
+++ b/asyncgit/src/sync/utils.rs
@@ -144,6 +144,29 @@ pub fn stage_addremoved(repo_path: &str, path: &Path) -> Result<()> {
     Ok(())
 }
 
+/// get string from config
+pub fn get_config_string(
+    repo_path: &str,
+    key: &str,
+) -> Result<Option<String>> {
+    let repo = repo(repo_path)?;
+    let cfg = repo.config()?;
+
+    // this code doesnt match what the doc says regarding what
+    // gets returned when but it actually works
+    let entry_res = cfg.get_entry(key);
+
+    let entry = match entry_res {
+        Ok(ent) => ent,
+        Err(_) => return Ok(None),
+    };
+
+    if !entry.has_value() {
+        Ok(None)
+    } else {
+        Ok(entry.value().map(|s| s.to_string()))
+    }
+}
 /// helper function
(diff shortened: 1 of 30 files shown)


WizardOhio24:
@pm100 You have included all the commits for fixing the missing cursor at the end of the line :anguished:, you might want to cherry pick this onto a new branch then force push that here.

asyncgit/src/sync/utils.rs
```diff
@@ -144,6 +144,29 @@ pub fn stage_addremoved(repo_path: &str, path: &Path) -> Result<()> {
...
+    let entry = match entry_res {
+        Ok(ent) => ent,
+        Err(_) => return Ok(None),
+    };
+
+    if !entry.has_value() {
+        Ok(None)
+    } else {
+        Ok(entry.value().map(|s| s.to_string()))
+    }
```
WizardOhio24, line 168: With the above code, a more rust way to write this might be:
```
    if let Some(value) = repo.config()?.get_entry(key)?.value() {
        return Ok(Some(String::from(value)));
    }
    Ok(None);
```
If anyone else has any suggestions feel free.

pm100:
@WizardOhio24 aha, total git noob here (so why am i working on gitui, cos I like tui projects). I think i know my mistake. In my mind 'git branch foo' makes a new clean branch . I subconsciously though that it would be from master.

 I assume I have to go
 - get checkout master
 - git branch foo
 - git checkout foo

Is there a magic command (rebase?) that will sort this out for me. 

I will sort this out before looking at your other suggestions

PR #417, opened 2020-11-10
Add scrollbar in branchlist

### LINKED ISSUE #392: scrollbar in branchlist
now that #368 landed it would be nice to have our Scrollbar in the branchlist aswell once It has enough entries

remique:
Fixes #392 

No one seemed to be working on it. Apparently everything was already set up, just had to draw the actual scrollbar.

### DIFF
--- a/src/components/select_branch.rs
+++ b/src/components/select_branch.rs
@@ -78,11 +78,19 @@ impl DrawableComponent for SelectBranchComponent {
                 .block(
                     Block::default()
                         .title(strings::SELECT_BRANCH_POPUP_MSG)
-                        .borders(Borders::ALL)
-                        .border_type(BorderType::Thick),
+                        .border_type(BorderType::Thick)
+                        .borders(Borders::ALL),
                 )
                 .alignment(Alignment::Left),
                 area,
+            );
+
+            ui::draw_scrollbar(
+                f,
+                area,
+                &self.theme,
+                self.branch_names.len(),
+                self.scroll_top.get(),
             );
         }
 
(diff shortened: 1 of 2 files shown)


WizardOhio24:
Thanks for the PR.

In the commit message scrollbar (the other location where this is used), this highlights the right bar using `THICK_VERTICAL` (that is `BorderType::Thick`), to show the min and max of the bar (when the commit message is too long).  Think this will be possible here? Perhaps set the right hand side to `BorderType::Double` if there is a scroll bar?  Could pass in the symbol to draw to the `draw_scrollbar` function and use `DOUBLE_VERTICAL` for this.

Otherwise LGTM!

remique:
Just had to remove `Block` border setting. It works just the same as commit message scrollbar and every other scrollbar-component.

WizardOhio24:
Hmm, the problem is that all popups have a thick border, and this one shouldn't really be different.  I tried it with a double line, here's how it looks: 
ScreenshotBranchList
Not bad.   We could just use the double thick line for scroll-bars(everywhere), so as to be consistent and work with both popups and components in tabs.

remique:
Ah indeed I did not notice that popups have thicker border. 

Just to clarify your last comment, all scrollbars in the app should have `DOUBLE_VERTICAL` bar no matter if it is a popup or a tab? Am I understanding correctly?

WizardOhio24:
Correct.

What I'm thinking is that all the scroll bars should be the same for consistency and they should work with both popups and components in tabs(so both single and thick). The only other option TUI offers is Double which looks alright on both.  So I'd say then that all the scrollbars should use Double.  (Just a small change in the scrollbar component)

WizardOhio24:
LGTM!  @extrawurst merge?

extrawurst:
@remique thanks for looking into this and @WizardOhio24 for the review - that's a huge help!

PR #421, opened 2020-11-13
Support force push

### LINKED ISSUE #274: support force push
see https://github.com/libgit2/libgit2/issues/4286 for how to do `force` with libgit2

we should try to push first and if it fails we ask if we should force push, this way we only need one `push` command

WizardOhio24:
Closes #274 

This currently force pushes (with SHIFT+P) but does not ask the user to confirm their action, I'll implement that as well.


WizardOhio24:
I added the confirm popup but the confirm box seems a bit small, I'll increase it's size.

WizardOhio24:
I think this can be merged(pending review), @extrawurst? The only thing is that it does not use the correct upstream branch at the moment, it uses the fact that the upstream must be named the same as the local in currently (in GitUI), so this will have to be changed when a upstream with a different name is supported.

extrawurst:
@WizardOhio24 why do you close these?

extrawurst:
did not have time to review them yet. still working myself through the PRs since yesterday

assets/vim_style_key_config.ron
```diff
@@ -69,5 +69,6 @@
     select_branch: ( code: Char('b'), modifiers: ( bits: 0,),),
     delete_branch: ( code: Char('D'), modifiers: ( bits: 1,),),
     push: ( code: Char('p'), modifiers: ( bits: 0,),),
+    force_push: ( code: Char('p'), modifiers: ( bits: 2,),),
```
extrawurst, line 72: doesn't this need to be `'P'` and `bits: 1` like with `delete_branch`?

extrawurst (changes requested):
@WizardOhio24 I think for `force` we should implement the finding of the remote name instead of simply assuming it is the same as the local branch. in case of force this could otherwise lead to bad surprises, don't you think?

WizardOhio24:
This now force pushes to the same place as a normal push would go to, which I think is just the set upstream repo.

assets/vim_style_key_config.ron
WizardOhio24: It works both ways, but I'll change it to 'P' and bits: 1.

asyncgit/src/sync/remotes.rs
```diff
@@ -111,6 +111,7 @@ pub fn push(
     repo_path: &str,
     remote: &str,
     branch: &str,
+    force: bool,
```
extrawurst, line 114: this should be tested:
a) that a push that would require `force` fails as expected
b) that force push rewrites history in that case correctly

WizardOhio24: Ok, will need to make tests for `branch_set_upstream` first, can't get it working, saying `panicked at 'Failed to setup': Git(Error { code: -3, klass: 4, message: "cannot set upstream for branch \'master\'" })', asyncgit/src/sync/remotes.rs:330:34`.

WizardOhio24: Force push test working, didn't need `branch_set_upstream`.

WizardOhio24: Don't merge yet, I'll do a test for re-writes history as well.

WizardOhio24: Done

extrawurst:
@WizardOhio24 thanks man!❤️

PR #422, opened 2020-11-13
Use left and right arrow keys to navigate log and commit details

### LINKED ISSUE #418: Use left arrow to exit commit details and return to commit log
I'm always frustrated when I press the right key to see the details of a commit in the commit log then press left and I can't get out of it, instead I have press esc.  Surely when you access the details using the right key you should be able to leave them with the left.

When you press the left key (on the left side) when viewing commit details, you should be able to leave the commit details and return to the commit log.

WizardOhio24:
Fixes #418 

Adds the ability to press the right arrow key in the commit log to go to the commit details and left to go back from the commit details to the commit log.
I much prefer this way of navigating the log and commits, it feels much smoother.


extrawurst:

extrawurst:
@WizardOhio24 can you resolve those merge conflicts so that we can merge this?

WizardOhio24:
That should work (apologies it took me so long).

extrawurst:
I find this too intrusive of a change not to support 'enter' anymore. the middle ground is now in #548 

closing in favour of #548

extrawurst:
one problem is also that the home-brew auto test actually simulates a 'enter' command which would break the ci for Mac brew releases. it might also be a matter of taste for existing users

WizardOhio24:
? This still supports 'enter' in the same way it did previously (how does it not?), it just also adds the right and left arrow keys as an option, #548 seems only to support the left arrow key with the commit details open.  With this PR, if someone used 'enter' previously, the exact same thing will work (and right arrow will work as well).

WizardOhio24:
I much prefer only using the arrow keys, feels much smoother, rather than having to press enter when you want to see a commits details (but you can still use enter, that is not removed).

PR #424, opened 2020-11-13
Use theme for push gauge

### LINKED ISSUE #401: push gauge background is not default but explicit black
this should be fixed to not break with customised terminal color schemes

WizardOhio24:
Fixes #401 

This makes the push gauge red rather than black because explicit black looks odd on most terminals whereas red looks OK.

### DIFF
--- a/src/components/push.rs
+++ b/src/components/push.rs
@@ -23,7 +23,6 @@ use crossterm::event::Event;
 use tui::{
     backend::Backend,
     layout::Rect,
-    style::{Color, Style},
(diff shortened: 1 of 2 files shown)


extrawurst:
Do we even need any explicit bg? What if we keep bg as `Reset` and keep default `fg` as `White` to be less flashy?

WizardOhio24:
Tried that, the problem is that the text becomes invisible when the bar fills up.  Both fg and bg must be defined as something.

PR #426, opened 2020-11-14
Give better error message for editor enviroment variable if variable is empty

WizardOhio24:
This is link to a discussion in #419. 

This gives a better error message for opening an external editor.  This is also linked to #400, but is different because this gives a specific error message, saying the one of the environment variables must be empty.

### DIFF
--- a/src/components/externaleditor.rs
+++ b/src/components/externaleditor.rs
@@ -75,7 +75,7 @@ impl ExternalEditorComponent {
         let mut editor = editor.split_whitespace();
 
         let command = editor.next().ok_or_else(|| {
-            anyhow!("unable to read editor command")
+            anyhow!("Environment config variable for opening an editor is empty, check \"GIT_EDITOR\", \"VISUAL\" and \"EDITOR\" environment variables are not empty.")
         })?;
 
         let mut editor: Vec<&OsStr> =


pm100:
here is what git says

image

WizardOhio24:
I do not get that.  Look at the lines changed, this PR could not cause that, it only changes a string which would be output in the event of an error, where the error can only be if the environment variables are empty:
EditorError

To see this, use the command `GIT_EDITOR="" cargo run` and try to open the editor(for example, by committing then ^e).  What did you do to produce that error?  It is definitely unrelated to this considering what this changes, so there should be another PR for it.

WizardOhio24:
Wait, ignore my last comment, just realised you meant git cli.  

Well, that's fine for git to say that, but it is a cli and GitUI is a UI and so should try to be eaiser to use than git, that's the whole point.  And the error that git gives is not helpful if an environment variable is not set correctly.  This PR gives the user a better error message so they should be able to resolve the issue without ever having to go to the internet.  I imagine that if a git cli user saw that git cli error and was relatively unfamiliar with git, they would spend a bit of time searching until they found the answer.

extrawurst:

extrawurst:
closed in favour of 156776ab6822f605e908f6ba137175666741ea16

PR #431, opened 2020-11-22
add vc runtime to wix / msi package

pm100:
No longer run `cargo wix init.` This is because we need to hand edit the main.wxs file that init creates. So instead this is now checked in

The vc redist binary msm is checked in too (it is not directly downloadable from MS, only as part of VC install)


extrawurst:
@pm100 thanks for working on this. I merge this now and will ask for someone second with a windows machine to verify this works

PR #441, opened 2020-12-10
copy commit hash

### LINKED ISSUE #281: Copy commit hash in log tab and commit info component
It is nice to provide a commit hash copy feature when focus on log tab or commit info component.

yanganto:
fix #281


extrawurst:
@yanganto thanks for looking into this. but I think u also need to add it to the `commands`

yanganto:
@extrawurst Thanks for reviewing. :)

Issue #480, opened 2021-01-08
[Feature Request] flag to choose theme


Th3Whit3Wolf:
**Is your feature request related to a problem? Please describe.**

I have set my linux environment up to dynamically change the colorscheme, as much as possible,  between light and dark mode. With programs that allow you to set the colorscheme it's pretty simple to setup.

**Describe the solution you'd like**

Add a flag 
```shell
    -C, --color <path>    Selects a colors file to use
```

**Additional context**
I can submit a PR if you are open to this idea.

PR #481, opened 2021-01-09
Allow user to set theme with flag

Th3Whit3Wolf:
### This pull request allows user to set gitui theme by passing a flag (-t,  --theme). Address [Issue 480](https://github.com/extrawurst/gitui/issues/480)

The argument it takes is files a file relative to the app_config directory, meaning if the (linux) user passes `-t example.ron` to gitui then the theme located at `~/.config/gitui/example.ron` will be used. If `~/.config/gitui/example.ron` doesn't exist then `theme.ron` will be used.

This works by setting environment variable `GITUI_THEME` while processing command line arguments and reading it in function get_theme_file.

### DIFF
--- a/THEMES.md
+++ b/THEMES.md
@@ -3,11 +3,13 @@
 default on light terminal:
 ![](assets/light-theme.png)
 
-to change the colors of the program you have to modify `theme.ron` file
+to change the colors of the default theme you have to modify `theme.ron` file
 [Ron format](https://github.com/ron-rs/ron) located at config path. The path differs depending on the operating system:
 
 * `$HOME/Library/Application Support/gitui/theme.ron` (mac)
 * `$XDG_CONFIG_HOME/gitui/theme.ron` (linux using XDG)
 * `$HOME/.config/gitui/theme.ron` (linux)
+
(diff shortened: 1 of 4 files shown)


Th3Whit3Wolf:
Anything I can do to help get this merged?

extrawurst:
@Th3Whit3Wolf can't we pass the param along differently? it feels wrong to abuse a env variable as a global

Th3Whit3Wolf:
Definitely. I did it that way to have the smallest impact on the codebase. I'll make changes tomorrow.

Th3Whit3Wolf:
I made the changes. `process_cmdline` now return `Result<CliArg>`

```rust
struct CliArgs {
    theme: PathBuf,
}
```
I think this should allow gitui to incorporate more flags down the road with less friction.

theme then gets passed to app which passes it into `Rc::new(Theme::init(theme_path))`

Because `process_cmdline` checks for valid input and returns `theme.ron` if it isn't the get_theme_file  was no longer needed and the save function could be made smaller.

Th3Whit3Wolf:
Anything I can do to help get this merged?

extrawurst:
@Th3Whit3Wolf sorry the code looks fine, can you add some documentation about this into the `THEMES.md` (https://github.com/extrawurst/gitui/blob/master/THEMES.md)?

PR #506, opened 2021-02-04
Add search bar at bottom of revlog

### LINKED ISSUE #429: Search commits for message or author
It would be good to be able to search the commit list for words in the commit message, or in the author name(i.e filter by author).  For example, if you wanted to go back several hundred commits to find what a commit did, or a previous implementation, and you knew what the commit message contains, it would be good to just search for it, rather than scrolling through hundreds of commit messages to find that one commit.

Press a button, perhaps `f` or `s`, and be able to search the commit list for words.  This should search both author and commit message. 

Scrolling through hundreds of commits

### LINKED ISSUE #449: Search commits by sha
Unable to view a commit by its sha.

Hit s, then enter sha

Using Up/Down, PageUp/PageDown is cumbersome.

**Additional context**
Nice work! Thanks.

WizardOhio24:
Closes #429, Closes #449, (part of 1.0 todo list)

Adds a search bar at the bottom of the log tab which allows for searching commits.  It works, but I'm not sure what keys are best for accessing this.  Go to `log` tab then press `s` (suggested in #449), the search box will pop up and be in focus, type to search, the results will appear in the log where the arrow keys can scroll through them.  Press `esc` to close the search box, then `esc` again to cancel the search and return to all commits.

TODO:
 - [x] Async filtering
 - [x] Filter by sha (with `:s`)
 - [x] Filter by author (with `:a`)
 - [x] Filter by message   (with `:m`)
 - [x] Filter by tag (with `:t`) 
 - [x] Support for complement (i.e. not) filtering (with `:!`)
 - [x] Support for union filtering (with `||`)
 - [x] Support for intersection filtering (with `&&`)
 - [x] Support case sensitive filtering (with `:c`)
 - [x] Support brackets and nested brackets in search
 - [x] <del> Make a wrapper around `git_log` which supports filtering and have revlog use that instead of `git_log`</del>, not going to implement because it might not be better right now

---

Within the search:
 - `:s xxx` filters only shas for `xxx` 
 - `:a xxx` filters only authors for `xxx` 
 - `:m xxx` filters only messages for `xxx`
 - `:t xxx` filters only tags for `xxx`
These can be combined to `:sam`(any order works, so `:asm` or `:sma` works as well) to search for all, which is equivalent to putting nothing at the start

Complement filtering is possible with `:!`, which returns anything not containing what was searched for, for example to get all commits whose messages do not contain 'fix' or 'remove':
 - `:!m fix &&:!m remove`

Case sensitive filtering is possible with `:c`, for example, to find messages which contain 'Fix' and do not contain 'fix':
 - `:c Fix`

This also adds support for intersection filtering using `&&`(i.e only if 'x' and 'y'), for example, to find commits where author name contains 'richard' and the commits message has 'fix' in it:
 - `:a richard&&:m fix`

And support for union filtering using `||`(i.e or), for example to find any commits whose message contains 'fix' or 'move' or any commits which contain aabb':
 - `:m fix ||:m move || aabb`
(notice spaces at start and end are irrelevant, they are trimmed, but `:` must come before a space if filtering by something).

Intersection is currently always higher priority than union, for example, to find commit message which contain 'fix' and 'readme' or contain `move` = `:m fix&&:m readme || move`  which is equivalent to `((:m fix&&:m readme) || move )`.

And so, for a more complex query:
 - `:m fix && :m revlog ||:a richard &&:m status ||:s aabb ||:mc Remove || :!cm bigger`
 
For another example, show the commits which have tags where the commits contain 'pre' or ('fix' and 'CD'):
 - `:t &&(:m pre ||(:m fix&&:m CD))`

---
###  Some screenshots
Screenshot_20210213_154144

Screenshot_20210213_152955

Screenshot_20210213_153014

Screenshot_20210213_153851


WizardOhio24:
I'm thinking a good way to implement this could be: in the log tab press `:` to bring up the search bar then type to search.  The up and down arrows should still work in the log.  Then press `esc` to close the search.

extrawurst:
@WizardOhio24 I don't see any change in the asyncgit crate which makes me assume you only filter on the local items. I would assume to filter the entire log history and not only the one we store in the application

WizardOhio24:
Could you provide an example, when you search for the first commit sha, this is the result:
Screenshot_20210204_125734


The commit number at the top is incorrect, but the commit is correct.
Even just what I should search for, that the search doesn't find.

extrawurst:
revlog.rs(37):
```
static LIMIT_COUNT: usize = 3000;
```

so we only ever have 3k revision in memory at once. checkout a repo that has more than that and try to find something old, then it won't be searched.

WizardOhio24:
Tried `:`, but now think `s` is better.  Here's how it works:

 Go to log tab
 Press `s` (it's also a command at the bottom)
 Type to search (updates dynamically)
 Press `esc` to close the search box
 Press `esc` again (now on log tab), to cancel search.

That seems to work reasonably well.

WizardOhio24:
And it now filters all commits (I think), which might be expensive, so it only does it once when beginning a search, if it must.

WizardOhio24:
Everything seems to work in this now, just need to stop the help message opening when typing into the text input.

src/keys.rs
```diff
@@ -85,7 +87,9 @@ impl Default for KeyConfig {
 			exit: KeyEvent { code: KeyCode::Char('c'), modifiers: KeyModifiers::CONTROL},
 			exit_popup: KeyEvent { code: KeyCode::Esc, modifiers: KeyModifiers::empty()},
 			open_commit: KeyEvent { code: KeyCode::Char('c'), modifiers: KeyModifiers::empty()},
-			open_commit_editor: KeyEvent { code: KeyCode::Char('e'), modifiers:KeyModifiers::CONTROL},
+            open_commit_editor: KeyEvent { code: KeyCode::Char('e'), modifiers:KeyModifiers::CONTROL},
```
extrawurst, line 90: please don't change the indentation format (gitui uses tabs)

extrawurst:
The solution is not to fetch all instead of adhering to the 3k limit. This is what makes gitui uniquely fast. Try your solution on the linux repo (1m commits).
We rather need to do the filtering inside asyncgit when walking the revisions.

WizardOhio24:
So, when getting the commits, filter them, then when there is enough to fill the screen, stop getting commits?

extrawurst:
why would we stop at any point? there might be more commits down the history that match, right?

WizardOhio24:
Yeah, no, what I mean was just filter the commits as they are being loaded, until there is enough to fill the log tab commit list, which has about 10-20 commits (for me).  But then, it could still be the case that it has to go through every commit, if a term is sparsely mentioned, or not there at all, which could cause it to lag.

WizardOhio24:
How about a loading feature that continues to go through the commits in the background, 1k at a time and adds them onto the list if they fit the search terms?  Then all of them will eventually appear, but the most recent will appear first (which is most likely what the user is wanting).

extrawurst:
> Yeah, no, what I mean was just filter the commits as they are being loaded, until there is enough to fill the log tab commit list, which has about 10-20 commits (for me). But then, it could still be the case that it has to go through every commit, if a term is sparsely mentioned, or not there at all, which could cause it to lag.

Looks like you did not check how the revlog parsing works. this is already happening asynchronous in a thread. it should not *lag*.

> How about a loading feature that continues to go through the commits in the background

that's exactly what happens already

WizardOhio24:
I did not realize,  but then if this already happens, what do you want implemented?  With adding `.filter` to `entries.iter() (asyncgit/src/revlog, line 159)` or even taking the commits in the `logwalker (asyncgit/src/sync/logwalker)`, then running get commit details over them all, would it not be the same as doing it at a higher level, if not faster at a higher level because then we're only getting the commit details once?  I've still to optimize the higher level code, but, I don't understand why filtering as the commit IDs are being loaded would be faster than filtering in the `commitlist (src/components/commmitlist)`.

WizardOhio24:
But, I don't understand this, when I change the `LIMIT_COUNT (asyncgit/src/revlog, line 37)` to e.g 5, it doesn't load any more, at all.

WizardOhio24:
Ok, so there's a few ways to do the filtering, first is to filter all of the commits loaded in, but this would cause the program to slow down on very large repos, e.g that have 1 million commits (this is how it was implemented initially), second is to restart the logwalker on every new filter, but this has the problem that then all the commits must be re-loaded into the program after the filtering is finished, which isn't good as it could take a while on large repos.  

Third, and the way I think is best, is to create an async log filterer which runs through the local commits in an async manner, so the commits can continue loading in and the filter can run without blocking the program and when finished (i.e the filter is empty) can return to the commit list.

WizardOhio24:
This now works, filters async, but I'm not sure if it would be better to create a wrapper around `git_log` in GitUI which contains the log and the filterer, then the `revlog` interaction with `git_log` is only through this wrapper, which would probably be simpler and better.  Need to consider that before ready for review.

WizardOhio24:
Finally finished the search, give me a few days to look it all over, check code is good, then I'll mark it ready for review.  But feel free to test it right now, from my view, it's functionality is finished (and all working correctly).

extrawurst:
@WizardOhio24 can you rebase this onto master?

Cargo.toml
```diff
@@ -40,6 +40,7 @@ serde = "1.0"
 anyhow = "1.0.38"
 unicode-width = "0.1"
 textwrap = "0.13"
+parking_lot = "0.11"
```
extrawurst, line 43: what reasoning went into using `Mutex` from `parking_lot` vs using the regular `Mutex` from `std` that we use everywhere else?

src/keys.rs
```diff
@@ -208,27 +210,21 @@ impl KeyConfig {
             | KeyCode::BackTab
             | KeyCode::Delete
             | KeyCode::Insert
-            | KeyCode::Esc => {
```
extrawurst, line 211: what changed here?

src/tabs/revlog.rs
```diff
@@ -166,6 +202,160 @@ impl Revlog {
             tags.and_then(|tags| tags.get(&commit).cloned())
         })
     }
+
+    fn get_what_to_filter_by(
```
extrawurst, line 206: this giant method could use some doc and test

src/components/utils/async_commit_filter.rs
```diff
@@ -0,0 +1,404 @@
...
+        self.filter_thread_mutex = Arc::new(Mutex::new(()));
+
+        let cur_thread_mutex = Arc::clone(&self.filter_thread_mutex);
+        self.is_pending_local.replace(true);
+
+        let tags =
+            Self::get_tags(&filter_strings, &mut self.git_tags)?;
+
+        rayon_core::spawn(move || {
+            // Only 1 thread can filter at a time
```
extrawurst, line 281: we use atomics in the other async bits to limit only one running at a time

src/components/utils/async_commit_filter.rs
```diff
@@ -0,0 +1,404 @@
...
+
+    /// `filter_strings` should be split by or them and, for example,
+    ///
+    /// A || B && C && D || E
+    ///
+    /// would be
+    ///
+    /// vec [vec![A], vec![B, C, D], vec![E]]
+    #[allow(clippy::too_many_lines)]
+    pub fn filter(
```
extrawurst, line 92: this needs unittesting

Cargo.toml
WizardOhio24: This was because I think it's faster and it's already used as a dependency in some of the crates GitUI uses, happy to use the normal mutexes instead though.

extrawurst: yeah lets keep it consistent

extrawurst:
closing in favour of https://github.com/extrawurst/gitui/pull/672

c02y:
I have been using tig for years, now I'm considering switching to lazygit or gitui, don't know which one is better, googled for a while, except for some benchmarks make two years ago, didn't find any useful info, so I tried to use both of them and decide by myself, just try gitui, didn't find a simple search function anywhere, it should be anywhere inside the session like tig or lazygit, so how to can find something in some tab like the list of log?

Sorry, I'm switching to lazygit, or maybe back to tig if I'm not satisfied with lazygit.

PR #507, opened 2021-02-06
Add popup if upstream branch not set when pushing

### LINKED ISSUE #492: push to specific remote
followup to #491:

in case of no remote being named `origin` and multiple remotes being known we currently pick the *first*.
we should rather open a popup and have the user select from the list of remotes.

WizardOhio24:
Closes #492

Adds a popup on push with the list of possible remote branches if no remote branch is set.  At the bottom of the list is the name of the branch which may exist in remote repo but which can be pushed (-u) and added if it is not there.  This allows the user to select one of the remote branches which are connected to their repo or push their own branch name. 
Screenshot_20210206_215743

Would it be better to not list all the remote branches?  i.e only list `origin/<branch_name>`, `<some_remote>/<branch_name>`, or do it the way it is right now, with all the branches?
Perhaps a drop down list, where `origin` and `<some_remote>` could be selected to push to them with the branch name, or they could be dropped down to selected a specific branch.

And should the user be able to set the branch upstream by pressing, e.g `U` on in the status tab to bring up the upstreams?  I think a dedicated branch tab for dealing with local/remote branches would be better, but would be happy to add a `Set upstream [U]` button.

Any suggestions on how to handle this would be appreciated.


extrawurst:
please refer to #491 for the details, the point is to allow selecting the remote not the branchname - a list of all available remotes is sufficient to fix #491

Issue #478, opened 2021-01-06
Warn when commit message is suspiciously long


savchenko:
**Is your feature request related to a problem? Please describe.**
Users might write extra-long commit messages which frustrate variety of workflows.
Consider someone pushing the following:
![image](https://user-images.githubusercontent.com/300146/103740901-06fc5680-5048-11eb-8ebe-9095d9810061.png)

**Describe the solution you'd like**
- Limit 1st line to 50 symbols
- Keep 2nd line empty
- Wrap the rest at 72 characters

**Describe alternatives you've considered**
Format each commit manually.

**Additional context**
N/A

extrawurst:
problem with this is that it is not well defined what limits are correct and I am shying away of requiring the user to fiddle in more config files.

the most common way to assure stuff like this is having (pre)commit-hooks

extrawurst:
to reiterate my feedback from the #508:

I am not a fan of this requiring new configuration, abusing the style config I not an option.

my suggestion for an MVP is:
* we hardcode the first-line-limit to 50
* we show a red message (`first line exceeds x chars`) on the bottom (of the dialog, left of the count) if the limit is exceeded

extrawurst:
looks like this now:
![Screenflick Movie 5](https://user-images.githubusercontent.com/776816/116701217-f0e41e80-a9c7-11eb-899b-741a212148af.gif)

savchenko:
I wonder how the 2nd line has appeared. Have you merged/implemented shift+enter somewhere in-between? :)

extrawurst:
> I wonder how the 2nd line has appeared. Have you merged/implemented shift+enter somewhere in-between? :)

that's what the external editor support is currently used for:
<img width="136" alt="image" src="https://user-images.githubusercontent.com/776816/116701824-a616d680-a9c8-11eb-8e54-414faa5553c3.png">

savchenko:
> edit: that is until #509 is merged

Subscribed. Thank you!

PR #508, opened 2021-02-08
Show commit message box in red if first line > 50 characters

WizardOhio24:
Linked to #478

Shows commit box in red if the first line is greater than 50, this can be configured in the theme and can be turned off entirely but 50 is the current default (Should it be off by default?).  If there were more options, there should probably be a separate config file, but because this is aesthetic only (and this would be the only config option) it could fit in theme.  

GitUI currently does not support making multi-line commit messages in the commit message box, so I'm not sure if there is a point in making the box red for multi-line messages which are unconventional (such as no empty 2nd line or >72 characters in the next paragraph as suggested in #478).

Screenshot_20210208_201559
Screenshot_20210208_201644


extrawurst:
1) as stated in #478 I am not a fan of this requiring new configuration, especially since we now abuse the style file for it.
2) making the entire box red is too much. the count in the corner makes more sense
3) the unofficial limit of 50 only applies to the first line. we show the total char count, my suggestion would be to show an additional text on the bottom saying `first line exceeds X chars`

so long story short my required MVP for this:
* we hardcode the first-line-limit to 50 
* we show a red message on the bottom (of the dialog, left of the count) if the limit is exceeded

for the time being I mark this PR as *draft*

WizardOhio24:
Now shows like this:
Screenshot_20210429_233225
:

Let me know what you think (I think this is visible enough, but not too visible). (This PR is still a draft, just have to clean up some code, I'll let you know when it's ready for review).

extrawurst:
@WizardOhio24 still the whole message length is counted:

image

> so long story short my required MVP for this:
>
> * we hardcode the **first-line-limit** to 50
> * we show a red message on the bottom (of the dialog, left of the count) if the limit is exceeded

PR #509, opened 2021-02-08
Add new line support for multi-line input box

WizardOhio24:
Adds multi-line support for multi-line input boxes, in particular the commit message box:
Screenshot_20210208_204919

This is done with ALT+Enter, because SHIFT+Enter does not work in terminals and so does not work in crossterm.  

The support is simple because it can sometimes be useful to add an extra line quickly and would be useful (and easier) to do in GitUI, but GitUI is not a text editor and so for large multi-line commit messages, a proper text editor should be used.


extrawurst:
Can you make this showup in the commands section aswell then?

WizardOhio24:
Done
Screenshot_20210218_193057

WizardOhio24:
I'm unsure the alt symbol is that clear, but should be fixable with #516
Screenshot_20210218_194626

extrawurst:
I just checked and it unfortunately does not work with this key binding on a Mac. it just commits, no line break. I'd like to find a binding that works on all platforms out of the box

extrawurst:
modifier keys+enter will not work reliably anyway (I think its a known limitation in crossterm)

terhechte:
What about `Enter` inserts a newline and `^c` (ctrl c or alt c) commits (just like `^a`) amends?

extrawurst:
> What about `Enter` inserts a newline and `^c` (ctrl c or alt c) commits (just like `^a`) amends?

Thats a good idea!

WizardOhio24:
`^c` is to exit GitUI, so I'd be worried the user could try to exit and instead commit whatever was in the textbox so right now I've changed it to `alt+c`, but that seems awkward. With nano it would be `^x`, would that be better?  (The user can always change it to what they want in the config) Happy to go with whatever people want.

extrawurst:
`Option+c` does not work either on a Mac..
image

at this point only something `Ctrl + ...` seems to be a workable option. lets start with `Ctrl + m` and gather some feedback (x with Ctrl breaks my wrist ^^)

extrawurst:
@WizardOhio24 see https://github.com/extrawurst/gitui/pull/645

after playing around with this PR I realize we cannot merge this without also adding scrolling to the `TextInputComponent` otherwise after adding a couple of newlines the user cannot see the cursor anymore

WizardOhio24:
Yeah, I mean, this was more for writing a commit message in a single line (i.e 50 chars) then adding a new line and putting some other detail (such as issues it was linked to).  If the user wanted to put in a lot of text it would be better to use nano or vim (or micro), this is just for small commits where 2-3 lines are needed.

extrawurst:
@WizardOhio24 thanks for updating this branch. still we can do better and add the scrollbar code we use all over the place already here aswel. this makes the cursor stay in view even when adding newlines.

WizardOhio24:
This is emitting a warning in clippy about `intersperse` (on iterators), this is fine, both functions will do the same thing.  All this is saying is that in the future, intersperse will be included in Rust and so will not use the itertools version.  It will go away when intersperse is in stable Rust.

extrawurst:
> This is emitting a warning in clippy about `intersperse` (on iterators), this is fine, both functions will do the same thing. All this is saying is that in the future, intersperse will be included in Rust and so will not use the itertools version. It will go away when intersperse is in stable Rust.

please look for `get_text_info` in `details.rs` on how to explicitly allow this. we wanna be warning free (this turned to be an error now on master)

src/components/textinput.rs
```diff
@@ -143,11 +171,16 @@ impl TextInputComponent {
...
         let mut txt = Text::default();
+
         // The portion of the text before the cursor is added
         // if the cursor is not at the first character.
         if self.cursor_position > 0 {
-            let text_before_cursor =
-                self.get_msg(0..self.cursor_position);
+            let text_before_cursor: String = self
+                .get_msg(0..self.cursor_position)
+                .lines()
```
extrawurst, line 180: we can derive `scroll_max` from lines, we do not need it to be a variable in `self`

src/components/textinput.rs
```diff
@@ -336,6 +384,13 @@ impl Component for TextInputComponent {
                 if e == self.key_config.exit_popup {
                     self.hide();
                     return Ok(EventState::Consumed);
+                } else if e == self.key_config.enter
+                    && self.input_type == InputType::Multiline
+                {
+                    self.msg.insert(self.cursor_position, '\n');
```
extrawurst, line 390: lets make this a `fn insert_new_line`

src/keys.rs
```diff
@@ -227,27 +229,21 @@ impl KeyConfig {
...
+            KeyCode::Char(c) => format!(
+                "{}{}",
+                Self::get_modifier_hint(ev.modifiers),
+                c
+            ),
+            KeyCode::F(u) => format!(
+                "{}F{}",
+                Self::get_modifier_hint(ev.modifiers),
+                u
+            ),
```
extrawurst, line 246: what changed here? if its only whitespaces, can this be reverted?

src/components/textinput.rs
```diff
@@ -327,6 +449,13 @@ impl Component for TextInputComponent {
             )
             .order(1),
         );
+
+        out.push(CommandInfo::new(
+            strings::commands::commit_new_line(&self.key_config),
+            true,
```
extrawurst, line 455: this should be conditional, right?

WizardOhio24:
This now grows the commit box as new lines are added up to a limit, at which it stops growing and starts using a scrollbar, which seems better when typing out multiline commits.

extrawurst:
typed around a bit:
```
07:17:06 [ERROR] panic: PanicInfo { payload: Any, message: Some(attempt to subtract with overflow), location: Location { file: "src/c
omponents/textinput.rs", line: 148, col: 13 } }
trace:
[...]
  11: gitui::components::textinput::TextInputComponent::decr_cursor_multiline
             at src/components/textinput.rs:148:13
  12: gitui::components::textinput::TextInputComponent::decr_cursor
             at src/components/textinput.rs:163:13
  13: gitui::components::textinput::TextInputComponent::backspace
             at src/components/textinput.rs:286:13
```

also `up/down` behaves weird when having empty lines in between:

Screenflick Movie 18

the scrollbar is also not keeping the cursor in view.
instead of adding new features like *auto growing* lets focus on getting one thing working first please.

src/components/textinput.rs
```diff
@@ -89,20 +99,155 @@ impl TextInputComponent {
...
+
+        // if the text box height increased,
+        // componsate by scrolling up one
+        if self.scroll_max
+            < (self.frame_height.get() as usize)
+                .saturating_sub(BORDER_SIZE * 2)
+            && self.scroll_max >= 3
+        {
+            self.scroll_top = self.scroll_top.saturating_sub(1);
+            //self.cur_line = self.cur_line.saturating_sub(1);
```
extrawurst, line 118: remove commented code please

src/components/textinput.rs
```diff
@@ -89,20 +99,155 @@ impl TextInputComponent {
...
+        }
+    }
+
+    /// Move the cursor up a line.
+    /// Only for multi-line textinputs
+    fn line_up_cursor(&mut self) {
+        let mut nearest_newline: usize = 0;
+        let mut prev_line_newline_loc = 0;
+        for (i, c) in self.msg.chars().enumerate() {
+            if c == '\n' {
```
extrawurst, line 173: how does this behave on windows where a newline is two chars wide?

src/components/textinput.rs
```diff
@@ -89,20 +99,155 @@ impl TextInputComponent {
...
+                chars_not_printed = 0;
+                prev_line_newline_loc = nearest_newline;
+                nearest_newline = i;
+                if nearest_newline > self.cursor_position {
+                    break;
+                }
+            }
+            // if !self.msg.is_char_boundary(i) {
```
extrawurst, line 217: commented code

src/components/textinput.rs
```diff
@@ -89,20 +99,155 @@ impl TextInputComponent {
...
+            // self.msg.is_char_boundary(i) c.is_alphanumeric() {
+            // unprintable
+            //    chars_not_printed += 1;
+            // }
+        }
+        self.cursor_position = self
+            .cursor_position
+            .saturating_sub(prev_line_newline_loc)
+            .saturating_add(nearest_newline);
+        // .saturating_add(chars_not_printed);
```
extrawurst, line 227: see above

src/components/textinput.rs
```diff
@@ -282,7 +452,22 @@ impl DrawableComponent for TextInputComponent {
                 InputType::Multiline => {
                     let area = ui::centered_rect(60, 20, f.size());
                     ui::rect_inside(
-                        Size::new(10, 3),
+                        Size::new(
+                            10,
+                            min(
```
extrawurst, line 457: there is `clamp` for this

src/strings.rs
```diff
@@ -584,16 +584,28 @@ pub mod commands {
...
     }
+    pub fn commit_new_line(
+        key_config: &SharedKeyConfig,
+    ) -> CommandText {
+        CommandText::new(
+            format!(
+                "New Line [{}]",
+                key_config.get_hint(key_config.enter),
+            ),
+            "make a new line in the commit message",
```
extrawurst, line 605: `insert newline`

src/components/textinput.rs
```diff
@@ -89,20 +99,155 @@ impl TextInputComponent {
...
+    }
+
+    /// Move the cursor down a line.
+    /// Only for multi-line textinputs
+    fn line_down_cursor(&mut self) {
+        //
+        let mut nearest_newline: usize = 0;
+        let mut prev_line_newline_loc = 0;
+
+        let mut chars_not_printed = 0;
```
extrawurst, line 206: variable is never used (read)

src/components/textinput.rs
WizardOhio24: I was trying to get the scrolling working with characters which occupy more than 1 byte, such as ö or é. This code should have worked, but it doesn't, nothing seems to.

src/components/textinput.rs
WizardOhio24: This is part of getting it working with multi-byte chars.

src/components/textinput.rs
extrawurst: No, google ‚windows line endings‘

extrawurst:
closing in favour of #1309

Issue #465, opened 2020-12-26
allow customising key symbols


extrawurst:
#464 showed that we should allow overwriting/customizing the symbols we use to show key codes. this can be added similarly to our general key configuration in `keys.rs` see `get_key_symbol`

josmith42:
I think I'd like to take a stab at this.

Question: in addition to implementing the feature, would you like the PR to also update the key config documentation so users know how to customize the symbols?

extrawurst:
docs for this: https://github.com/extrawurst/gitui/blob/master/KEY_CONFIG.md#key-symbols

PR #516, opened 2021-02-13
Allow customising key symbols

josmith42:
* Allows users to customize the symbols used for showing key codes
* Customized symbols are stored in `key_config.ron`
* Example symbols config file added in `assets/alternate_key_symbols.ron`
* Addresses issue [465](https://github.com/extrawurst/gitui/issues/465)


extrawurst:
@josmith42 sorry for taking so long to check this out. one thing that bothers me here is that this increases the key.ron file further. how about making the symbols a separate ron file?

josmith42:
@extrawurst Yes, that makes sense. I had actually considered making the symbols a separate file when I did this originally. I don't remember why I decided not to, but I agree that's probably the best way to do it. I'll work on making the change and add another commit to this PR.

extrawurst:
I will mark this as draft in the meantime, feel free to switch it back once you think its ready for a round 2

josmith42:
@extrawurst I have put the symbols settings into a separate ron file. I did this by making a separate `SymbolConfig` struct and adding it as a non-serializable field in `KeyConfig`. This seemed like the best way to implement it that wasn't too invasive. Please let me know your feedback. Thank you!

extrawurst:
@josmith42 please run `make check` locally before pushing to reduce the turn around times due to broken CI

josmith42:
@extrawurst My apologies. Should be fine now.

src/keys.rs
```diff
@@ -19,6 +19,110 @@ use crate::args::get_app_config_path;
...
+{
+    let mut f = File::open(config_file)?;
+    let mut buffer = Vec::new();
+    f.read_to_end(&mut buffer)?;
+    Ok(ron::de::from_bytes(&buffer)?)
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct SymbolConfig {
+    pub enter_symbol: String,
```
extrawurst, line 52: I think we can ditch the `_symbol` postfix considering the struct name

src/keys.rs
```diff
@@ -81,6 +185,9 @@ pub struct KeyConfig {
     pub force_push: KeyEvent,
     pub pull: KeyEvent,
     pub abort_merge: KeyEvent,
+
+    #[serde(skip)]
+    symbols: SymbolConfig,
```
extrawurst, line 190: what's the reasoning to put this in here and skip in serde even though we use serde to serialise it?

assets/alternate_key_symbols.ron
```diff
@@ -0,0 +1,20 @@
...
+    enter_symbol: "ENTER",
+    left_symbol: "◀",
+    right_symbol: "▶",
+    up_symbol: "▲",
+    down_symbol: "▼",
+    backspace_symbol: "BACK",
+    home_symbol: "HOME",
+    end_symbol: "END",
+    page_up_symbol: "PgUp",
+    page_down_symbol: "PgDn",
```
extrawurst, line 11: Whats the reasoning for ALLCAPS vs. CamelCase?

src/keys.rs
josmith42: So I did the skip here so that the symbols don't show up in `key_config.ron`, since that file is already getting big. I'm serializing the symbols elsewhere in a separate file.

I'm not terribly fond of this implementation, though. `symbols` has to have default values when the key config is read from `key_config.ron` before it's overwritten with the data in `symbol_config.ron`. The other option I thought of was to make a struct that contains a `KeyConfig` object and a `SymbolConfig` object. That could then be passed around where the current `KeyConfig` struct is. I started going down this route, but this would have ended up touching nearly every file in the code base, because `KeyConfig::get_hint()` would need the data from both objects.

assets/alternate_key_symbols.ron
josmith42: None, other than I had in my head that it was "standard". Looking at my personal keyboard, though, it's totally not. I'll make it CamelCase.

src/keys.rs
josmith42: Done.

josmith42:
@extrawurst I addressed the issues you raised in the most recent commit. Please let me know of anything else, and if my explanation of why I implemented it the way I did makes sense.

extrawurst:
superseded by https://github.com/extrawurst/gitui/pull/997

PR #520, opened 2021-02-18
Add GitUI version at top right

### LINKED ISSUE #443: Feature: GitUI splash screen
It would be cool if GitUI opened up with a splash screen (that just displays for a few seconds before going into the actual GitUI application).

Essentially GitUI should appear in ASCII blocks in the center of the terminal for 1-4 seconds(whatever works best), then should transition to the application.  This could be annoying for users who open and close GitUI often so there should also be a way to turn this off.
Below is an example of what this could look like
```
        ████████████████████████████████████████████████████████████████████
  
        ████████████  ███████████  ███████████  █████     █████  ███████████
        ████████████  ███████████  ███████████  █████     █████  ███████████
        ████             █████        █████     █████     █████     █████
        ████    ████     █████        █████     █████     █████     █████
        ████     ███     █████        █████     █████     █████     █████   
        ████████████  ███████████     █████     ███████████████  ███████████
        ████████████  ███████████     █████     ███████████████  ███████████
        
        ████████████████████████████████████████████████████████████████████
```
It would be good for the version to be shown here as well(somehow).
This will also need to determine if the terminal size is big enough to support this width/height of characters and if not perhaps not show it (or show something different?).

WizardOhio24:
Closes #443

Adds `GitUI vX.Y.Z` version at the top right (a bit difficult to see with the konsole theme).
Screenshot_20210218_191626

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -192,7 +192,19 @@ impl App {
 
         self.cmdbar.borrow().draw(f, chunks_main[2]);
 
-        self.draw_tabs(f, chunks_main[0]);
+        let tab_chunks = Layout::default()
+            .direction(Direction::Horizontal)
+            .constraints(
+                [Constraint::Min(57), Constraint::Length(16)]
+                    .as_ref(),
+            )
+            .split(chunks_main[0].inner(&Margin {
+                vertical: 0,
+                horizontal: 1,
+            }));
+
+        self.draw_tabs(f, tab_chunks[0]);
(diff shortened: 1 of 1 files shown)


extrawurst:
Like stated in https://github.com/extrawurst/gitui/issues/443#issuecomment-756379073 I see zero need for this - sorry that you wasted your effort on this :(

extrawurst:
if I miss something lets discuss further on discord

PR #536, opened 2021-02-23
Fix MacOS force push tests sometimes failing

### LINKED ISSUE #534: flaky test_force_push_rewrites_history failures
@WizardOhio24 unfortunately `test_force_push_rewrites_history` fails randomly. do you mind looking into the CI pipelines and figure out why this is the case? otherwise I have to revert the *force-push* PR :(

```
failures:

---- sync::remotes::tests::test_force_push_rewrites_history stdout ----
thread 'sync::remotes::tests::test_force_push_rewrites_history' panicked at 'assertion failed: `(left == right)`
  left: `false`,
 right: `true`', asyncgit/src/sync/remotes.rs:536:9


failures:
    sync::remotes::tests::test_force_push_rewrites_history
```

WizardOhio24:
Closes #534 

Test this a few times before merging, it seems only the MacOS tests are effected.
Just opening this at the moment so the CI will run the test.  Will mark it ready for review when it's ready.

### DIFF
--- a/asyncgit/src/sync/remotes.rs
+++ b/asyncgit/src/sync/remotes.rs
@@ -533,6 +533,8 @@ mod tests {
             true
         );
 
+        std::thread::sleep(std::time::Duration::from_millis(20));
+
         assert_eq!(
             upstream
                 .find_commit((other_repo_commit_ids[0]).into())
--- a/asyncgit/src/sync/remotes.rs
+++ b/asyncgit/src/sync/remotes.rs
@@ -409,7 +409,6 @@ mod tests {
     }
 
     #[test]
-    #[ignore]
     fn test_force_push_rewrites_history() {
         use super::push;
         use std::fs::File;
@@ -536,7 +535,7 @@ mod tests {
 
         assert_eq!(
             upstream
-                .find_commit((commit_ids[0]).into())
+                .find_commit((other_repo_commit_ids[0]).into())
                 .unwrap()
                 .parents()
                 .next()


WizardOhio24:
Seems to work, it was because the upstream was being checked before it was able to process the incoming push, i.e the MacOS was going too fast.  Do run the CI on this a few times just to make sure, but I believe it's fixed.

extrawurst:
I doubt it works: https://github.com/extrawurst/gitui/pull/535/commits/7c35577f173910f92a5631894a0b9062681461a4

without this the test does not even push any the files

extrawurst:
regarding the timing: there must be a way to make this more reliable than *waiting a few secs*

WizardOhio24:
True, forgot to stage the files, although, does there need to be files?  It's really the commits ids that need to be different, so I guess the files are unnecessary.

WizardOhio24:
With the timing, I guess it's because the `find_commits` executes before the `push` has made it's changes/been received. Perhaps `ProgressNotification` on `push` can be used and when it's finished, then continue to check the commit ids are different?

extrawurst:
push is not async. can you explain what the last assert is supposed to test actually?

extrawurst:
that last assert basically checks if the rewritten upstream still has the same first commit id that it had before the rewrite which should not be the case. I dont think this PR fixes the test flakiness. 

but I agree the unittest is very verbose for what it actually does: testing if rewriting the entire history of a repo works via *force*.
the fact that you use two repos looks like you actually want to test the diamond shape problem where two repos with the same state commit two different things and we test the force outcome - this is not happening here though

WizardOhio24:
So, the two repos push to the upstream, the first one normally, then the second one with a (required) force push.  Because these were pushing to the same branch, but different histories(diverged), the last test checks that the commit from the force push is now in the upstream repo and the parent is the first commit.  Because a normal push tried and failed and a force push was required, if this commit is here and the parent is correct then the force push worked.  If the commit is not in the upstream, the force push failed (i.e this checks that history was re-written).

Because these are separate repos, waiting a short amount of time allows the push to be received in the upstream and for the upstream to make it's changes internally (previously `find_commits` was being executed before the upstream had received the push).  So rather than waiting a short amount of time, you could also use the push progress sender to receive a notification from the upstream to tell the test that the push has completed successfully (I think).  

But I agree this test needs sorted out, I'm happy to fix this, or you can if you want.

extrawurst:
> the last test checks that the commit from the force push is now in the upstream repo and the parent is the first commit.

those are the last two asserts, the very last tests for the wrong parent. you do not keep the original first commit in upstream since you rewrite history to remove the first and the second commit. both `repo` and `other_repo` have nothing to do with each other, dont share a common commit or any history. so checking against an ID from another repo fails naturally. 

the only reason this does work to a degree is that for some reason the parent of the first commit we do in the test in the test repo that `repo_init` creates (including a empty first commit) has the same commit hash *most of the time* but relying on this is of course bad and makes the test flaky.

my fix in #535 works but it only cures the symptoms of the bad tests. I should have paid more attention in the review. closing this in factor of the other PR

PR #549, opened 2021-02-25
Fix vim default config

### LINKED ISSUE #512: "Open editor" shortcut doesn't work (using vim bindings)
**Describe the bug**
When Commit dialog is open and custom hotkey for "Open editor" is defined, pressing the hotkey does nothing.

**To Reproduce**
1. Stage something
2. Commit
3. Type some text, press corresponding shortcut to open an editor
4. Nothing happens

**Expected behavior**
Editor opens.

**Screenshots**
image

 - OS/Distro + Version: Debian Bullseye
 - GitUI Version: 0.11.0
 - Rust version: 1.48.0

**Additional context**
"Open editor" function within the main window works as expected. I have tried various key combinations for both.

yanganto:
Fix #512

### DIFF
--- a/assets/vim_style_key_config.ron
+++ b/assets/vim_style_key_config.ron
@@ -28,7 +28,9 @@
     exit_popup: ( code: Esc, modifiers: ( bits: 0,),),
 
     open_commit: ( code: Char('c'), modifiers: ( bits: 0,),),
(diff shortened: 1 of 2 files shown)


extrawurst:
Good catch with the command text, I did not even see that! 👍🏻

PR #551, opened 2021-02-25
Update Readme

yanganto:
Update installation guide for Nix/NixOS


extrawurst:
cool thanks! 
how many package manager does linux have? :D

yanganto:
No matter which unix like OS you use you can always use Nix as package management including MacOSX :)

Issue #594, opened 2021-03-20
Why were `focus_workdir` and `focus_stage` removed?


terhechte:
**Describe the bug**
The recent merge of [#575](https://github.com/extrawurst/gitui/commit/b5ef9b10f1e0e6e0c21a55d9083b4e40c1f759b8#diff-dd5ba1bca5a2e982ec6f7d24bac5d1b5588e12912d838585931ce9547411cccb) removed the ability to quickly switch between stage and workdir. I rely on these in my daily work, for example when I accidentally stage something and want to unstage it. Now I have to scroll to the bottom of the unstaged which takes forever. 
I understand that the `s` key is used for stage / onstage lines now but it would have been great to keep `focus_workdir` / `focus_stage` at least in the `key_config.ron` so I can use it. If needed I can help out with a simple PR that reintroduces it via the old, removed functions from #575.

extrawurst:
Hi @terhechte I am not a fan of duplicated functionality, if find myself and users I talked to mostly use the up/down keys to navigate. if you have a very long list of items to scroll down to switch to the stage there is `Shift+Down` to go to the bottom right away. Are you sure this won't work for you?

Iff we re-add these special keys, we are going to need a fitting shortcut: `[s]tage` is taken, `[I]ndex` is taken. Any ideas?

terhechte:
The downside of Shift+Down is that I am loosing my current position / selection in Unstaged. Example: I'm staging a whole file as a series of some files. But then I realize I want to onstage it and only stage some lines. So I need to go to staging, unstage it, then go back to Unstaged and continue what I was doing. I'd prefer going back to where I was, not first having to go to the bottom of Unstaged because then I maybe have to hit arrow up 5-10 times until I'm back at the original position.

I do get that not everybody works this way, after all we're all peculiar in our own ways. This function doesn't even need two shortcuts, does it? In the end the user is just going back and forth between Workdir and Stage, so it could be one key that toggles. What about `g` for "Go" (sadly, `t` for toggle is used for tags). Or `w` for S**w**itch?

extrawurst:
The *'not losing where I was'*-part makes sense me! 
I like the *toggle* idea, lets take `g` for that. I am happy to accept a PR from a fellow Hamburger on this :)

terhechte:
[...]

PR #575, opened 2021-03-08
Stage/unstage lines

extrawurst:
see #59

PR #595, opened 2021-03-20
Add `g` to toggle between staging and workdir (`workarea`)

terhechte:
As discussed in https://github.com/extrawurst/gitui/issues/594 This adds `g` to toggle between staging and workdir. Needed to find a good term, went with `work area` which would be either staging or workdir. Open for suggestions :)

Note that the `vim` key config uses `w` instead of `g` as that (kinda) matches the vim `w` for moving between windows. Whereas `g` in vim can be used for jumping to the top. This isn't used yet but I'd love to add it here in the future.


src/tabs/status.rs
```diff
@@ -236,6 +247,10 @@ impl Status {
         self.focus == Focus::Diff
     }
 
+    const fn can_toggle_workarea(&self) -> bool {
+        !matches!(self.focus, Focus::Diff)
```
extrawurst, line 251: let's reuse `!is_focus_on_diff()`

src/strings.rs
```diff
@@ -503,6 +503,18 @@ pub mod commands {
             CMD_GROUP_GENERAL,
         )
     }
+    pub fn toggle_workarea(
```
extrawurst, line 506: lets use two explicit methods here, one for `status_focus_workdir` and `status_focus_stage` then we can go back to the short terms `To Stage`/`To Unstated` like it was before (see https://github.com/extrawurst/gitui/pull/575).

src/tabs/status.rs
```diff
@@ -515,6 +530,16 @@ impl Component for Status {
                 self.can_focus_diff(),
                 (self.visible && !focus_on_diff) || force_all,
             ));
+            out.push(
+                CommandInfo::new(
```
extrawurst, line 534: lets use two different commands depending on where the focus is (see above)

extrawurst (changes requested):
added a few comments.

extrawurst:
how about using `w` in regular key bindings aswell, now that we have not strong reason to use `g` anyway?

terhechte:
Good Point ☺️

savchenko:
Would it make sense to allow <kbd>w</kbd> from the Diff view as well?
It currently does nothing and the only way to switch back is via <kbd>&larr;</kbd>.

extrawurst:
I do not think this would we intuitive. I prefer the current situation where you always go back and forth with `left`/`right` between the file and the matching diff of that file

PR #604, opened 2021-03-24
Fix commit_amend in vim style key config

silmeth:
The same problem as with uppercase <kbd>E</kbd> for opening editor – <kbd>Shift</kbd> cannot be used as modifier in commit-message editing floating window. Use <kbd>Ctrl</kbd> + <kbd>A</kbd> instead.


extrawurst:
@yanganto, @jonstodle I need some VI input on this one

jonstodle:
LGTM

PR #609, opened 2021-03-30
Fix build

Elbandi:
On ubuntu 18.04 LTS, there is a build error:

```
error: `std::option::Option::<T>::as_ref` is not yet stable as a const fn
   --> src/components/commitlist.rs:101:9
    |
101 |         self.tags.as_ref()
    |         ^^^^^^^^^^^^^^^^^^

error: aborting due to previous error
```

This patch is fix that.

### DIFF
--- a/src/components/commitlist.rs
+++ b/src/components/commitlist.rs
@@ -97,7 +97,7 @@ impl CommitList {
     }
 
     ///
-    pub const fn tags(&self) -> Option<&Tags> {
+    pub fn tags(&self) -> Option<&Tags> {
         self.tags.as_ref()
     }
 


extrawurst:
`gitui` requires the most recent stable rust. the pipelines will fail otherwise

Elbandi:
i know. but many ppl using LTS versions (company policy, tradition, etc), you exclude them from using this good program.

anyway, linux-musl build successful :)

extrawurst:
I am not linux person, but you can install a recent rust compiler even on an ubuntu LTS version, right?

Elbandi:
actually no. our policy "no hacky handmade install, only ubuntu dpkg/debs", like almost everyone.

extrawurst:
@Elbandi feel free to join discord to discuss this further, I still do not see the issue with installing a modern rust compiler

extrawurst:
as discussed on discord: we never cared for backwards compat with older rust versions. gitui is still in active pre-1.0 development and will try to keep msrv 1.50 from now on since eae99ea

PR #634, opened 2021-04-11
Add blame view

### LINKED ISSUE #484: file blame
in our file trees we should be able to jump into a file blame mode known from all major git GUI tools.

for a first mvp it is probably enough to scroll by line and show the hash + author + date each line in a split-screen side-by-side.

from each line we should allow opening the matching commit details.

cruessler:
This closes #484.

This is a rough draft intended to get initial feedback on the direction. To
test, you can press Enter when the selection in a file tree is on a file that
is tracked by git and you will see the blame for that file. I’ll also attach a
screenshot.

Known issues:

- [x] Highlight and scrolling is not yet implemented.
- [x] There is duplicated code that should be removed, in particular
  `string_width_align` which can likely moved to a better location somewhere in
  `utils`.
- [ ] The git operations have no async wrapper yet.
- [x] The way it is made sure that only one of `blame_file_popup` and
  `inspect_commit_popup` is shown can certainly be improved.
- [x] `BlameFileComponent::get_lines` is rather large, parts of it should
  probably be factored out.
- [x] `App::process_internal_events` is almost hitting clippy’s limit of 100
  lines per function.
- [x] There are no tests yet.
- [x] When you press Enter on a file that is not tracked by git, `gitui` exits
  with an error.

I will address these issues in the coming days.

Let me know what you think!


cruessler:
First draft of the blame view

extrawurst:
Really cool for a start! I agree with all your todo boxes.

just a few additional thoughts:
* I would argue the `FileBlame` should contain the lines of the file and the hunks should only reference line indices. this would be more memory efficient and easier to pipe the lines through a highlighter like `syntect` later (https://lib.rs/crates/syntect)
* the indentation seems off (see the `use` lines not lining up)
* `get_lines` always processes all lines not only the ones visible
* I'd like to see a `|`-line separator between the annotation part and the actual content + line numbers
* the title should contain the blamed file and its commit id (we probably later want to be able to blame files at other points in time in the history/revision for example when inspecting an older commit)
* we started using `unicode-truncate` crate now to do the truncation, you can remove a couple of helpers when switching to those

cruessler:
@extrawurst I’ll be working on this over the next days, and once I’ve addressed all the issues, I’ll update the PR.

cruessler:
Although I’m not yet done, many of the outstanding issues are now addressed. I was able to use `tui::widgets::Table` which made aligning the different pieces of information quite easy. `Table` can even manage large parts of the scroll state for us. I think I’m coming closer, but there still are a few issues on my list, in addition to those already mentioned in this thread:

- [ ] `FileBlame`, the data structure that holds blame data for a file, stores `BlameHunk` for each line. This is less memory-efficient, but easier to use. I’m of course happy to change that if a different trade-off is desired.
- [ ] `get_rows`, previously `get_lines`, runs on every call to `draw`. If we stored the intermediate result somewhere, this would not be necessary (I ran into lifetime issues when I first tried).
- [ ] the title does not contain the commit id yet.

gitui-git-blame-second-draft

src/components/blame_file.rs
```diff
@@ -0,0 +1,310 @@
...
+        };
+
+        let mut cells = if show_metadata {
+            self.get_metadata_for_line_blame(width, hunk_for_line)
+        } else {
+            vec![Cell::from(""), Cell::from(""), Cell::from("")]
+        };
+
+        cells.push(
+            Cell::from(format!("{:>4}{}", line_number, VERTICAL_BAR))
```
extrawurst, line 245: how does this work? is it aligning the number to a width of 4 digits? how does this behave in files with 5 digit line numbers or more?

src/components/blame_file.rs
```diff
@@ -0,0 +1,310 @@
...
+            || "0000000".into(),
+            |hunk| hunk.commit_id.get_short_string(),
+        );
+        let author_width =
+            (width.saturating_sub(19) / 3).max(3).min(20);
+        let truncated_author = blame_hunk.map_or_else(
+            || "<no author>",
+            |hunk| hunk.author.unicode_truncate(author_width).0,
+        );
+        let author = format!("{:20}", truncated_author);
```
extrawurst, line 271: in the commit log we do not simply truncate the author, we add `..` - we should make this a reusable function and do the same here

src/components/blame_file.rs
```diff
@@ -0,0 +1,310 @@
...
+            .as_ref()
+            .map_or(0, |file_blame| file_blame.lines.len() - 1);
+
+        let new_selection = match scroll_type {
+            ScrollType::Up => old_selection.saturating_sub(1),
+            ScrollType::Down => {
+                old_selection.saturating_add(1).min(max_selection)
+            }
+            ScrollType::Home => 0,
+            ScrollType::End => max_selection,
```
extrawurst, line 300: we should also support pageup/down

src/app.rs
```diff
@@ -560,6 +536,10 @@ impl App {
             }
             InternalEvent::TabSwitch => self.set_tab(0)?,
             InternalEvent::InspectCommit(id, tags) => {
+                // Having to explicitly close another popup before showing a new
+                // one is not an optimal solution because this constraint can’t
+                // be automatically enforced.
+                self.blame_file_popup.hide();
```
extrawurst, line 542: so far we were able to solve this by stacking popups on top of each other and rely on a well defined order. in this case we might still get by with the same. `blame-popup`>`commit-popup` and if we should add going to a `commit-popup` based on a hunk in the the `blame-popup` we close the blame. what do you think?

long term we might need a breadcrum system so we can implement nice `go-back` functionality but that is a battle for another day

asyncgit/src/sync/blame.rs
```diff
@@ -0,0 +1,89 @@
...
+        .lines()
+        .enumerate()
+        .map(|(i, line)| {
+            // Line indices in a `FileBlame` are 1-based.
+            let corresponding_hunk = blame.get_line(i + 1);
+
+            if let Some(hunk) = corresponding_hunk {
+                let commit_id = CommitId::new(hunk.final_commit_id());
+                // Line indices in a `BlameHunk` are 1-based.
+                let start_line = hunk.final_start_line() - 1;
```
extrawurst, line 58: weird that it does not complain, this could technically wrap around..

asyncgit/src/sync/blame.rs
```diff
@@ -0,0 +1,89 @@
...
+    ///
+    pub lines: Vec<(Option<BlameHunk>, String)>,
+}
+
+///
+pub fn blame_file(file_path: &str) -> Result<FileBlame> {
+    let repo_path = CWD;
+    let repo = utils::repo(repo_path)?;
+
+    // https://github.com/rust-lang/git2-rs/blob/7912c90991444abb00f9d0476939d48bc368516b/examples/blame.rs
```
extrawurst, line 40: what does that mean?

asyncgit/src/sync/blame.rs
```diff
@@ -0,0 +1,89 @@
...
+    let commit_id = "HEAD";
+    let spec = format!("{}:{}", commit_id, file_path);
+    let blame = repo.blame_file(Path::new(file_path), None)?;
+    let object = repo.revparse_single(&spec[..])?;
```
extrawurst, line 44: isn't there `as_slice` for this?

asyncgit/src/sync/blame.rs
```diff
@@ -0,0 +1,89 @@
...
+pub struct FileBlame {
+    ///
+    pub path: String,
+    ///
+    pub lines: Vec<(Option<BlameHunk>, String)>,
+}
+
+///
+pub fn blame_file(file_path: &str) -> Result<FileBlame> {
+    let repo_path = CWD;
```
extrawurst, line 37: you want this to be a param, to be able to unittest this

extrawurst (changes requested):
awesome work! we are close

I would actually suggest we get a first iteration of this in and move some of the todos into a followup-pr so that more people can start playing around with this.

I'd suggest we split away the async-part for now.
this means we can focus on testing and usability:

* regarding the `FileBlame` structure I'd say lets keep it like this for now and worry about it further if/once we need to.
* lets run `get_rows` on every draw for now and optimise later (after doing an actual profile session on a big ass file blame)
* no panic on untracked files
* add all the commands where you can trigger a `blame` (maybe rather use `[b]` because `[enter]` is used already in a couple of places)
* fix the popup stacking
* add unittests
* some smaller comments in the inline comments

src/components/blame_file.rs
cruessler: Yes, this is aligning numbers to 4 digits width, padding left with spaces. I changed the line to take the maximum number of lines into account, so 4 is not hard-coded anymore.

src/components/blame_file.rs
cruessler: Thanks for the hint. Changed.

src/components/blame_file.rs
cruessler: Changed.

src/app.rs
cruessler: Thanks for the info. I was able to delete the extra call and rely on the order `draw` is called in. As far as I can see, there is no need to call `hide` anymore.

asyncgit/src/sync/blame.rs
cruessler: The following lines were taken from one of `git2-rs`’s examples and I wanted to attribute the code. Seeing that that was rather confusing than helpful, I removed the comment. :-)

asyncgit/src/sync/blame.rs
cruessler: Thanks for the hint! `&spec` also works (the code came from one of `git2-rs`’s examples).

asyncgit/src/sync/blame.rs
cruessler: Thanks for the hint! I added `repo_path` (and `commit_id`) as parameters to `blame_file`.

cruessler:
I have updated the code. As far as I can see, there are no major issues left! There was only one issue: `[b]` is already taken in some views (it opens `branches`), so `branches` is currently shadowed if `blame` can be used on an item that is selected. Does that work or do you want to use a different shortcut?

cruessler:
’blame_file’ currently calls ’get_commit_info’ for every line. This can of course be optimized.

src/components/blame_file.rs
```diff
@@ -0,0 +1,372 @@
...
+    current_height: std::cell::Cell<usize>,
+}
+
+static COMMIT_ID: &str = "HEAD";
+static NO_COMMIT_ID: &str = "0000000";
+static NO_AUTHOR: &str = "<no author>";
+static AUTHOR_WIDTH: usize = 20;
+
+fn get_author_width(width: usize) -> usize {
+    (width.saturating_sub(19) / 3).max(3).min(AUTHOR_WIDTH)
```
extrawurst, line 43: there is actually `clamp` for this since 1.50: https://doc.rust-lang.org/core/cmp/trait.Ord.html#method.clamp

src/components/blame_file.rs
```diff
@@ -0,0 +1,372 @@
...
+    file_blame: Option<FileBlame>,
+    table_state: std::cell::Cell<TableState>,
+    key_config: SharedKeyConfig,
+    current_height: std::cell::Cell<usize>,
+}
+
+static COMMIT_ID: &str = "HEAD";
+static NO_COMMIT_ID: &str = "0000000";
+static NO_AUTHOR: &str = "<no author>";
+static AUTHOR_WIDTH: usize = 20;
```
extrawurst, line 40: this is rather `AUTHOR_MIN_WIDTH`, right?

src/components/blame_file.rs
```diff
@@ -0,0 +1,372 @@
...
+
+fn get_author_width(width: usize) -> usize {
+    (width.saturating_sub(19) / 3).max(3).min(AUTHOR_WIDTH)
+}
+
+const fn number_of_digits(number: usize) -> usize {
```
extrawurst, line 46: funny how this is not in `std` and not even a crate for it :)

src/components/blame_file.rs
```diff
@@ -0,0 +1,372 @@
...
+
+        let mut cells = if show_metadata {
+            self.get_metadata_for_line_blame(width, hunk_for_line)
+        } else {
+            vec![Cell::from(""), Cell::from(""), Cell::from("")]
+        };
+
+        let line_number_width = self.get_line_number_width();
+        cells.push(
+            // U+2502 is BOX DRAWINGS LIGHT VERTICAL.
```
extrawurst, line 282: TUI has some constants for these things actually

extrawurst (approved):
just a few remaining comments, but no deal breaker so I will merge this

extrawurst:
@cruessler really great work! Lets do the rest in smaller followups👍 maybe you can start by writing an issue that collects what should still be done, like *making it async*, optimise the *get_commit_info*..

cruessler:
@extrawurst I imagine there might be follow-ups that could be a `good first issue`. If you agree, I would open those in separate issues and provide a few instructions on how an implementation might look like. Otherwise, I would be perfectly happy to do the changes myself. :-)

src/components/blame_file.rs
cruessler: Thanks for the hint! Changed.

src/components/blame_file.rs
cruessler: I changed it to `MAX_AUTHOR_WIDTH` and introduced an additional `MIN_AUTHOR_WIDTH`.

src/components/blame_file.rs
cruessler: I was surprised by that, too. :-)

src/components/blame_file.rs
cruessler: Thanks for the hint, changed.

extrawurst:
@cruessler tbh I would like you to take care of the #656 first (all small stuff) and then the async-part separately (both of which is essential before we can release a version with blaming, so that we keep true to our high standards of usability and performance)

cruessler:
@extrawurst No worries, I didn’t know you wanted to prioritize a quick release. I’ll work on the remaining issues this weekend (so you know when to expect the first PR).

PR #639, opened 2021-04-13
Updates vim_style_key_config.ron path

irmantas:
With commit #b0b37a8 vim style config path in KEY_CONFIG.md was unchanged


extrawurst:
cool thanks!

PR #640, opened 2021-04-13
Fixes #632

### LINKED ISSUE #632: checkout already checked out remote branch
right now we get an error when we try to checkout a remote branch *again* - it would be much nicer to go the list of local branches again and mark the local branch that tracks the remote branch.

we want a method that checks if there is a local branch that tracks the remote branch we currently have selected.
if that is the case we want to switch back to local branch view and select that branch instead of doing a new checkout. (if we find multiple such branches, we pick the first).

Note: simply checking for a name match does not cut it, since technically a branch could have any name and still be set to track a specific remote branch

shahsurajk:
## Fix: 
* Checks for local branch when checking a remote branch out

**Note**:
I'm a first time contributor to this project. Let me know if i've not done something as per the contribution guidelines

### DIFF
--- a/asyncgit/src/sync/branch/mod.rs
+++ b/asyncgit/src/sync/branch/mod.rs
@@ -252,7 +252,7 @@ pub fn checkout_branch(
         if let Err(e) = repo.checkout_head(Some(
             git2::build::CheckoutBuilder::new().force(),
         )) {
-            // This is safe beacuse cur_ref was just found
+            // This is safe because cur_ref was just found
             repo.set_head(
                 bytes2string(cur_ref.name_bytes())?.as_str(),
             )?;
@@ -264,7 +264,7 @@ pub fn checkout_branch(
     }
 }
 
-///
+/// Modify HEAD to point to a branch then checkout head, does not work if there are uncommitted changes
 pub fn checkout_remote_branch(
     repo_path: &str,
     branch: &BranchInfo,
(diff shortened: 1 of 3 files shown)


shahsurajk:
is there a way to check these error locally?

brunogouveia:
Did you running `cargo clippy`?

extrawurst:
The CI pipeline is defined in the repo itself. we use GitHub actions. you can run the same as the CI locally: 
```
make check
```

shahsurajk:
> Did you running `cargo clippy`?

Clippy threw this error: 
```
error reading Clippy's configuration file `/home/dvfd/IdeaProjects/gitui/.clippy.toml`: unknown field `msrv`, expected one of `blacklisted-names`, `cognitive-complexity-threshold`, `cyclomatic-complexity-threshold`, `doc-valid-idents`, `too-many-arguments-threshold`, `type-complexity-threshold`, `single-char-binding-names-threshold`, `too-large-for-stack`, `enum-variant-name-threshold`, `enum-variant-size-threshold`, `verbose-bit-mask-threshold`, `literal-representation-threshold`, `trivial-copy-size-limit`, `too-many-lines-threshold`, `array-size-threshold`, `vec-box-size-threshold`, `max-trait-bounds`, `max-struct-bools`, `max-fn-params-bools`, `warn-on-all-wildcard-imports`, `third-party` at line 1 column 1

```

shahsurajk:
> > Did you running `cargo clippy`?
> 
> Clippy threw this error:
> 
> ```
> error reading Clippy's configuration file `/home/dvfd/IdeaProjects/gitui/.clippy.toml`: unknown field `msrv`, expected one of `blacklisted-names`, `cognitive-complexity-threshold`, `cyclomatic-complexity-threshold`, `doc-valid-idents`, `too-many-arguments-threshold`, `type-complexity-threshold`, `single-char-binding-names-threshold`, `too-large-for-stack`, `enum-variant-name-threshold`, `enum-variant-size-threshold`, `verbose-bit-mask-threshold`, `literal-representation-threshold`, `trivial-copy-size-limit`, `too-many-lines-threshold`, `array-size-threshold`, `vec-box-size-threshold`, `max-trait-bounds`, `max-struct-bools`, `max-fn-params-bools`, `warn-on-all-wildcard-imports`, `third-party` at line 1 column 1
> ```

Figured out the problem, it was my rust version.

extrawurst:
Hi @shahsurajk thanks for looking into this but I think we want another solution: 
we want a method that checks if there is a local branch that tracks the remote branch we currently have selected. 
if that is the case we want to switch back to local branch view and select that branch instead of doing a new checkout. (if we find multiple such branches, we pick the first)

shahsurajk:
Umm, @extrawurst so this change actually does the exact same thing.🤔

If you have a local branch tracked a remote and if you try to checkout the same remote once again, the change pushes you back to the local branch view. It also checks if the local branch selected is actually the one tracked by the remote specified. 

What is the other solution you were proposing? I might've misunderstood it altogether. 

Also, do you mind checking the branch out, cause when I was testing it, it actually worked like what you mentioned. 🤔

If not, I'll happily close this PR. Sorry for the friction.

asyncgit/src/sync/branch/mod.rs
```diff
@@ -290,18 +290,52 @@ pub fn checkout_remote_branch(
     };
 
     let commit = repo.find_commit(branch.top_commit.into())?;
-    let mut new_branch = repo.branch(&name, &commit, false)?;
-    new_branch.set_upstream(Some(&branch.name))?;
+
+    // find existing local branch for the remote branch
+    let branch_to_checkout = if let Ok(matched_local_branch) =
+        repo.find_branch(&name, BranchType::Local)
```
extrawurst, line 296: this only checks if there is a local branch with the same name.
a) if this fails it does not mean there is a local branch with another name that track this remote branch
b) if this finds a branch with the same name it does not necessarily mean that this tracks this remote branch

asyncgit/src/sync/branch/mod.rs
```diff
@@ -264,7 +264,7 @@ pub fn checkout_branch(
     }
 }
 
-///
+/// Modify HEAD to point to a branch then checkout head, does not work if there are uncommitted changes
```
extrawurst, line 267: checking for this inside of `checkout_remote_branch` is too late. we want a separate function that checks if we can find any local branch (we need to iterate all) that is a tracking branch of a specific remote branch

extrawurst (changes requested):
I left more inline explanation - feel free to pursue this further if you like

shahsurajk:
closing in favor or a re-implementation.

PR #660, opened 2021-04-24
Improve blame view

### LINKED ISSUE #656: improve blame
@cruessler thanks for implementing blaming! now that https://github.com/extrawurst/gitui/issues/484 landed there are a few things that I noticed:

- [ ] blame key should be capital `B` to be unique across the entire app and different to `branches`
- [ ] scrollbar (we have some helpers to draw them already)
- [ ] the current commit hash of the file being blamed (in the title) says `HEAD` although it would be nice to actually show the hash
- [ ] the hashes in the view that match the last commit hash should be bold to identify the last changes (just an idea, let me know what you think)
- [ ] allow viewing a specific commit that I am currently selecting in the diff to inspect (we have all that, you just need to send the right command and close the blame)
- [ ] we probably should not allow blaming a binary file (I am surprised it does not crash on it):

Screenshot 2021-04-21 at 16 21 01

cruessler:
- Set default shortcut to `B` instead of `b` because the latter would
  shadow `[b]ranches`.
- Add scrollbar.
- Show resolved commit id in title instead of `HEAD`.
- Make commit id bold if it is the commit id the file is blamed at.
- Don’t run blame on a binary file.
- Add shortcut for inspecting a commit in blame view.

This closes #656.

### DIFF
--- a/asyncgit/src/sync/blame.rs
+++ b/asyncgit/src/sync/blame.rs
@@ -1,7 +1,10 @@
 //! Sync git API for fetching a file blame
 
 use super::{utils, CommitId};
-use crate::{error::Result, sync::get_commit_info};
+use crate::{
+    error::{Error, Result},
+    sync::get_commit_info,
+};
 use std::io::{BufRead, BufReader};
 use std::path::Path;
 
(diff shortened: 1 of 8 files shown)


asyncgit/src/sync/blame.rs
```diff
@@ -22,28 +25,47 @@ pub struct BlameHunk {
...
+        BlameAt::Commit(commit_id) => *commit_id,
+    };
 
-    let spec = format!("{}:{}", commit_id, file_path);
+    let spec = format!("{}:{}", commit_id.to_string(), file_path);
     let blame = repo.blame_file(Path::new(file_path), None)?;
     let object = repo.revparse_single(&spec)?;
     let blob = repo.find_blob(object.id())?;
+
+    if blob.is_binary() {
```
extrawurst, line 65: lets do this **before** the `blame_file` call to not spent long time on calculating the blame and then *early* out anyway ;)

extrawurst:
Really cool! just one small thing but I can fix this after the merge! Great work! now last thing: https://github.com/extrawurst/gitui/issues/661

PR #664, opened 2021-04-24
Call get_commits_info instead of get_commit_info

cruessler:
This results in far fewer calls to `repo.find_commit` in almost all
cases.

This is a follow-up to #634.


extrawurst:
Smart👍

PR #667, opened 2021-04-25
Add async wrapper to blame

### LINKED ISSUE #661: async blame
thanks to @cruessler https://github.com/extrawurst/gitui/pull/660 is fixed now and the last bit before being able to release file blaming is: make it non-blocking.

right now calculating a file blame blocks the entire app. we have mechanisms for a few other async git operations, we should apply the same conecept here too.

cruessler:
- Rename `self.path` to `self.file_path` in `BlameFileComponent`.
- Take into account that `draw_scrollbar` subtracts the area’s height
  before calculating the scrollbar’s position.

The implementation closely mirrors the one in `asyncgit/src/blame.rs`. Let me
know if I understood the async part correctly.

I tested the implementation locally in my `gitui` repository, and I did not
encounter any issues, but I also did not spend a large amount of time on
testing. If you think the implementation looks good, I will test it on a few
more repositories.

This closes #661.

### DIFF
--- a/src/components/blame_file.rs
+++ b/src/components/blame_file.rs
@@ -11,9 +11,10 @@ use crate::{
 };
 use anyhow::Result;
 use asyncgit::{
-    sync::{blame_file, BlameAt, BlameHunk, CommitId, FileBlame},
-    CWD,
+    sync::{BlameHunk, CommitId, FileBlame},
+    AsyncBlame, AsyncNotification, BlameParams,
 };
+use crossbeam_channel::Sender;
 use crossterm::event::Event;
 use std::convert::TryInto;
 use tui::{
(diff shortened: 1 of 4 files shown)


src/components/blame_file.rs
```diff
@@ -67,7 +69,7 @@ impl DrawableComponent for BlameFileComponent {
     ) -> Result<()> {
         if self.is_visible() {
             let path: &str = self
-                .path
+                .file_path
                 .as_deref()
                 .unwrap_or("<no path for blame available>");
```
extrawurst, line 75: it would be great to draw a different text while the blame is pending

extrawurst (commented):
wow you are a machine! we can release this soon❤️

> The implementation closely mirrors the one in `asyncgit/src/blame.rs`.

that's the new one, which one did you mirror?

anyhow it looks good, just one small improvement comment added.

cruessler:
My mistake, I meant `asyncgit/src/diff.rs`. :smiley:

src/components/blame_file.rs
cruessler: Thanks for the idea! I extracted the code to `self.get_title` and now it includes the information whether a blame is pending.

extrawurst:
Nice work!👍

PR #668, opened 2021-04-25
Use git2::DiffLineType enum instead of char values (#655)

### LINKED ISSUE #655: cleanup some stringly typed code
now that https://github.com/rust-lang/git2-rs/commit/ae256db3b27dd7dedab02fa5c051bd7adedf7de7 was released we can switch our usage of the diff line types to enum values:
used for diff and line-ops OTOH

wandernauta:
The DiffLineType enum in asyncgit has fewer variants: it maps context lines and
line types we don't expect to None. This behavior is kept unchanged.

Now calls `.origin_value()` rather than `.origin()` except in one debug log.

closes #655

### DIFF
--- a/asyncgit/src/sync/diff.rs
+++ b/asyncgit/src/sync/diff.rs
@@ -27,6 +27,19 @@ pub enum DiffLineType {
     Delete,
 }
 
+impl From<git2::DiffLineType> for DiffLineType {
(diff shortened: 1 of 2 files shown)


extrawurst:
@wandernauta welcome and thank you for this contribution ❤️

PR #671, opened 2021-04-25
Load a commit.template message if one is configured (#546)

### LINKED ISSUE #546: support commit.template
based off of https://github.com/extrawurst/gitui/issues/517

git commit uses a default template for commit messages that the `commit.template` config allows the user to overwrite:
https://www.git-scm.com/book/en/v2/Customizing-Git-Git-Configuration

it should be fairly easy to use that config in gitui and pre-populate a commit msg text

wandernauta:
If a commit message template is set with...

    git config commit.template /path/to/some/file.txt

...this fills the commit message box (and therefore, later, the external
editor) with the contents of that file, as long as you don't have something
more interesting typed there already.

Like the `git commit` command, this doesn't let you create a commit that has
just the template as its message, unchanged.  Unlike the `git commit` command,
not being able to read the template file is not fatal: specifically, if the
commit.template setting is pointed at a file that does not exist, this does
nothing.

Resolves #546.

### DIFF
--- a/src/components/commit.rs
+++ b/src/components/commit.rs
@@ -13,12 +13,12 @@ use crate::{
 use anyhow::Result;
 use asyncgit::{
     cached,
-    sync::{self, CommitId, HookResult},
+    sync::{self, utils::get_config_string, CommitId, HookResult},
     CWD,
 };
 use crossterm::event::Event;
 use std::{
-    fs::File,
+    fs::{read_to_string, File},
     io::{Read, Write},
     path::PathBuf,
 };
@@ -35,6 +35,7 @@ pub struct CommitComponent {
     queue: Queue,
     key_config: SharedKeyConfig,
     git_branch_name: cached::BranchName,
+    commit_template: Option<String>,
 }
 
 impl DrawableComponent for CommitComponent {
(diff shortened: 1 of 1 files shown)


src/components/commit.rs
```diff
@@ -154,12 +162,22 @@ impl CommitComponent {
...
     ///
     pub fn update(&mut self) -> Result<()> {
         self.git_branch_name.lookup().map(Some).unwrap_or(None);
+
+        self.commit_template.get_or_insert_with(|| {
+            get_config_string(CWD, "commit.template")
+                .ok()
+                .unwrap_or(None)
+                .and_then(|path| read_to_string(path).ok())
+                .unwrap_or_else(String::new)
```
extrawurst, line 178: we should make the `commit_template` be `None` if we cannot read the file

src/components/commit.rs
```diff
@@ -154,12 +162,22 @@ impl CommitComponent {
...
             git_branch_name: cached::BranchName::new(CWD),
+            commit_template: None,
         }
     }
 
     ///
     pub fn update(&mut self) -> Result<()> {
         self.git_branch_name.lookup().map(Some).unwrap_or(None);
+
+        self.commit_template.get_or_insert_with(|| {
```
extrawurst, line 173: `update` will be called quit often at runtime. how about wrapping all of it in a `if is_visible() {}`?

extrawurst:
fixed in 3ad1911

PR #688, opened 2021-05-07
use CredComponent for PullComponent via RefCell shared with main app

sum12:
this allow the CredComponent to be reused across the all other
components.
Since at any time only one using compoenent (here PullComponent only for
now) is visible only one these sharing entities will try to get hold of
the shared component via the RefCell

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -132,6 +132,7 @@ impl App {
                 sender,
                 theme.clone(),
                 key_config.clone(),
+                cred_popup.clone(),
             ),
             pull_popup: PullComponent::new(
                 &queue,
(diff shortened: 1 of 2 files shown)


sum12:
@extrawurst I know this is probably annoying. But can you give this another approval ?
I have done a `make check` `make test` and `make build-release` so atleast the linux part should pass the CI

(Thinking of another smaller PR just become a contributor before trying to work further on this on)

sum12:
@extrawurst any input on this ?

sum12:
[try 2] @extrawurst does this go in the right direction ?

extrawurst:
> [try 2] @extrawurst does this go in the right direction ?

sorry for not giving you feedback on this yet, I hope to be able to on the weekend

sum12:
Is this still relevant ( even after resolving the conflicts) ? 
@extrawurst

Issue #81, opened 2020-05-20
Show the branching structure in the log.


jugglerchris:
**Is your feature request related to a problem? Please describe.**
It can be useful seeing the branching/merging structure in the log view, in a similar way to `git log --oneline --decorate --graph`

**Describe the solution you'd like**
The graph view visible in the log window.

**Describe alternatives you've considered**
I haven't got any other alternatives other than not doing it.  :-)

extrawurst:
Definitely on my list aswell 👍

bew:
nop little bot, don't close that please! :smiley:

extrawurst:
@bew volunteering to take a go at it? :)

bew:
That would have been a good idea, thanks for the offer, but not right now anyhow.
I'm in the _long_ process of re doing all my configs properly, and I still far from having a proper rust editor setup (I'd also need to learn writing rust, not just read, this project could help bootstrap the process for sure, but I'm not ready for now)

kiuKisas:
up ?

kting28:
Still a planned feature?

extrawurst:
A few ideas how this could be visualized: https://stackoverflow.com/questions/1057564/pretty-git-branch-graphs

MaiLunJiye:
suggestion: I hope we can merge, checkout and reset in the `graph view` like the vscode extension name `git graph`(https://marketplace.visualstudio.com/items?itemName=mhutchie.git-graph).

joseffilzmaier:
It would already be nice to just show the remote branches and local branches in the log viewer

extrawurst:
@joseffilzmaier not sure what you mean. There is little chance to have multiple references be on the same branch (we only show a single branch currently)

P4Cu:
If `gitui` would ever draw more branches here's how libgit2 authors imagine it: https://github.com/libgit2/libgit2/blob/main/examples/log.c
Though they've not added the part that's related to `--graph` but it can be deduced out of parents.

extrawurst:
I put a bunch thought into this already. The biggest issue is keeping the revolt tab as efficient as it is right now. I don’t want to start building a graph in memory to traverse. It’s a matter of finding an efficient way to store everything needed to draw the graph structure and jump around everywhere.

another factor is being able to draw iteratively while not everything is parsed yet (like now)

the example in libgit2 unfortunately is too simple to cover any of these questions.

P4Cu:
[...]

PR #705, opened 2021-05-13
Add popup for tags

### LINKED ISSUE #483: more tag support
right, now I can only create a tag.

lets add a popup or tab (not sure yet) that allows:

* list all tags (mark the ones not pushed to remote)
* delete a tag
* fetch tags

cruessler:
This closes #483.

This is a preview that has all the foundations in place so that we can now quickly iterate on the features that should be part of the first release.

@extrawurst You mentioned the ability to jump to a specific commit in the log view. Did you have something like `InternalEvent::JumpToCommitInRevlog(CommitId)` in mind?

The selection logic, in particular `TagListComponent::move_selection`, is quite similar to the one in `BlameFileComponent`, in particular `BlameFileComponent::move_selection`, so there is some potential for code re-use, but I did not yet explore that in-depth.

Let me know what you think!


cruessler:
gitui-tags-popup-first-draft

extrawurst:
Really cool start! Thanks @cruessler 
 
> This is a preview that has all the foundations in place so that we can now quickly iterate on the features that should be part of the first release.

I actually think the toughest part is yet to come 🙈 I think we will want more than the current bare *tag-name-list* - the very least we should show with each entry: *timestamp*, *creator* (and maybe a *short-desc*)

This makes the whole process (if not already) require to be able to run async (since we need to pull commit infos on each `CommitId`). Albeit I am cool to make it as a `sync` function first (MVP style) and make the `async`ification a followup.

> @extrawurst You mentioned the ability to jump to a specific commit in the log view. Did you have something like `InternalEvent::JumpToCommitInRevlog(CommitId)` in mind?

Exactly.
Keep in mind though that tags we find in the repo are not necessarily on our current branch. We need to take this into account.

> The selection logic, in particular `TagListComponent::move_selection`, is quite similar to the one in `BlameFileComponent`, in particular `BlameFileComponent::move_selection`, so there is some potential for code re-use, but I did not yet explore that in-depth.

That would actually be cool, there are probably even more places we have such code!

On top of what is mentioned above what is still missing:
* [ ] fetch tags (is already part of the *push tags* functionality)
* [ ] deleta a tag
* [ ] sort the tag list based on the *timestamp*

cruessler:
Thanks for the in-depth feedback! I’ll update the PR over the next few days.

cruessler:
@extrawurst Do you want the popup to have some sort of indicator whether a tag is annotated, e. g. bold font or a symbol?

extrawurst:
> @extrawurst Do you want the popup to have some sort of indicator whether a tag is annotated, e. g. bold font or a symbol?

I don’t see why right now. Any reason to do so? I doubt most people actually know there is a dofference 🤣

cruessler:
Fine with me, that will likely also be easier to implement. 😀 Just wanted to make sure so that I can adapt the data structures to all use cases from the start.

cruessler:
@extrawurst Does the following layout work for you? As soon as the layout is done, I’ll refactor a bit and implement the rest.

gitui-tags-popup-second-draft

extrawurst:
@cruessler that looks great! lets sort the tags by the date of the tag though :)

extrawurst:
@cruessler really cool!

so I just double checked and we even have a unittest `test_push_pull_tags` that shows that regular fetching already fetches tags aswell. so for me this is not strictly an MVP feature to have a dedicated tag-fetching. is there anything you still wanna do, otherwise please update to current master (that should fix the nightly ci steps aswell) and mark this as ready for review ❤️

cruessler:
@extrawurst I think that’s a good idea, I’ll update the PR soon!

cruessler:
@extrawurst I added a command “Select commit” and tested it in the `gitui` as well as in the `linux` repository. As far as I could see, jumping to a commit worked without issues.

There is one minor issue I’m not sure how to handle best. If the commit is not part of the current log because it is on a different branch or has not been loaded yet, the popup closes, but nothing else happens. That might be confusing. We could pass `AsyncLog` to `TagListComponent` to make sure the target can actually be shown before sending a command. What do you think?

src/tabs/revlog.rs
```diff
@@ -166,6 +166,12 @@ impl Revlog {
             tags.and_then(|tags| tags.get(&commit).cloned())
         })
     }
+
+    pub fn select_commit(&mut self, id: CommitId) {
```
extrawurst, line 170: we can make this return `Result` and then feedback whether we were able to find the commit in question. then we can open a msg popup (we use that for other *errors* in other places) to tell this to the user and then we can keep the tag popup actually open, right?

extrawurst: feel free to mention that this is a planned feature anyway (see #81)

cruessler: That sounds like a good idea. I changed the code as suggested. I was not sure about the wording for #81, though. If you have a suggestion, I will incorporate it, otherwise you could also merge and then change the message. Whatever works best. :smile:

extrawurst:
@cruessler cool stuff!

extrawurst:
the only thing that seems odd is the pageup/down behaviour, it seems one-off:

Screenflick Movie 19

cruessler:
You’re right! I changed that.

The behaviour came from `BlameFileComponent::move_selection` where it is harder to see that something is off. One might even argue that it is more correct to jump to the first invisible line because that is the first line of the next page. On the other hand, I think it makes sense to go to the last visible line because then the movement of the selection is more obvious.

extrawurst:
ok I am going to merge this now 🥳

Issue #725, opened 2021-05-21
integrate syntax highlighter in file-tree-view


extrawurst:
based off of https://github.com/extrawurst/gitui/issues/714

extrawurst:
map is done:

![Screenflick Movie 15](https://user-images.githubusercontent.com/776816/119244434-fcfb7000-bb70-11eb-8bc8-6259aea4264d.gif)

left to do:
* [ ] scrollbar
* [ ] a bunch of //todos 
* [ ] make git file fetching async aswell

extrawurst:
next update:

![Screenflick Movie 16](https://user-images.githubusercontent.com/776816/119277582-d39a1d00-bc20-11eb-9f6d-b8ca285d6b6c.gif)

PR #727, opened 2021-05-22
add syntax highlighting

extrawurst:
see #725 
* [x] cleanup `SyntaxText` type
* [x] put highlighting into thread


extrawurst:
unfortunately using syntact for highlighting brings some serious bloat along:

binary size:
6.4 -> 8.8 
4.4 -> 5.6 (stripped)

compilation unites:
135 -> 172

PR #735, opened 2021-05-24
Use default key config and theme when values are missing

hpcsc:
## Background
- I have gitui key and theme config files versioned in my dotfiles repository. Whenever I update gitui, there are a few new keys that are supported by gitui and not in my config files. gitui refuses to load my config files and creates new default ones instead. This is very inconvenient since I need to compare my config files and default config files to see what values are missing to update

## What is in this PR?
- Use `serde(default = "path")` for fields in key config and theme to specify default values if they are missing from config files during deserialization. This does create a bunch of private functions (`default_xxx()`) but this is the only way that I can see from serde documentation

### DIFF
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -418,7 +418,6 @@ mod tests {
         let expected_tab_status = KeyEvent { code: KeyCode::Char('6'), modifiers: KeyModifiers::SHIFT };
         assert_eq!(config.tab_status, expected_tab_status);
 
-        let expected_default_tab_log = KeyEvent { code: KeyCode::Char('2'), modifiers: KeyModifiers::empty() };
-        assert_eq!(config.tab_log, expected_default_tab_log);
+        assert_eq!(config.tab_log, KeyConfig::default_tab_log());
     }
 }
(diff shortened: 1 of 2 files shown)


src/keys.rs
```diff
@@ -1,144 +1,259 @@
...
-};
-use serde::{Deserialize, Serialize};
 use std::{
     fs::{self, File},
     io::{Read, Write},
     path::PathBuf,
     rc::Rc,
 };
 
+use anyhow::Result;
```
extrawurst, line 11: what is this reordering good for?

src/ui/style.rs
```diff
@@ -1,57 +1,60 @@
...
-};
-use serde::{Deserialize, Serialize};
 use std::{
     fs::{self, File},
     io::{Read, Write},
     path::PathBuf,
     rc::Rc,
 };
+
+use anyhow::Result;
```
extrawurst, line 11: needless reorders again, this confuses the review

extrawurst (commented):
ok I checked this out and I like that we try to fix this issue of breaking the configs every time we add stuff. but the files get bloated a lot using this approach where we need a separate method for each field. I wonder if we can find a more elegant solution using custom serialise: https://serde.rs/deserialize-struct.html

in a perfect world we can detect a missing field and only pick that off the `Default::default` and keep this as is

extrawurst:
Superseded by optional overwrites in key config

PR #736, opened 2021-05-24
Added Full path to key_config.ron in windows

imranZERO:


extrawurst:
correct way to write this is using `%APPDATA%`, I added this: 693bc25

PR #750, opened 2021-05-29
allow components to consume events before quit

franciscod:
this allows using the same bind for both exit_popup and exit


extrawurst:
closing after we figured out chars like `q` cannot be used for closing text input anyway.

PR #753, opened 2021-05-31
honor showUntrackedFiles config

### LINKED ISSUE #752: Support using gitui with yadm
This is half a feature request and half a bug. 
I use YADM for managing my dotfiles, which essentially creates a bare git repo in the $HOME directory.
YADM adds a configuration option so when I do `yadm status` it doesnt try show all the untracked files.

https://git-scm.com/docs/git-status#Documentation/git-status.txt---untracked-filesltmodegt

On the other hand when I try to use gitui (using `gitui -d ~/.config/yadm/repo.git`), 
it gets stuck 'forever' probably trying to find/enumerate all the untracked files.
(Sublime Merge also seems to have this problem, but for example verco doesn't)

I suspect similar problems would happen in large repositories with large amounts of changed/untracked files

I see a few ways this could be fixed

1. check for the status.showUntrackedFiles config option and dont spend time trying to list untracked files/directories
2. add a `no-untracked` flag to do something similar to above
3. detect when there are alot of untracked files and stop trying to find more than is visible in the window
   - find the rest on-demand
4. somehow make gitui still/more responsive while finding/rendering the untracked files anyway
5. ETA: there is also the `gui.displayuntracked` which is more directly related to GUIs and untracked files, it would be a good idea to look at this option too

Unfortunately gitui is not really usable with YADM

extrawurst:
closes #752

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## Added
+- honor `config.showUntrackedFiles` improving speed with a lot of untracked items ([#752](https://github.com/extrawurst/gitui/issues/752))
+  
 ## Fixed
 - wrong file with same name shown in file tree ([#748](https://github.com/extrawurst/gitui/issues/748))
 
--- a/asyncgit/src/status.rs
+++ b/asyncgit/src/status.rs
@@ -31,19 +31,14 @@ pub struct Status {
 pub struct StatusParams {
     tick: u128,
     status_type: StatusType,
-    include_untracked: bool,
 }
 
 impl StatusParams {
     ///
-    pub fn new(
-        status_type: StatusType,
-        include_untracked: bool,
-    ) -> Self {
+    pub fn new(status_type: StatusType) -> Self {
         Self {
             tick: current_tick(),
             status_type,
-            include_untracked,
         }
     }
 }
--- a/asyncgit/src/status.rs
+++ b/asyncgit/src/status.rs
@@ -93,10 +93,9 @@ impl AsyncStatus {
         let hash_request = hash(&params);
 
         log::trace!(
-            "request: [hash: {}] (type: {:?}, untracked: {})",
+            "request: [hash: {}] (type: {:?})",
             hash_request,
             params.status_type,
-            params.include_untracked,
         );
 
         {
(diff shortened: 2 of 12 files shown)


IndianBoy42:
Should you check [gui.displayuntracked](https://git-cola.readthedocs.io/en/latest/git-cola.html#gui-displayuntracked)?

I can't seem to find if it's an official flag or only for git-cola. It is mentioned in the official git repo [here](https://github.com/git/git/blob/master/Documentation/config/gui.txt)

extrawurst:
> Should you check [gui.displayuntracked](https://git-cola.readthedocs.io/en/latest/git-cola.html#gui-displayuntracked)?
> 
> I can't seem to find if it's an official flag or only for git-cola. It is mentioned in the official git repo [here](https://github.com/git/git/blob/master/Documentation/config/gui.txt)

I went for the more standard config

IndianBoy42:
It partially works, it doesn't show the untracked files. However it still attempts to stage them when I run "Stage All"

I think either Stage-All should ignore untracked files (if `status.showUntrackedFiles == no`). Or we add a separate key for 'Update-All' which mimics `git add -u`

For example these commits in my fork:
1. https://github.com/IndianBoy42/gitui/commit/8ea8bfa49908c996dd520106e153e24ffdfda384
2. https://github.com/IndianBoy42/gitui/commit/4b2215abe9a01ab90a1a10d170ffab17d9b3b7ff

IndianBoy42:
I have manually 'rebased' my fork onto your new  way of getting the `showUntrackedFiles` option

https://github.com/IndianBoy42/gitui

If you are okay with it I can send a pull request after this PR is merged

extrawurst:
@IndianBoy42 I will fix the stage_all it should honour the same config. but an additional command for staging untracked files when you configured it to not show them does not carry its own weight IMO

extrawurst:
@IndianBoy42 `stage_all` honors the config too now

IndianBoy42:
This works well now.

Sorry to keep bothering but I remembered another thing that I added in my branch.

You may want the 'stash untracked' option to be unchecked by default if `showUntrackedFiles==no`. This is not a big deal because you can always unselect it before stashing, but its a little less error prone

extrawurst:
> This works well now.
> 
> Sorry to keep bothering but I remembered another thing that I added in my branch.
> 
> You may want the 'stash untracked' option to be unchecked by default if `showUntrackedFiles==no`. This is not a big deal because you can always unselect it before stashing, but its a little less error prone

should work now

PR #762, opened 2021-06-02
fix path handling on windows

### LINKED ISSUE #761: File tree pane: folders won't collapse on Windows
## Description

The new file tree pane [3] has some problematic behavior on Windows.

Using Left/Right Arrow [→←] to collapse/expand folders will correctly toggle the little triangle ▾ ▸ next to the folder (and jump to parent/child on second press). However the file list is not updated so all the subfolders are still visible in the tree after a collapse. 
All subfolders are visible when opening the file tree (but marked collapsed ▸) so the result is a list of all folders and files in the repository.

Note: the bug is **not** present when using Ubuntu via WSL.

## Reproduction

Launch gitui, open File Tree pane with [3].

**Expected behavior**
Only the first-level folders and files are visible

**Actual behavior**
All folders and files are visible

In the file tree, use arrow keys to collapse and expand some folders

**Expected behavior**
Files and folders appear and disappear when their parent is collapsed / expanded

**Actual behavior**
No change in the file list, excepted for the triangles ▾ ▸ 

## Context

- OS/Distro + Version: Windows 10, reproduced on both cmd.exe and Windows Terminal
- GitUI Version 0.16.0
- Rust version: 1.52.1

## Additional Notes

It's not a big issue (for me) since the file tree pane is not the main feature I use in gitui. However I understood that the file tree component is to be reused in other places, in particular in the status view. That would really get in my way (and most Windows users too).

extrawurst:
fixes #761

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   
 ## Fixed
 - wrong file with same name shown in file tree ([#748](https://github.com/extrawurst/gitui/issues/748))
+- filetree collapsing broken on windows ([#761](https://github.com/extrawurst/gitui/issues/761))
 
 ### Internal
 - use git_repository_message [[@kosayoda](https://github.com/kosayoda)] ([#751](https://github.com/extrawurst/gitui/issues/751))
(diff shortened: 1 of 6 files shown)


enizor:
works on windows \o/

Issue #742, opened 2021-05-27
fetch tags


extrawurst:
thanks to @cruessler and following #483 we now have two nice to have features left:

* [x] explicit fetch-tags command **repo wide** (right now I think we only fetch them for the current branch)
* [x] visualise tags in the list with an arrow icon if a tag is *pushable* (its not present on remote, we have that buried already in the *push-tags-feature*) (done in https://github.com/extrawurst/gitui/pull/776)

those can be tackled individually of course

extrawurst:
closed via #1013

PR #776, opened 2021-06-07
Show symbol if tag is not yet pushed

cruessler:
This partly addresses #742.

Let me know whether I put the async code in the right place or whether you’d like it moved to a different location!

gitui-tags-remote-indicator

### DIFF
--- a/asyncgit/src/lib.rs
+++ b/asyncgit/src/lib.rs
@@ -33,6 +33,8 @@ mod progress;
 mod push;
 mod push_tags;
 pub mod remote_progress;
+///
+pub mod remotes;
 mod revlog;
 mod status;
 pub mod sync;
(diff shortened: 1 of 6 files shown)


extrawurst:
Screenflick Movie 29

this does not seem to work quiet yet I am afraid. if I tag a commit and open the tags popup it does not show the new symbol

extrawurst:
> @extrawurst Is this possibly related to authentication being necessary? To test this hypothesis, I updated the PR to call `extract_username_password` and pass the result to `tags_missing_remote` (previously, `None` was passed instead). Can you try whether this has an effect on your machine?

nope that did not fix it. I suspect its related to the fact that you only ever update the `missing_remote_tags` once and then never again

cruessler:
You’re right, spawning the async job inside `if self.missing_remote_tags.is_none()` was not thought through. I removed the conditional and put the call to `spawn` inside `open`, so now the missing tags are refreshed every time the popup is opened.

extrawurst:
works great now! 👍

PR #789, opened 2021-06-15
Update vim_style_key_config.ron

ryuheechul:
exiting is one of the most frequent action of all, I think it make sense to exit easily with just `q`.

But it might be just me. In that case feel free to close this PR :)


extrawurst:
I am no vim guy, I can’t make that call. But it might be related to #750

extrawurst:
@ryuheechul not saying its not valid, but I need some other vim user to give feedback on this. why did you close it with no comment?

PR #792, opened 2021-06-20
Add filter to AsyncLog

cruessler:
This is a small change that makes it possible to reuse the logic in `AsyncLog` for the file history view. `AsyncLog` passes the filter to `FileLogWalker` unchanged.

@extrawurst What do you think? Does it make sense to share the revlog logic? An alternative would be to add a `FileRevlog`.


extrawurst:
looks good to me

Issue #771, opened 2021-06-05
new quit key binding


extrawurst:
Since this comes up so often my plan is to introduce a new `quit_app` key-binding that only reacts if no popup is open.

For reference: #770 , #766 

this means:
* in popups there is still `exit_popup` (ESC) to close it which allows closing text input popups that otherwise collide with `q`-key
* we still have `exit` (CTRL+C) that exits the app no matter in what place
* the new `quit_app` (Q or ESC) will close the app only when no popup is open

I am hoping to be able to cater for all the needs that where raised so far.

dbr:
Been using this change recently and it works really well. Thanks!

extrawurst:
> Been using this change recently and it works really well. Thanks!

happy to hear

PR #795, opened 2021-06-23
add bindings q and ESC for exit if no popup visible

olocoder:
This is an attempt at adding the new quit key binding [#771](https://github.com/extrawurst/gitui/issues/771)

My first PR ever, feedback is greatly appreciated


extrawurst:
Thank you! I pushed a little followup, just a few renames and cleanup: 8d6a75b

PR #812, opened 2021-07-10
Fixed formatting in "Cargo Install"

JeremyStarTM:
I fixed the formatting in the section "Cargo Install".

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ Binaries available for:
 
 ### Cargo Install
 
-The simplest way to start playing around with `gitui` is to have `cargo` build and install it with `cargo install gitui`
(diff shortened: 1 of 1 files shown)


extrawurst:
This is not really an improvement, it’s ambiguous for non-rustaceans whether you need to explicitly cargo build before cargo install. What’s the problem you try to fix here?

JeremyStarTM:
That somebody doesn't understand it. I first thought that simply entering 'cargo' is building it. It didn't work and I entered 'cargo build* and it worked. Good, a normal human person will understand that cargo build might be the command, but still.

extrawurst:
superseded by e35b196

PR #817, opened 2021-07-12
Support commit signing with gpgme

dnaka91:
Support for creating signed commits with the help of `gpgme` as discussed in #97.

Currently still WIP, especially many potential corner cases due to manually implementing `amend`
as the `git2` crate doesn't provide a function to _amend + re-sign_ commits easily.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,11 +2,11 @@ name: CI
 
 on:
   schedule:
-    - cron: '0 2 * * *' # run at 2 AM UTC
+    - cron: "0 2 * * *" # run at 2 AM UTC
   push:
-    branches: [ '*' ]
+    branches: ["*"]
   pull_request:
-    branches: [ master ]
+    branches: [master]
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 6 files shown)


dnaka91:
Currently builds fail due to the missing `gpgme` library.  made a few adjustments so it should work now on Linux and MacOS, but it's likely to still fail on Windows. I guess we'll have to set env vars accordingly.

Also, what would be the preferred way of linking? I guess only dynamic linking is an option, as static linking against GPLed libs would require to change the license of this project too.

If we link dynamically, the build process becomes more complicated, and distribution as well. We have to ask users to install the `gpgme` libraries.

- On Linux that's just a `apt-get install gpgme` (or similar for other package managers.
- On MacOS it's a simple `brew install gpgme`.
- On Windows that would be either `choco install gpg4win` or manual installation. Problem here is that the installation target dir can be customized and gitui may not find them every time.

Kqzz:
would this fix this issue?

image

mainrs:
As I already mentioned in https://github.com/sdttttt/gcr/pull/53, `gpgme` is imo not a good choice for a cross-platform binary. It it inevitabely hard to get to compile on Windows and even if one manages to compile it, it only works in 32bit mode.

I think the only viable solution would be to either bite the bullet and actually try to make gpgme work for 32bit Windows on CI. Or to move on to `sequoia_openpgp` and try to that library.

dnaka91:
I actually never really started using gitui that much so I don't really find the motivation to complete this.

Compilation on windows doesn't really seem solvable neither so it's better to go another course. As mentioned sequoia-gpg might be one solution or maybe just doing the same as git and shell out to gpg.

Therefore I'm closing this in favor of another way of adding gpg support.

PR #818, opened 2021-07-15
vim style config: allow Q in pop-up inputs

silmeth:
Change `exit` (which exits the app regardless of context) to <kbd>Ctrl</kbd>+<kbd>C</kbd>, and move <kbd>Shift</kbd>+<kbd>Q</kbd> to `quit` (which doesn’t cause quitting on upper-case `Q` insert in popup inputs, cf. #771).

+ make indendation consistent

### DIFF
--- a/vim_style_key_config.ron
+++ b/vim_style_key_config.ron
@@ -26,8 +26,8 @@
 
     open_help: ( code: F(1), modifiers: ( bits: 0,),),
 
-    exit: ( code: Char('Q'), modifiers: ( bits: 1,),),
-	quit: ( code: Char('q'), modifiers: ( bits: 0,),),
+    exit: ( code: Char('c'), modifiers: ( bits: 2,),),
+    quit: ( code: Char('q'), modifiers: ( bits: 0,),),
     exit_popup: ( code: Esc, modifiers: ( bits: 0,),),
 
     open_commit: ( code: Char('c'), modifiers: ( bits: 0,),),


extrawurst:
Hi @silmeth thanks for your PR. I am no VIM user myself. So forgive me being cautious about this. could you explain to me why this was approached with yet another set of keys in another PR? see #789 ;)

silmeth:
From what I see the issue I’m having *wasn’t* approached in the other PR.

In vim you generally use `:q` for quitting (the `:` enters command mode in which you can enter the `q` command to quit). <kbd>Ctrl</kbd>+<kbd>C</kbd> and <kbd>Esc</kbd> can both be used to exit to ‘normal mode’ (the default mode of moving around in vim), so they both also have some kind of ‘exit/quit’ semantics (though they do not exit the program) – that’s why I think <kbd>Ctrl</kbd>+<kbd>C</kbd> is fine for `exit`.

The other PR tried to change the uppercase `Q` to lowercase `q` – which kinda makes sense (since you exit vim using `:q`, not `:Q`) – but it did not fix the issue of `Q` / `q` not being possible to enter in popup windows.

I didn’t change `Q` to `q` as I didn’t want to change too much the current key mapping (I’m fine with either <kbd>Q</kbd> or <kbd>Shift</kbd>+<kbd>Q</kbd> – I only want to fix the issue of quitting on `Q` in popup inputs.

silmeth:
Since `quit` was set to `q`, arguably just changing the `exit` line to <kbd>Ctrl</kbd>+<kbd>C</kbd> would be fine too (and would change only one line) – I wanted to keep the older behaviour of exiting on <kbd>Shift</kbd>+<kbd>Q</kbd> (which I think was in the config since its introduction).

extrawurst:
> Since `quit` was set to `q`, arguably just changing the `exit` line to Ctrl+C would be fine too (and would change only one line) – I wanted to keep the older behaviour of exiting on Shift+Q (which I think was in the config since its introduction).

ok lets do that, lets only change the exit then..
what do you think @yanganto (you made the initial vi bindings, and capital Q was still your decision) ? now that we have the dedicated `quit` command, how do you feel about this change?

yanganto:
Yes, I use Shift + `Q` to quit Gitui at this moment, but I think the `Esc` to quit when non-pop window is also a good idea.

extrawurst:
> Yes, I use Shift + `Q` to quit Gitui at this moment, but I think the `Esc` to quit when non-pop window is also a good idea.

actually the idea is: `ctrl+c` to `exit` (currently `Q`)
so the question is if you ignore you habit and imagine a new gitui user comes and expects VI bindings, would `ctrl+c` be appropriate for *close gitui no matter where you are in the app* (I assume so because usually that's exactly what `ctrl+c` does) and `q` as a regular quit when in no popup.

for your personal muscle memory you can still rebind `exit` to `Q`

silmeth:
> for your personal muscle memory you can still keep `exit` bound to `Q`

or, as initially proposed, rebind `quit` to `Q` instead (since `exit` as `Q` makes it impossible to input `Q` in popups which is the main issue with this binding)

yanganto:
Thanks for mention me.  It is good for making the new users comfortable with much easier memory configure.  Changing to `ctrl` + `c` is also good to me. :)

yanganto:
rebind quit to Q instead is good.  The reason I love `Shift` + `Q` is that `ctrl` + `c` hurt my fingers, and `Shift` + `Q`  does not hurt and is not easy to trigger without intention.

yanganto:
@silmeth Thanks for making this. :smile:

yanganto (approved):

extrawurst:
@silmeth like I said, lets change `exit` to `ctrl+c`. we double checked, `tig` also exits on `ctrl+c` and `q` 👍

extrawurst:
@silmeth thank you ❤️

Issue #799, opened 2021-06-30
Improve documentation on themes, please


jeffwright13:
The default color scheme for gitui in iTerm2 is illegible on my Mac (iTmer2). The light cyan with yellow lettering can be almost impossible to see:
https://imgur.com/U7TCMv7

I am trying to change my color scheme, but the theme file doesn't seem to have an entry for the cyan command blocks with yellow (if it is listed here, please let me know). It would be good to have all possible elements and their names documented.
https://imgur.com/8qT7w63

Finally, the [THEMES.md file](https://github.com/extrawur ast/gitui/blob/master/THEMES.md) (as of 2021-06-29) incorrectly shows that the location of the theme config file (on MacOS) is $HOME/Library/Application Support/gitui/theme.ron. In fact, it is located at ~/.config/gitui. This is frustrating to a new user and in my case, was almost enough to make me uninstall. But the 'app' itself is 733t so if there is a way for me to surmout these issues, I would like to stay!

Macbook Pro 2012 Intel Catalina 10.15.7, 15" Retina, iTerm2 w/ "Bright Lights" color scheme applied.

jeffwright13:
**update**
I changed my iTerm color scheme to "Dark Default" and this made the whole program easy to see. So perhaps A modified request would be to add a blurb in the README (or wherever) that suggests changing the terminal color scheme if things are not visible?

horta:
Same here. I'm using Kitty terminal with gruvbox-like theme. Can't see the bottom options.
<img width="1474" alt="Screenshot 2021-07-08 at 09 51 38" src="https://user-images.githubusercontent.com/514522/124892796-185ff480-dfd2-11eb-8522-814022a92abf.png">

extrawurst:
> Same here. I'm using Kitty terminal with gruvbox-like theme. Can't see the bottom options.

they use `command_fg` as foreground color and `selection_bg` as background color, see the [documentation](https://github.com/extrawurst/gitui/blob/master/THEMES.md) for more details

Issue #317, opened 2020-10-08
Thoughts on checking `~/.confg/gitui` or `$XDG_CONFIG_HOME/gitui` on mac before defaulting to the existing dir?


nerdrew:
Many terminal apps use `~/.config/<app-name>` even on mac. I assume other also have `~/.confg` (or `$XDG_CONFIG_HOME/`) checked into VCS somewhere. It'd be nice to have my gitui config live with all my other terminal app configs :)

Happy to open a PR if this is welcome.

extrawurst:
I would need more input from more Mac users on this. 
@jonstodle what are your 2c on this?

jonstodle:
I honestly expect terminal apps to put config in either `~/` directly or (preferably) in `~/.config/<app name>`. Logs and temp files I still expect to be put in their idiomatic macOS locations.

terhechte:
Very longtime Mac user here; I was puzzled to find that the configuration resides in `~/Library..` as that is normally reserved for non-terminal GUI apps. Terminal apps nowadays prefer `~/.config` (compared to the `~/.` that used to be the default aeons ago)

extrawurst:
I love it, special treatment everywhere... so we can use `.config` on macOS without any conditions?

non-descriptive:
looks like [XDG](https://stackoverflow.com/questions/3373948/equivalents-of-xdg-config-home-and-xdg-data-home-on-mac-os-x) directories are conventional just among devs, otherwise it should fallback to recommended [apple fs guide](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW1).

shawnohare:
I symlink most of my configs with a stow-like bash script, and greatly appreciate apps that use XDG on Unix-like systems so that my my linking is consistent across OSes. The vast majority of the apps I use follow XDG, but there's a few outliers that prove annoying when I have to use a mac (e.g., for work).

PR #819, opened 2021-07-16
Fix the path of the theme file on mac

pe:
Thats where it is on my mac. (And on [other peoples mac](https://github.com/extrawurst/gitui/issues/799#issue-933238213))


extrawurst:
good point, this doc was not updated in https://github.com/extrawurst/gitui/issues/317

extrawurst:
good point, this doc was not updated in https://github.com/extrawurst/gitui/issues/317

Issue #381, opened 2020-10-28
View log/history of a file


jarod:
**Is your feature request related to a problem? Please describe.**
As what `gitk <filename>` do. You can view commit logs of the file specified. Then select one of the logs to view changes made to the file in that commit.

**Describe the solution you'd like**
not serious thoughts:
* on launch: `gitui <filename>` 
* on any file list view add a [file log] command which go to log tab view logs filtered by the selected file

 `any file list view` above include but not limit to:
* tab [Status] -> `Unstaged Changes` and `Staged Changes` window
* tab [Log] -> `Files` window when view a commit log
* tab [Stashing]
* tab [Stashes]

**Additional context**
`gitk Cargo.toml` on `ripgrep` repo:
![image](https://user-images.githubusercontent.com/7960/97382357-2ec39580-1906-11eb-955d-428d52fc86aa.png)

extrawurst:
file history is a feature I definitely want to support down the line!

tpo:
I'd like to add that it'd be nice if gitui had a file browser view/tab which would tell me which files/dirs are committed, staged, dirty (there's a diff), (... have merge conflicts), where I can press a key and see the diff, the history, blame, etc.

Ah, it's not hard to be wanting things ... :-)

Thanks for gitui!

extrawurst:
@tpo  I am not sure how the current `Status`-tab is not providing you a way to see what is staged or *dirty*? the rest is not covered yet, that's true.

tpo:
Hi @extrawurst

> I am not sure how the current Status-tab is not providing you a way to see what is staged or dirty?

Yes it is. However what I am missing is a view that is first and foremost file based. Not based in the first place on whether something is dirty or staged.

That said - this is my first impression: my first reaction when starting gitui was "uuuhm, where are my files?" and I instantly started searching whether there's a tab or keystroke that would bring the respective view up. 
the rest is not covered yet, that's true.

extrawurst:
@tpo I agree such a tab would have some use, but considering my most used git GUIs none of them focuses on such a view as a primary one. almost all (lazygit, tig, fork) focus on either *uncommitte changes* or *revlog*. dunno about tig/lazygit but in fork a file-tree is only accessible via a specific commit.

long story short, such a file tree has no super high priority for me personally right now.

laktak:
While I agree that it's not the most important feature it would be nice to select a path and do a git log on it.

extrawurst:
[...]

PR #841, opened 2021-08-15
Add popup for file history

cruessler:
This adds a popup for showing the history of a single file. The PR is already in a state that should give you an idea of how the final version could look like.

There are a few issues that would need to be resolved before this can be considered ready:

- Gathering the commit history is very slow, compared to running `git log -- <path>` on the command line. I will investigate and have a look at how `git` as well as `tig` do things to get an idea of how to speed things up.
- I will manually test the new popup to make sure the history matches `git log` (I already have one minor discrepancy on my list).
- `commands` does not return navigation commands for `self.list` yet.
- When I had the commit almost prepared, I noticed that the “Log” tab shows commit details on <kbd>Enter</kbd>, to the right of the log. Do we want that same behaviour here?

Looking forward to your feedback!


extrawurst:
Hi @cruessler - first of all, thanks for tackling this! The time you put into this is very appreciated!

Functionally I did not see a specific case where the history was wrong, but I did not test that very throughly yet.
Here is my list of topics I think we should tackle before merging as a MVP:

- [ ] allowing to open the history of a file from the repo file tree view (tab 3)
- [ ] scrollbar in the history view

In terms of post-MVP steps I find these two most important:

- [ ] support the details side panel you mentioned above
- [ ] use the revlog progress notifications to show a progress indicator (in percent, not in commits like in the log)

cruessler:
@extrawurst That took longer than expected, not least due to the fact that hard tabs were introduced to the codebase. :-) I hope it’s okay if I sometimes don’t get to answering your feedback immediately, but take a bit longer. Your comments are very much appreciated, though!

I addressed the first two items of your list. By adding the scrollbar to `CommitList`, we get it in the “Log” tab for free.

Let me know what you think!

extrawurst:
> @extrawurst That took longer than expected, not least due to the fact that hard tabs were introduced to the codebase. :-) 

sorry for that. but rustfmt can only enforce hard tabs and not the other way around :(

> I hope it’s okay if I sometimes don’t get to answering your feedback immediately, but take a bit longer. 

no problem at all, we all do this in our spare time ❤️

> I addressed the first two items of your list. By adding the scrollbar to `CommitList`, we get it in the “Log” tab for free.
> 
> Let me know what you think!

looks good! is there any gotchas left that you know of, or do you think we can merge this as an MVP?

extrawurst:
@cruessler I revisited #381 and realise now that we are using the regular commit list here. I don't think this is what we want for the history view. I think we should aim for something more bespoke, similar to the blame view (short commit list on the left side and a diff of the current selected commit on the right) - similar to how fork visualises it - see https://github.com/extrawurst/gitui/issues/381#issuecomment-850253697

even if that means we have to make the single entry on the left side pane more than one line but I think it should contain the following:
* short hash
* timestamp
* author 
* message (as much as fits)
* maybe even the symbol of what kind of file diff it was in this commit (add, rename, modified)

cruessler:
Having a specialized list that shows additional information makes sense to me. I'll see what I can do!

extrawurst:
@cruessler are you still on this?

cruessler:
@extrawurst Yes, I was planning on working on it this weekend!

cruessler:
@extrawurst I worked on the PR today. While the changes are not ready yet, I made good progress and will likely be able to present results over the coming week. Your proposal works quite well, I think.

ajschmidt8:
Looking forward to this feature addition!!! Thanks for all your work on this @cruessler and @extrawurst

cruessler:
@extrawurst I’ve tried adding markers for “added/modified/deleted”, but I’m a bit stuck at the moment, and could need some help/input/ideas. The list of commits currently is a `Vec<LogEntry>` and does not contain the file status. One option would be to build an async way of getting it. Another way would be to introduce a `FileLogWalker` that includes a file’s status in the commit log, but I’m worried that might duplicate a lot of code in `LogWalker`. What do you think?

extrawurst:
@cruessler I checked out what we have so far and I am a little worried to add more functionality to the current `AsyncLog` its already combining the revlog and filtering case and still uses the *old* approach without `AsyncJob`.

What do you think about building a new `FileLogJob` that internally uses `AsyncJob` via composition and hooks into the notifications to followup each batch of commitIds with looking up the markers of the file in that particular commit. this way we can reuse `AsyncJob` as is and add the state lookup in an async way.

`FileLogJob` will then only expose LogEntries we also have the marker for.

extrawurst:
@cruessler Merry Christmas 🎄
any updates ? :)

cruessler:
@extrawurst Not yet, but I want to work on the PR over the holidays. Merry Christmas! 🎄😀

extrawurst:
@cruessler cool progress!!

Screenshot 2022-01-10 at 01 31 15

extrawurst:
@cruessler if the modification type makes this overly complicated maybe we can do a MVP that we polish a bit?

* diff side-by-side
* scrollbar
* page-up/down support

extrawurst:
@cruessler ping? 🥺

cruessler:
@extrawurst Sounds like a good idea! I’m really sorry for not having responded to your earlier suggestion, unfortunately it did get lost among all the others.

cruessler:
@extrawurst I implemented your suggestions, and I think the PR is now ready for a first test and review. I have not yet cleaned up the code. I started extracting out `AsyncFileLogJob`, but I did not finish, so there are a few lines commented out that I possibly wanted to incorporate into the new trait. Let me know what you think and how to best proceed from here! And thanks for encouraging me to finish the work! :-)

asyncgit/src/file_log.rs
```diff
@@ -0,0 +1,77 @@
...
+	Response(Result<Vec<(CommitId, StatusItemType)>>),
+}
+
+///
+#[derive(Clone, Default)]
+pub struct AsyncFileLogJob {
+	state: Arc<Mutex<Option<JobState>>>,
+}
+
+impl AsyncFileLogJob {
```
extrawurst, line 22: this is not used right now, so lets rip it out or finish it - I don't wanna merge dead code

src/components/diff.rs
```diff
@@ -98,6 +98,23 @@ impl Selection {
...
+	Visible,
+	Focused,
+}
+
+impl Visibility {
+	fn is_visible(&self) -> bool {
+		*self != Self::Hidden
+	}
+
+	fn is_focused(&self) -> bool {
```
extrawurst, line 113: how is this not complaining that it is not const? 🤔

src/components/diff.rs
```diff
@@ -106,7 +123,7 @@ pub struct DiffComponent {
 	selection: Selection,
 	selected_hunk: Option<usize>,
 	current_size: Cell<(u16, u16)>,
-	focused: bool,
+	visibility: Visibility,
```
extrawurst, line 126: I think in other places we use the diff component we store the info if its visible in the owner. can we do the same here?

src/components/diff.rs
extrawurst: I actually think we do not need to hide the diff pane at all and always show it

src/components/file_revlog.rs
```diff
@@ -0,0 +1,555 @@
...
+	async_file_log: AsyncSingleJob<AsyncFileLogJob>,
+	git_diff: AsyncDiff,
+	theme: SharedTheme,
+	queue: Queue,
+	sender: Sender<AsyncGitNotification>,
+	diff: DiffComponent,
+	visible: bool,
+	repo_path: RepoPathRef,
+	file_path: Option<String>,
+	table_state: std::cell::Cell<TableState>,
```
extrawurst, line 49: lets make this a `use`

src/components/file_revlog.rs
```diff
@@ -0,0 +1,555 @@
...
+						self.theme.commit_author(false),
+					),
+				]);
+
+				let mut text = Text::from(spans);
+				text.extend(Text::raw(entry.msg.to_string()));
+
+				let cells = vec![Cell::from(""), Cell::from(text)];
+
+				// let cells = vec![
```
extrawurst, line 279: lets remove or add a fitting //TODO

src/components/file_revlog.rs
```diff
@@ -0,0 +1,555 @@
...
+						self.move_selection(ScrollType::PageDown)
+					} else {
+						false
+					};
+
+				return Ok(selection_changed.into());
+			}
+		}
+
+		Ok(EventState::NotConsumed)
```
extrawurst, line 488: a popup should always consume, right now we will control the invisible elements _behind_ it. try typing `f`

src/components/file_revlog.rs
```diff
@@ -0,0 +1,555 @@
...
+		self.show()?;
+
+		self.update()?;
+
+		Ok(())
+	}
+
+	///
+	pub fn any_work_pending(&self) -> bool {
+		self.async_file_log.is_pending() || self.git_diff.is_pending()
```
extrawurst, line 122: `git_log` missing

src/components/inspect_commit.rs
```diff
@@ -165,6 +165,7 @@ impl Component for InspectCommitComponent {
 		self.visible = true;
 		self.details.show()?;
 		self.details.focus(true);
+		self.diff.show()?;
```
extrawurst, line 168: see above, this would not be needed if we keep the DiffCompo untouched

src/keys/key_list_file.rs
```diff
@@ -130,6 +132,8 @@ impl KeysListFile {
 			enter: self.enter.unwrap_or(default.enter),
 			blame: self.blame.unwrap_or(default.blame),
 			edit_file: self.edit_file.unwrap_or(default.edit_file),
+			file_history: self.file_history.unwrap_or(default.file_history),
+			inspect_commit: self.inspect_commit.unwrap_or(default.inspect_commit),
```
extrawurst, line 136: lets make the diff open permanent, then we do not need `inspect_commit` and we can reuse `Details` to open the commit details

extrawurst:
@cruessler it really depends on how much time you have to work on this. the biggest thing is actually the missing `AsyncFileLogJob` abstraction to combine the revlog fetching and then doing the `get_commits_info` on them. since I want to merge this as soon as possible I would say lets clean up the few small comments, remove the undone and unused `AsyncFileLogJob` and lets merge this

src/components/file_revlog.rs
cruessler: There is a name conflict between `std::cell::Cell` and `tui::widgets::Cell` both of which are used in `file_revlog.rs`. Currently, the latter is accessible as `Cell`, but I think it makes sense to switch. What do you think?

extrawurst: oh I see, never mind then

asyncgit/src/file_log.rs
cruessler: I deleted the code.

src/components/diff.rs
cruessler: Makes sense. I implemented your suggestion and made `DiffComponent` always visible.

src/components/diff.rs
cruessler: I removed `visibility` and restored `focused`.

src/components/file_revlog.rs
cruessler: Removed.

src/components/file_revlog.rs
cruessler: I added `return Ok(EventState::Consumed)` inside the `if self.visible` block.

src/components/file_revlog.rs
cruessler: Thanks for catching that!

src/components/inspect_commit.rs
cruessler: I removed the line.

src/keys/key_list_file.rs
cruessler: I completely removed `inspect_commit`.

cruessler:
@extrawurst Next week I probably have little free time, so from my perspective it makes sense to merge first and address the outstanding issues later. I’d be happy to address them myself, but I totally understand if you don’t want to wait for that. I’m sure I would find other open issues. :-)

I found the reason for the file contents not being shown and pushed an additional commit.

extrawurst:
@cruessler thanks for sticking with this! this is a great addition! checkout when this effort started: august 2021! Awesome work🥳

cruessler:
@extrawurst I’m super happy this finally got merged! :tada: Thanks a lot for being so patient!

PR #842, opened 2021-08-15
check branch name validity while typing

### LINKED ISSUE #559: on-demand branchname validity (while typing)
libgit2 exposes a helper function to check a branch name validity: https://libgit2.org/libgit2/#HEAD/group/branch/git_branch_name_is_valid

lets run this against the input **before** it gets confirmed.
this way we can give live feedback to the user whether the name is valid.

extrawurst:
closes #559


extrawurst:
ezgif com-optimize

PR #860, opened 2021-08-19
compare two commits

### LINKED ISSUE #852: compare two commits
since 25a49e2 we can mark entries in the log:

Screenflick Movie 11

we can use that to open a diff of two different commits

extrawurst:
closes #852


extrawurst:
todo

- [ ] show the author, date, she for *old* and *new*
- [ ] don't show message box
- [ ] allow triggering compare via revolt

PR #866, opened 2021-08-21
Issue/768/support markdown emoji

### LINKED ISSUE #768: Support Markdown emoji
I'm using a lot of markdown emoji in my commits message, but in gitui, the emoji are not parsed.

Replace all markdown emoji by it's Unicode equivalent.

I have found this lib that can help: https://lib.rs/crates/gh-emoji 

**Additional context**
image

This should look like this:
image

andrewpollack:
Thank you again for the advice on the previous PR! This is helping me learn better Rust coding practices, and I very much appreciate it :smile: 

This Pull Request fixes/closes #768.

It changes the following:
- Adds `gh-emoji` crate to dependencies.
- Updated changelog with png of change in action.
- Added new `fn emojifi_string(s: &str) -> String` to `utils/mod.rs`.
- Utilized above function to emojify messages in `LogEntry`.

I followed the checklist:
- [:heavy_check_mark: ] I added unittests
- [:heavy_check_mark: ] I ran `make check` without errors
- [:heavy_check_mark: ] I tested the overall application
- [:heavy_check_mark: ] I added an appropriate item to the changelog


andrewpollack:
My local `cargo fmt -- --check` fails on this line here https://github.com/extrawurst/gitui/blob/779a811b58a756f0dfcd8358d9a6025219142f77/asyncgit/src/sync/tags.rs#L45 , but when I make the suggested changes, they fail for the linting that's run in the CI/CD here. Any thoughts on what I should do here?

src/components/utils/mod.rs
```diff
@@ -53,6 +53,14 @@ pub fn string_width_align(s: &str, width: usize) -> String {
 	}
 }
 
+// Replace markdown emojis with Unicode equivalent
+// :hammer: --> 🔨
+#[inline]
+pub fn emojifi_string(s: &str) -> String {
+	let r = gh_emoji::Replacer::new();
```
extrawurst, line 60: Replacer uses a regex internally and creating this over and over again is considered an antipattern. We should use lazy_static to do it just once.

src/components/utils/mod.rs
```diff
@@ -53,6 +53,14 @@ pub fn string_width_align(s: &str, width: usize) -> String {
...
+	r.replace_all(&*s).to_string()
```
extrawurst, line 61: Too bad we cannot replace in-place here and we create a string copy everytime even if no emoji was found. That’s a deal breaker.
We should take the string in as a mutable reference And check whether the Cow is Cow::Owned (the result from replacer) and only then change the original string

extrawurst (changes requested):

src/components/utils/mod.rs
andrewpollack: Agreed -- happy to find ways to fix the performance. A couple of questions -- 
1. Would this mean that the function would be: `pub fn emojifi_string(s: &mut Cow<&str>) -> String`?
2. Since the `msg` of `LogEntry` is typed String, does this mean I would need to cast it using this `Cow`? I'm a little unfamiliar with `Cow`, but my initial thought is that casting `c.message` in any form would mess up with this check. Is there any example of this pattern of converting `String` into `Cow` without clobbering the typing checks?

andrewpollack: I gave it a go, and did what I believe should only change the original string should an emoji be present! 
I should note I am quite new to Rust, so thank you for your patience, and I appreciate the guidance in this PR :smile: 

As well, I had to change the original `CommitInfo` header to `mut` to allow the passing of `&mut c.message`!

extrawurst: you still made a `to_string` even if the replacer still only had a borrow, thats what we do not want. I cleaned that up and I think its good to go now

extrawurst (approved):

extrawurst:
@andrewpollack thanks for your contribution. Checkout d72dcddf14f8684b5d686ad74600c24a763eb15d To see what I improved

src/components/utils/mod.rs
andrewpollack: Thank you for this! This is super valuable for me to look at and reference when doing clean implementations in the future :]

PR #903, opened 2021-09-10
Debianize

Elbandi:
This commit adds the required files to build for ubuntu/debian. Tested(=builded) on ubuntu bionic LTS and focal LTS, they already have new rust. Unfortunately debian havent new rust, it cannot build with main packages. (but you can install rust packages from focal)


debian/changelog
```diff
@@ -0,0 +1,5 @@
+gitui (0.17.1-1) unstable; urgency=medium
```
extrawurst, line 1: is there no way to populate this via the actual changelog we have?

debian/compat
```diff
@@ -0,0 +1 @@
+11
```
extrawurst, line 1: what does that mean?

debian/control
```diff
@@ -0,0 +1,44 @@
+Source: gitui
+Section: vcs
+Priority: optional
+Maintainer: Stephan Dilly <dilly.stephan@gmail.com>
+Build-Depends:
+ debhelper (>= 11),
+ dh-cargo (>= 2),
+ cargo:native,
+ rustc:native (>= 1.51)
```
extrawurst, line 9: we actually need 1.50 as a minimum rust version currently.

debian/gbp.conf
```diff
@@ -0,0 +1,4 @@
+[DEFAULT]
+debian-branch = debian/master
+upstream-branch = master
+upstream-tag = v%(version)s
```
extrawurst, line 4: that is not the git tag I assume? otherwise the 's' at the end might be wrong

debian/install
```diff
@@ -0,0 +1 @@
+debian/tmp/usr/bin/gitui
```
extrawurst, line 1: is that the install location? why would we install into `tmp`?

extrawurst (commented):
Hi @Elbandi thanks for tackling this. but in order for me to merge this I had to ask a bunch of noob question, please bear with me. also: do you have a good source I can read up on this procedure for future reference? I am no Debian guy so I this might seem stupid :)

debian/changelog
Elbandi: i dont know how. :( but it is not usual to do this. this changelog is for "debian packaging changelog", but the project own changelog can be included as document to /usr/share/doc

debian/compat
Elbandi: it's for debhelper compatibility minimum level.

debian/control
Elbandi: oks, i found only a 1.51 rustc package for testing

debian/gbp.conf
Elbandi: The %()s might be familiar from [Python format strings](https://docs.python.org/2/library/stdtypes.html#string-formatting). so it is just a formatter character.

debian/install
Elbandi: the upstream (here the cargo command) puts the installed files in this temp directory. 
after that the debhelper selects the files, and move to the right package path (eg X file to A package, Y file to B package)

boydjohnson:
I'm just an onlooker with an opinion, but I don't think this PR is the best approach. It would require users to build the binary each time they install the package. A better approach is using [cargo deb](https://crates.io/crates/cargo-deb) to build a deb binary package. Then there would be no dependency on cargo and rustc.

extrawurst:
@boydjohnson this PR seems abandoned anyway. Would you be up to work on a PR to add deb creation to the pipeline?

boydjohnson:
Yeah, I will create a PR this week that modifies the `.github/workflows/cd.yml` file, `Cargo.toml`, and possibly `Makefile` to create a binary deb on github actions.

Elbandi:
This is the debian package build guide: https://wiki.debian.org/HowToPackageForDebian
My PR follows this standard.

Debian/ubuntu releases isnt a bleeding edge distro. They pick up a _stable release_ from github, and build the source and puts the package to repo.

If you want build a package for every commit, there is solution for do that in "debian way" (eg: create a new changelog entry from commit messages at build step).

boydjohnson:
I think you are right, @Elbandi. I will test building a deb source and binary from this PR.

debian/rules
```diff
@@ -0,0 +1,22 @@
...
+
+override_dh_auto_clean:
+	cargo clean
+	rm -rf $(CARGO_HOME)
+
+override_dh_auto_test:
+	cargo test --release
+
+override_dh_auto_build:
+#	cargo build --release
```
boydjohnson, line 19: Is this commented out?

boydjohnson:
Hi @Elbandi I was able to build a deb binary using `dpkg-buildpackage -b --no-sign -r=fakeroot`. I did this on Ubuntu. I successfully installed the deb on Debian (Bullseye) and Ubuntu (Bionic) (host machine).

This looks good to me. Let me know what I can do to help.

debian/rules
Elbandi: Yepp, cargo install do again the build process internally. so this "cargo build" is useless. the empty _override_dh_auto_build_ is needed, to "drop" the default build command.
Maybe i sould write the reason here. :/

Elbandi:
here is a sample: https://github.com/dmx-systems/dmx-build-deb/blob/cf0e0292c67d601b9a1404763c5a4756fd9d6b6c/.gitlab-ci.yml#L162-L164
this creates a new changelog entry, and build the debian package. but you can tweak with `git describe --tags` the build version. etc.etc

boydjohnson:
@extrawurst I gave bad information in 

> I'm just an onlooker with an opinion, but I don't think this PR is the best approach. It would require users to build the binary each time they install the package. A better approach is using [cargo deb](https://crates.io/crates/cargo-deb) to build a deb binary package. Then there would be no dependency on cargo and rustc.

Specifically that cargo deb is the better approach. This PR makes it possible to build debian source and debian binary packages. From reading at the debian wiki link up above, this is the only way that we could get it included in an official apt repository.

extrawurst:
@boydjohnson I am not sure I can follow. does this mean your PR is obsolete? i can't merge this one either so we are back to square one.

boydjohnson:
@extrawurst If this PR built deb packages in github actions would you be willing to merge it?

extrawurst:
@boydjohnson yes we need this to be part of CI *and* CD - also its would be great to document how to update for new release (like by now we are already at 0.19 where in the PR it is still 0.17.1

boydjohnson:
@Elbandi Would you be willing to do the work of adding building the deb to both github actions workflows?

Also, since @extrawurst is listed as the maintainer, he would have to provide a gpg key to sign the source. @extrawurst, do you have a gpg key and are you comfortable signing the source?

extrawurst:
@boydjohnson can you elaborate what exactly needs to be signed? I never signed any commit in here, I guess you are not talking about that, right? what do I need to gpg-sign and why?

extrawurst:
I am sorry I am such a noob when it comes to deb. feel free to take this conversation to discord (link on readme) to simplify teaching me

boydjohnson:
No problem, @extrawurst. Yeah, so as maintainer of the debian package, you would sign the source so that people could verify that it is what you say it is. The key would probably be a secret that you added to github and would be used in the github actions to sign. Let's see what @Elbandi says.

Elbandi:
I rebased the branch to the latest release, and started to play with GH build. but there are some mysterious errors and segfaults. only ubuntu amd64 builds work :(
i need to work on it...
https://github.com/Elbandi/gitui/actions/runs/1551533874

tmpgnh:
@Elbandi you need help with this? I'm using a Debian/bookworm and was able to `debuild` your branch with no problems (well, a single lintian error on the chlog file, maybe false positive).

extrawurst:
setting to draft for the time being

debian/control
```diff
@@ -0,0 +1,44 @@
+Source: gitui
+Section: vcs
+Priority: optional
+Maintainer: Stephan Dilly <dilly.stephan@gmail.com>
+Build-Depends:
+ debhelper (>= 11),
```
williamdes, line 6:
```suggestion
 debhelper-compat (= 13),
```

debian/compat
williamdes, line 1: this is not needed anymore: https://salsa.debian.org/phpmyadmin-team/shapefile/-/commit/14f00dbf72ebc4404325f1cc0ab63996ec9424da
https://lintian.debian.org/tags/package-uses-old-debhelper-compat-version.html

debian/gbp.conf
```diff
@@ -0,0 +1,4 @@
+[DEFAULT]
+debian-branch = debian/master
```
williamdes, line 2:
```suggestion
debian-branch = debian/latest
```
To be compatible with DEP-14: https://dep-team.pages.debian.net/deps/dep14/

williamdes (approved):
I am a Debian Maintainer (DM) (just a random guy that maintains packages on Debian/Ubuntu)
An willing to help package this one

See:
- https://wiki.debian.org/Teams/RustPackaging
- https://salsa.debian.org/rust-team/debcargo-conf/blob/master/README.rst

Let me know if you need help

semioticrobotic:
Just wanted to express my desire to see this implemented! I need to move to Pop!_OS and would love to have gitui in the upstream repositories on Ubuntu Jammy.

williamdes:
> Just wanted to express my desire to see this implemented! I need to move to Pop!_OS and would love to have gitui in the upstream repositories on Ubuntu Jammy.

You should file a RFP: https://wiki.debian.org/RFP
For that you will need to send an email to submit, here is an example: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1030845;msg=5

Elbandi:
@williamdes i checked your suggestions, and you have (almost) right.
My goal was to make a package to everyone can use (and _build_) on yet _supported_ systems, not only for bandnew bleedingedge versions: strech on debian (in 2021 was supported...), and bionic LTS on ubuntu.

MitaiDart:
I just installed Debian 12 and by the apt install gui command it finds nothing
does it make sense to wait or is it abandoned?

lonix1:
The project already supports many OSs/distros - look at the README. Debian and derivatives (Ubuntu notably) should be included too, given their massive popularity. Leave this open so someone with the requisite knowledge can assist someday.

alerque:
@lonix1 I think you missed my point. Of course this project should get Debian & Ubuntu packaging! I fully support that. I'm also aware of the other distros that package this. I am the package maintainer for the Arch Linux package for example—which also accounts for the Manjaro, Parabola, and other supported distros. I'm also a contributor to Homebrew, Nix, Alpine, OpenSUSE, Void Linux and others on that list. The point is all those distros that have packages have the packaging source material in their own packaging repositories, ***not in this repository***. This is the way it should be with this too.

This is a code PR trying to push the packaging support files into this repository. I believe that particular action should be turned down, hence this PR should be close. A tracking issue for requesting support and discussing who can help where would be fine, but not a PR attempting to land this code here.

tillea:
Andreas Tille from Debian speaking:  We strongly prefer the *Debian packaging* on salsa.debian.org (and we happily accept MRs there).   My 27 years experience of debian/ dirs bundled into upstream source is that it was not helpful for the Debian maintenance and our packaging tools even ignore/throw away the debian/ dir if bundled into the upstream source.  Please don't do it and rather talk to the Debian maintainers.
Thank you for supporting Debian anyway.

lonix1:
@alerque If that's how it's done for those other packages, then the same should apply here too - agreed.

PR #924, opened 2021-09-29
Use git_message_prettify

### LINKED ISSUE #917: Use git_message_prettify
I think we right now do the message cleanup manually but we can and should use `git_message_prettify` ([message_prettify](https://docs.rs/git2/0.13.22/git2/fn.message_prettify.html))

R0nd:
This Pull Request fixes/closes #917.

It changes the following:
Adds message cleanup using built-in git_message_prettify function on commit
Changes custom cleanup for git_message_prettify after using external commit message editor

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/src/components/commit.rs
+++ b/src/components/commit.rs
@@ -11,7 +11,7 @@ use crate::{
 };
 use anyhow::Result;
 use asyncgit::{
-	cached,
+	cached, message_prettify,
 	sync::{
 		self, get_config_string, CommitId, HookResult, RepoState,
 	},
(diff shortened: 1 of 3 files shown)


src/components/commit.rs
```diff
@@ -235,6 +223,19 @@ impl CommitComponent {
 		Ok(())
 	}
 
+	fn prettify_message(&mut self, msg: String) -> String {
```
extrawurst, line 226: no need for this function

src/components/commit.rs
```diff
@@ -193,7 +181,7 @@ impl CommitComponent {
 			)));
 			return Ok(());
 		}
-		let mut msg = msg;
+		let mut msg = self.prettify_message(msg);
```
extrawurst, line 184: we can use ?-op to propagate a result

extrawurst (changes requested):

src/components/commit.rs
R0nd: I figured that the potential `Error` result from `message_prettify` needed handling in the context of this class, and since it's used twice might as well make it a function. Do you suggest we should duplicate the error handling for each use case?

extrawurst:
@R0nd thanks ❤️

PR #948, opened 2021-10-16
Improve redraw performance

### LINKED ISSUE #869: decouple input from re-draw
* drawing is significantly slower than input
* this is most notable in complex widgets like the (rev)log
* we should PoC if it is improving the experience if we decouple both
* this would mean instead of every input triggering a redraw, we buffer up inputs that happen during a ongoing redraw

R0nd:
This Pull Request fixes/closes #869.

It changes the following:
- Improves redraw performance by moving sync repo operations off the UI thread

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -387,7 +387,7 @@ impl App {
 			self.push_tags_popup.update_git(ev)?;
 			self.pull_popup.update_git(ev);
 			self.select_branch_popup.update_git(ev)?;
-			self.commit.update_git(ev)?;
+			self.commit.update_git(ev);
 		}
 
 		self.files_tab.update_async(ev);
--- a/src/components/commit.rs
+++ b/src/components/commit.rs
@@ -81,22 +81,14 @@ impl CommitComponent {
 	}
 
 	///
-	pub fn update_git(
-		&mut self,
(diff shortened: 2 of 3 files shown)


extrawurst:
> I tried scrolling through revlog continuously, and batching up to 10 input events was enough to clear the input event congestion (for a debug build on my setup)

ok I was asking if you can see a noticeable difference though?

R0nd:
I've recorded screencasts do demonstrate the case - in both recordings I was holding the down arrow key until the revlog passed ~1400 commit number. The master build keeps processing the congested events long after the 1400 commit mark and the version with event batching stops scrolling just in time, showing the absence of congested events.
[master](https://asciinema.org/a/ObfY2hOyubpxy3ll0DvYHO6Lg?t=6)
[queue-updates](https://asciinema.org/a/J3PgxM90r6JmKPCylUWjVrGsN?t=11)

R0nd:
The revlog update in particular, it runs a `sync::get_commits_info` on the UI thread. Which is the reason of #869 in the first place.

extrawurst:
if we want to go forward with this we would need to bring it up to date with master

PR #954, opened 2021-10-21
Make gh-emoji optional

jirutka:
gh-emoji crate includes *images* of GitHub's emoji - this is quite big dependency. It increases the binary size by 1 MiB; that's +25 % when building v0.18.0 on Alpine Linux with build flags to optimize size. I consider it an unnecessary bloat that should be optional.

Note: I’m maintainer of Alpine Linux’s [gitui](https://pkgs.alpinelinux.org/packages?name=gitui) package ([APKBUILD](https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/community/gitui/APKBUILD)).


jirutka:
The Security audit job is misleading, the reported issue is not related to the changes I made:

> Error: Unable to publish audit check! Reason: HttpError: Resource not accessible by integration

PR #962, opened 2021-10-24
Popup history

### LINKED ISSUE #846: popup back button history
currently we have no history of where the user comes from when opening the blame/commit-detail popup. would be nice to keep track of that and allow going back to where the user came from.

simple use case:
* go to file tab
* choose a file and open blame
* choose a commit and inspect inside blame
* commit details are open now
* close commit details
* we are back in the file tab

I would expect to be back in the blame popup

R0nd:
This Pull Request fixes/closes #846.

It changes the following:
- implements popup history

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
- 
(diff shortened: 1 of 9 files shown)


extrawurst:
the problem with this approach is:
* The way we open/close popups using channels is not enough: it does not preserve the parameters each popup takes in its `open` function, which the history might overwrite
* I am not a fan of the ergonomics, having a channel for each popup

extrawurst:
closed in favour of #1124

PR #982, opened 2021-11-10
use Path to fix blame filename seperators (#981)

### LINKED ISSUE #981: blame file path wrong on windows
based on https://github.com/extrawurst/gitui/pull/791

extrawurst:
closes #981

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -642,7 +642,7 @@ impl App {
 				self.tag_commit_popup.open(id)?;
 			}
 			InternalEvent::BlameFile(path) => {
-				self.blame_file_popup.open(&path)?;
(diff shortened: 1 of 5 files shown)


extrawurst:
@enizor can you please checkout this branch and see if it fixes the issue on windows?

enizor:
Still got `<no blame available>` :/

Without debugging, I'd guess the problem stays the same - `PathBuf` still uses `\` when formatted into a `&str` while `git-revparse-single` ([asyncgit/src/blame.rs:55](https://github.com/extrawurst/gitui/blob/3c8eb7e0490bb2ff78ffb81629476824b3793279/asyncgit/src/sync/blame.rs#L55)) expects `/`.

PR #993, opened 2021-11-17
File find popup up/down keys resolves #977

### LINKED ISSUE #977: Fuzzy Finder: Move up/down with vim keys
- I also don't know how to scroll to other selections -- `j`/`k` don't work given the context.

- I'd like to be able to select other matches. 

If it's already possible to have multiple key mappings per command, then I can simply do that and map `C-j` and `C-k` for this. If not, I guess the solution would be to split out `move_up` and `move_down` to read-only lists and "picker" style lists.

Curiosidad-Racional:
This Pull Request fixes/closes #977.

It changes the following:
- adds up and down key for file find popup

I followed the checklist:
- [ ] I added unittests


vim_style_key_config.ron
```diff
@@ -44,6 +44,8 @@
     end: ( code: End, modifiers: ( bits: 0,),),
     move_up: ( code: Char('k'), modifiers: ( bits: 0,),),
     move_down: ( code: Char('j'), modifiers: ( bits: 0,),),
+    popup_up: ( code: Char('p'), modifiers: ( bits: 2,),),
```
extrawurst, line 47: shouldn't this use arrow keys because p/n will be used as text input in the fuzzy finder?

Curiosidad-Racional: As you know, `( bits: 2,)` means control key, then `C-p/C-n` will be used for up and down, and `p/n` would be typed.

extrawurst: The problem with these modifiers is that they are notoriously inconsistently supported in the various different terminals

extrawurst: And additionally they often get picked up as special commands - but hey I am no expert with the vim key bindings - I trust you there 👍

extrawurst:
Please add an appropriate entry to the changelog

vim_style_key_config.ron
Curiosidad-Racional: Thank you

Curiosidad-Racional:
> Please add an appropriate entry to the changelog

Added entry and resolved conflicts.

extrawurst:
Thank you @Curiosidad-Racional 👍

PR #1001, opened 2021-11-21
simplify key bindings file handling

### LINKED ISSUE #946: do not create or replace 'key_config.ron'
gitui creates the configuration file (`~/.config/gitui/key_config.ron`) when executed the first time.   
If you decide to do some changes to this file to hold your own key-mappings then if an error is found or a key is missing this file is overwritten with new information. Since I have all my config files in git (dotfiles) and the file is just a symlink it will break.

1. don't create a file.
2. if a file is found, the file contains only the remaps that the user wants, everything else should be as default
3. If an error is found, ignore the error and show a small error footer in the ui.

If, for some reason, `gitui` still wants to create the configuration file when executed for the first time. Then a user-created file with the name `key_config_user.ron` could be used to override some or all settings.

**Additional context**

thanks for a great project 👍

extrawurst:
closes #946

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
- 
+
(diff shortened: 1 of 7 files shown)


sheeley:
Thanks so much for this!

PR #1051, opened 2021-12-19
Disable fetch,pull,push on local only repo

### LINKED ISSUE #1047: status: disable pull cmd on local-only branches
based on conversations in #1045

uniqueNullptr2:
This Pull Request fixes/closes #1047

It changes the following:
dis able the following buttons and functions in repos without any remote branches
- push in status tab
- pull in status tab
- fetch in branches window of status tab

I followed the checklist:
- [ ] I added unittests
- [x] I ran `make check` without errors -> shows error event fn too long but is unrelated to my changes
- [x] I added an appropriate item to the changelog -> not sure where

### DIFF
--- a/src/components/branchlist.rs
+++ b/src/components/branchlist.rs
@@ -247,7 +247,7 @@ impl Component for BranchListComponent {
 				if !self.local {
 					self.has_remotes =
 						get_branches_info(&self.repo.borrow(), false)
-							.map(|branches| branches.len() > 0)
+							.map(|branches| !branches.is_empty())
 							.unwrap_or(false);
 				}
 				self.update_branches()?;
--- a/src/components/branchlist.rs
+++ b/src/components/branchlist.rs
@@ -42,6 +42,7 @@ pub struct BranchListComponent {
 	repo: RepoPathRef,
 	branches: Vec<BranchInfo>,
 	local: bool,
+	has_remotes: bool,
 	visible: bool,
 	selection: u16,
 	scroll: VerticalScroll,
@@ -200,7 +201,7 @@ impl Component for BranchListComponent {
 
 			out.push(CommandInfo::new(
 				strings::commands::fetch_remotes(&self.key_config),
-				true,
+				self.has_remotes,
 				!self.local,
 			));
 		}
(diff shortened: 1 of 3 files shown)


uniqueNullptr2:
After looking through the errors I notice 2 distinct errors happening.

most of the time clippy complains about the eventfunction being too long. Unlike I wrongly stated in the first message this seems to be related to me adding a few lines to that function which is already pretty convoluted.
As a fix  I could move the assignment of has_remote to the new function.

the other error I am not sure. it happens in asyncgits logwalker but reading the small snipped shown by clippy could it be related to adding new properties to status and branchlist? this error only seems to happen on nightly.

src/components/branchlist.rs
```diff
@@ -243,6 +244,12 @@ impl Component for BranchListComponent {
 					.map(Into::into);
 			} else if e == self.key_config.keys.tab_toggle {
 				self.local = !self.local;
+				if !self.local {
```
extrawurst, line 247: this needs to go into a `check_remotes` fn or something and it needs to be called also on `update` so that it will update for changes happening outside of gitui

src/tabs/status.rs
```diff
@@ -66,6 +67,7 @@ pub struct Status {
 	index_wd: ChangesComponent,
 	diff: DiffComponent,
 	git_diff: AsyncDiff,
+	local_only: bool,
```
extrawurst, line 70: lets not cache this here unless we identify this is an expensive operation

src/tabs/status.rs
```diff
@@ -583,6 +590,11 @@ impl Status {
 		self.git_branch_state
 			.as_ref()
 			.map_or(true, |state| state.ahead > 0)
+			&& !self.local_only
+	}
+
+	const fn can_pull(&self) -> bool {
+		!self.local_only
```
extrawurst, line 597: we can do the `get_branches_info` check in here. I think it was a rather quick function to run

extrawurst:
@uniqueNullptr2 please update your branch with master, then the nightly errors should be fixed.

>  ... appropriate item to the changelog -> not sure where

Check the changelog, tons of examples there, its a fix, you can do it :)

extrawurst:
@uniqueNullptr2 thanks for implementing this. I changed it to not close #1047 because this PR only fixes it for repos without any remote. it does not disable pull on local-only-branches when there is a remote defined

PR #1091, opened 2022-01-22
fix credential.helper usage

### LINKED ISSUE #800: It should support auto login for push
When I use git command, I don't have to type the user name and password  everytime for push. Because I saved my password by [credential helper](https://stackoverflow.com/questions/11403407/git-asks-for-username-every-time-i-push). 

If the user is using a credential helper feature, the gitui should not ask a user name and password

extrawurst:
closes #800

### DIFF
--- a/asyncgit/src/sync/cred.rs
+++ b/asyncgit/src/sync/cred.rs
@@ -103,11 +103,14 @@ fn git_credential_fill(url: &str) -> Option<BasicAuthCredential> {
 	let protocol = url.scheme();
 
 	let cmd = format!("protocol={}\nhost={}\n", protocol, host);
-	let cmd = format!("printf \"{}\" | git credential fill", cmd);
+	let cmd =
+		format!("printf \"{}\" | git credential-store get", cmd);
 
 	let bash_args = vec!["-c".to_string(), cmd];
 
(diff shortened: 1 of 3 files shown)


asyncgit/src/sync/cred.rs
```diff
@@ -70,6 +80,69 @@ pub fn extract_username_password(
...
+	scope_time!("git_credential_fill");
+
+	let url = url::Url::parse(url).ok()?;
+
+	let host = url.domain()?;
+	let protocol = url.scheme();
+
+	let cmd = format!("protocol={}\nhost={}\n", protocol, host);
+	let cmd =
+		format!("printf \"{}\" | git credential-store get", cmd);
```
jgoldhammer, line 107: I would change to this

```suggestion
		format!("printf \"{}\" | git credential fill", cmd);
```

extrawurst: problem with that is that it breaks if you don't have credentials stored yet. not sure why though (yet)

jgoldhammer: You are right - I had not checked that. 
`git credential fill` asks for the credentials interactively...

Hmm- I do not see any other solution yet.

PR #1107, opened 2022-01-26
Fix 1104 performance regression

### LINKED ISSUE #1104: [0.20] performance regression repo_state in hot-path
after fixing #1102 another smaller one got visible:

Screenshot 2022-01-26 at 00 27 42

obviously we need to cache the `repo_state`

extrawurst:
closes #1104


extrawurst:
Screenshot 2022-01-26 at 09 52 20

much better

PR #1124, opened 2022-02-02
generic popup stacking solution

### LINKED ISSUE #846: popup back button history

extrawurst:
closes #846 

- [x] blame popup can be returned to
- [x] revision-files can be returned to (+file selection?)
- [x] inspect-commit can be returned to
- [x] file-history can be returned to (+selection)
- [x] commit-compare can be returned to

### DIFF
--- a/filetreelist/src/filetree.rs
+++ b/filetreelist/src/filetree.rs
@@ -54,6 +54,11 @@ impl FileTree {
 	}
 
 	///
+	pub const fn selection(&self) -> Option<usize> {
+		self.selection
+	}
+
+	///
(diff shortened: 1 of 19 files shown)


extrawurst:
Screenflick Movie 35

PR #1170, opened 2022-03-09
switch focus to index after staging last file

### LINKED ISSUE #1169: auto switch focus when staging all
after staging the last item we are supposed to switch to the staging area automatically to be quickly able to commit.
this works the other way around from empty-staging-to-workdir:

Screenflick Movie 44

extrawurst:
closes #1169


extrawurst:
not an ideal solution, see code comment

PR #1183, opened 2022-03-31
Sort fuzzy_matcher results based on score

mark2185:
It changes the following:
- exact matches have a higher priority and are placed first in the results

I followed the checklist:
- [ ] I added unittests

Can you give me some pointers regarding how to test this?

I tested it manually, previously by searching files with the pattern `gif` (in the `gitui` repo), results such as `github/FUNDING.yml` were above the ones actually containing `gif` in its name.


extrawurst:
good catch! don't you want to add something to the changelog?

mark2185:
> good catch! don't you want to add something to the changelog?

Sure do! Added an entry under `Fixed`.

extrawurst:
Thanks for this contribution 🥳

PR #1201, opened 2022-04-20
Fix - stashlist removes marked commit IDs from state after stash drop

lightsnowball:
This Pull Request fixes issue related to dropping multiple stashes (no issue number as far as I can tell). Once we drop marked stash, GitUI's state becomes invalid and the user is not able to drop/pop stash anymore. The scenario is covered in the gif below (after dropping, there remains 1 stash commit in the marked commits state).

gitui

It changes the following:
Fixes the described problem.

I followed the checklist:
- [ ] I added unittests

I'll add unittests (not sure how should I test it tbh) and changelog later if this fix makes sense.


extrawurst:
thanks for discovering the bug, your solution to fix it was way too complicated for me to understand. please don't take this personally and use the opportunity to check how I fixed it in https://github.com/extrawurst/gitui/pull/1207

PR #1237, opened 2022-06-16
mention macports in install section

fs111:


README.md
```diff
@@ -126,6 +126,12 @@ sudo emerge dev-vcs/gitui::dm9pZCAq
 brew install gitui
 ```
 
+### MacPorts (macOS)
```
extrawurst, line 129: I don't know macports but it would be great if the title also links to the receipe or whatever its called in this case

fs111: done

extrawurst:
Thanks ❤️

PR #1250, opened 2022-07-05
Allow to scroll diffs horizontally

### LINKED ISSUE #1017: Allow to scroll diffs horizontally
Sometimes there are diffs that a longer then console allows. It would be nice to have some kind of horizontal scroll to view whole diff. 
image

First idea how to make it - allow scroll horizontally when right arrow button pressed in diff view. When left arrow pressed - reset horizontal scroll and change focus to changed/staged. It could be not really ergonomic in some cases since diff view resets.

Second idea - allow to reorder panels and make staged and unstaged stand side-by-side on top part and diff on the most of the bottom part of the console. The bad in this case - we need either divide vertical scrolling with buttons other then up/down arrow or make complicated jump keys to change focus back to file panels.

M1cha:
This Pull Request fixes/closes #1017.

This is a simple implementation that works really well according to my testing.

It may need some improvements like an UI indicator for both the current column and the keymap.
We should talk about that as part of this PR.

One issue that I've found:
It is hard to use keyrepeat on the `left` key to scroll to the beginning without leaving the diff because starting from column 0 the key suddenly has a different meaning and jumps all the way back up.
A solution would be to provide separate keymaps for scrolling.

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * switch focus to index after staging last file ([#1169](https://github.com/extrawurst/gitui/pull/1169))
 * fix stashlist multi marking not updated after dropping ([#1207](https://github.com/extrawurst/gitui/pull/1207))
 * exact matches have a higher priority and are placed to the top of the list when fuzzily finding files ([#1183](https://github.com/extrawurst/gitui/pull/1183))
+* support horizontal scrolling in diff view ([#1017](https://github.com/extrawurst/gitui/issues/1017))
 
 ## [0.20.1] - 2021-01-26
 
--- a/src/components/blame_file.rs
+++ b/src/components/blame_file.rs
@@ -123,6 +123,7 @@ impl DrawableComponent for BlameFileComponent {
 				//
 				// https://github.com/fdehau/tui-rs/issues/448
 				table_state.selected().unwrap_or(0),
+				ui::Orientation::Vertical,
 			);
 
 			self.table_state.set(table_state);
--- a/src/components/commitlist.rs
+++ b/src/components/commitlist.rs
@@ -7,7 +7,7 @@ use crate::{
 	keys::SharedKeyConfig,
 	strings::{self, symbol},
 	ui::style::{SharedTheme, Theme},
-	ui::{calc_scroll_top, draw_scrollbar},
+	ui::{calc_scroll_top, draw_scrollbar, Orientation},
 };
 use anyhow::Result;
 use asyncgit::sync::{CommitId, Tags};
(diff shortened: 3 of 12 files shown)


extrawurst:
looks nice! lets paint a scrollbar only as long as `scroll` is not `0`

M1cha:
I'm not sure if that's the best way since you might still wanna scroll vertically while being scrolled horizontally and still see where in the document you you are.

extrawurst:
Don’t know what the horizontal scrolling has to do with it

M1cha:
well you can scroll both directions at the same time and while being scrolled horizontally I'd still like to know where I am vertically.

extrawurst:
Still no idea why we worry about the existing scrollbar (vertical).
For the horizontal Scrollbar I would try to not convolute the view as long as we did not start scrolling to the right.

M1cha:
I've implemented a scroll bar :) A few things:
- it's using the vertical bar symbol. that looks a little weird but I couldn't find a better one
- calculating the longest line is not yet correct. 1) I'd have to convert tabs to spaces (again) before doing the calculation, but also, it only accounts for text, not for any meta data. I tried asking the span list for the longest line but that seems to trip on the selection which always has the full width. any ideas?

extrawurst:
looks like a good start. but yeah it definitely misses the correct max-scroll pos. I did not have time to check the code yet. you can go for what ever solution as long as it is accurate and cache the result if it is potentially expensive (it would scale with the length of the diff) as long as the diff is not updated

extrawurst:
are you still going to work on this?

cruessler:
@extrawurst, @M1cha I have a few relatively minor changes locally that address the issues I have found so far, but I don’t have push access to this PR. I can open a separate PR from my fork and link to this one there. Is that ok?

[cruessler/diff-horizontal-scroll](https://github.com/cruessler/gitui/tree/diff-horizontal-scroll)

M1cha:
that'd make sense. btw the longest_line calculation was insane. It's better to either use for loops like this:
```rust
self.longest_line = 0;
if let Some(diff) = self.diff.as_ref() {
	for hunk in &diff.hunks {
		for line in &hunk.lines {
			self.longest_line = self.longest_line.max(line.content.len());
		}
	}
}
```

or something a little more functional:
```rust
self.longest_line = self
	.diff
	.iter()
	.flat_map(|diff| diff.hunks.iter())
	.flat_map(|hunk| hunk.lines.iter())
	.map(|line| line.content.len())
	.max()
	.unwrap_or(0);
```

I pulled these snippets out of old chats with co-workers about this and didn't verify if those solutions still match what's currently on my branch.

cruessler:
@M1cha Thanks for the hint, I was able to incorporate your suggestion into the new PR!

extrawurst:
closed in favour of #1327

PR #1256, opened 2022-07-13
Add word motions to text input

Rodrigodd:
It changes the following:
- Add common word motions to the text input component, more precisely `Ctrl-Left`, `Ctrl-Right`, `Ctrl-Backspace`/`Ctrl-W` and `Ctrl-Delete`. A word motion move/delete until the next/previous start of a word. A word is being defined as a continuous sequence of alphanumeric characters.

I followed the checklist:
- [x] I added unittests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * submodules support ([#1087](https://github.com/extrawurst/gitui/issues/1087))
 * customizable `cmdbar_bg` theme color & screen spanning selected line bg [[@gigitsu](https://github.com/gigitsu)] ([#1299](https://github.com/extrawurst/gitui/pull/1299))
 * use filewatcher instead of polling updates ([#1](https://github.com/extrawurst/gitui/issues/1))
-* word motions to text input ([#1256](https://github.com/extrawurst/gitui/issues/1256))
+* word motions to text input [[@Rodrigodd](https://github.com/Rodrigodd)] ([#1256](https://github.com/extrawurst/gitui/issues/1256))
 
 ### Fixes
 * remove insecure dependency `ansi_term` ([#1290](https://github.com/extrawurst/gitui/issues/1290))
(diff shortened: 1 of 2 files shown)


extrawurst:
this does not work on macos

Rodrigodd:
I have tested this on Windows and Linux (WSL). Unfortunately I don't have access to a macOS to test this. Can you tell me how exactly this is not working? The only platform dependent code I think might not be working is the `ctrl` modifier check, but this seems to be equivalent to the others in the codebase.

extrawurst:
can you please record how this should look like?

Rodrigodd:
@extrawurst I recorded it inside visual studio code for the keystrokes overlay. I notice that `Ctrl-Delete` don't work in vscode integrated terminal, but it does work in Windows Terminal.

https://user-images.githubusercontent.com/51273772/188209077-e6a13fdf-4f86-493b-a7eb-7cef66daa2d6.mp4

extrawurst:
it breaks on unicode text with multibyte content, try some emojis. it will run into some endless loop

Rodrigodd:
@extrawurst the infinite loop is fixed.

src/components/textinput.rs
```diff
@@ -366,6 +435,43 @@ impl Component for TextInputComponent {
...
+					KeyCode::Delete if is_ctrl => {
+						if let Some(pos) = self.next_word_position() {
+							self.msg.replace_range(
+								self.cursor_position..pos,
+								"",
+							);
+						}
+						return Ok(EventState::Consumed);
+					}
+					KeyCode::Backspace | KeyCode::Char('w')
```
extrawurst, line 447: I am on the fence about the special case of `ctrl+w`. I am probably going to remove it for now. its hard to customize this way and I am not aware this being some standard

extrawurst:
also the changelog entry is still missing

src/components/textinput.rs
Rodrigodd: As far as I know, Ctrl-W is the standart shortcut for "delete word" in a linux terminal (Ctrl-Delete don't work).
Because gitui is a terminal app, I tought this shortcut would be pertinent.
Vim, for example, has the same shortcut (it is where I learned about it).

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * submodules support ([#1087](https://github.com/extrawurst/gitui/issues/1087))
 * customizable `cmdbar_bg` theme color & screen spanning selected line bg [[@gigitsu](https://github.com/gigitsu)] ([#1299](https://github.com/extrawurst/gitui/pull/1299))
 * use filewatcher instead of polling updates ([#1](https://github.com/extrawurst/gitui/issues/1))
+* word motions to text input ([#1256](https://github.com/extrawurst/gitui/issues/1256))
```
extrawurst, line 18: see the format of other contributions please - your handle is missing

extrawurst (changes requested):
not much left to do. could you record a nice gif we can put in the changelog of the new functionality?

Rodrigodd:
> could you record a nice gif we can put in the changelog of the new functionality?

I don't have a easy recording workflow setup.

extrawurst:
Thanks for implementing this!

Issue #1115, opened 2022-01-30
refactor file history lookup into AsyncFileLogJob


extrawurst:
followup on #841

The idea is to abstract the job of fetching the revlog commits touching the file in question and parallelise this with look up of the commits that match to figure out what was the `subject line`,`author`... and the file modification type. right now this is all happening in the application level code which is not ideal and harder to unittest.

The simple solution is to build this in phases:  1) revlog 2) commit Infos 3) diffs (for the modification type)
The downside of this approach is: we only all all the information of the file log once everything is done and not in iterative batches.

The more complex but nicer solution: The AsyncJob spawns another Thread-Worker. While the AsyncJob does the revwalk we send every commitId we find to match into a queue and the Worker picks it up and does the info and diff lookup. As soon as the worker is done with one CommitId this will popup right away in the file history list. This should give a very reactive and fast feeling.

@cruessler what do you think?

cruessler:
@extrawurst I think the second option sounds like it gives the best user experience. Is something like this already implemented somewhere so that it could serve as a starting point?

extrawurst:
you can checkout the other examples of the `AsyncJob` being implemented. none of them is spawning a child thread though but that should not be much more complicated. think you want to try it? I can help you once you have a draft

cruessler:
I’ve started addressing the issue in #1267.

PR #1267, opened 2022-08-03
Factor out file history lookup into AsyncFileLogJob

cruessler:
This PR partly addresses #1115. As of 2022-08-03, it is still a draft.

It changes the following:
- It factors out getting commit info into `AsyncFileLogJob`. It does not move getting commit info into a worker thread.

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/asyncgit/src/revlog.rs
+++ b/asyncgit/src/revlog.rs
@@ -269,15 +269,21 @@ impl AsyncJob for AsyncFileLogJob {
 		&mut self,
 		_params: RunParams<Self::Notification, Self::Progress>,
 	) -> Result<Self::Notification> {
+		let mut notification = AsyncGitNotification::FinishUnchanged;
+
 		if let Ok(mut state) = self.state.lock() {
 			*state = state.take().map(|state| match state {
-				JobState::Request(_, _) => JobState::Response(()),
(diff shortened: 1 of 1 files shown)


cruessler:
@extrawurst Currently, components can use `AsyncLog::get_slice` to get a slice of the revlog even if it has not been fully read yet. In this PR, I want to first implement the existing functionality using `AsyncJob`, so it seems I would need something similar to `get_slice`. Is that assumption correct? If so, can this be done using `AsyncJob`? There is `set_progress`, but I’m not sure whether this is the right function to use. I’m even starting to wonder whether my overall approach is correct.

extrawurst:
the AsyncLog specific do not have to be squeezed into the trait functions `run/progress`. check out `AsyncSyntaxJob::result` it is not part of the generic interface and can be used specific to the use case. Not sure this answers your question. WDYT?

extrawurst:
@cruessler is this still relevant? or can it close

cruessler:
@extrawurst I’m going to close as the codebase has probably changed significantly since I started working on this. I started wondering, though, whether we might want to consider moving some of `gitui`’s async code to something like `tokio`. Have you got any experience with `tokio`?

PR #1299, opened 2022-08-29
Improve UI selection and command bar

Gigitsu:
It changes the following:
- Selected line in log and files tab now fill the entire container
- Command bar visible commands uses a different background colour than the selection_bg colour

### Before the patch
files tab before

log tab before

### After the patch
files tab after

log tab after

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 * submodules support ([#1087](https://github.com/extrawurst/gitui/issues/1087))
+* selected lines in files and log lists now fills the entire container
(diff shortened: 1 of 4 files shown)


Gigitsu:
The nightly build on windows won't compile.

I can see this error started from this [commit on master](https://github.com/extrawurst/gitui/commit/ef3ece552dc04414ee6c69b7ad41186106bbfe58) but I cannot find the reason. I tried to [compare my last commit to last working commit](https://github.com/extrawurst/gitui/compare/ef3ece552dc04414ee6c69b7ad41186106bbfe58...962397f57b671e3bae33bfbf09cdd71ea1bc8f8c) without finding a reason for the error.

extrawurst:
looks good to me, will make the command bar color default back to blue on master after merge

PR #1322, opened 2022-09-10
fix: returns 0 exit code on error

### LINKED ISSUE #1315: Gitui returns zero exit code on error
**Describe the bug**

When calling `gitui` in a non-git directory, I get the expected error:
```shell
> gitui
invalid path
please run gitui inside of a non-bare git repository
```
However the exit code is zero:
```shell
> echo $?
0
```

**Expected behavior**
I'd expect a non-zero exit code e.g. `1`.

This is on macOs 12.5.1 using `gitui` installed via `nix`
```
> gitui --version
gitui 0.21.0
```

**Additional context**
This is a problem when scripting shortcuts to call `gitui` in a directory as errors get swallowed.

snapdgn:
This Pull Request fixes/closes #1315.

It changes the following:
- changes exit code to 1, while initializing gitui in a bare git repo (let me know if the exit code needs to be something else)

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -113,7 +113,7 @@ fn main() -> Result<()> {
 
 	if !valid_path(&cliargs.repo_path) {
 		eprintln!("invalid path\nplease run gitui inside of a non-bare git repository");
-		return Ok(());
+		process::exit(1);
 	}
 
 	let key_config = KeyConfig::init()
--- a/src/main.rs
+++ b/src/main.rs
@@ -112,8 +112,7 @@ fn main() -> Result<()> {
 	asyncgit::register_tracing_logging();
 
 	if !valid_path(&cliargs.repo_path) {
-		eprintln!("invalid path\nplease run gitui inside of a non-bare git repository");
-		process::exit(1);
+		bail!("invalid path\nplease run gitui inside of a non-bare git repository");
 	}
 
 	let key_config = KeyConfig::init()


src/main.rs
```diff
@@ -113,7 +113,7 @@ fn main() -> Result<()> {
 
 	if !valid_path(&cliargs.repo_path) {
 		eprintln!("invalid path\nplease run gitui inside of a non-bare git repository");
-		return Ok(());
+		process::exit(1);
```
extrawurst, line 116: Did you check if there other places that should return non zero?

snapdgn: If you mean, checking at all potential places for all kinds of errors which should've returned a non-zero exit code. Sorry, I haven't. Will do :)

SimonAdameit: Thanks for tackling this! Wouldn't it be more ideomatic to use `return Err(..)` or `bail!(...)` here, as `main` returns a `Result`? I think an error should get turned into exit code `1` automatically.

SimonAdameit (changes requested):

src/main.rs
snapdgn: hey, thanks a lot for the review. I completely agree with you, will update this accordingly. Also, please do let me know, if there are any other places i could potentially look into, would be very grateful. Thanks!

extrawurst:
Thanks!

PR #1371, opened 2022-10-03
Display tags and branches in the revlog

### LINKED ISSUE #1252: Show tags and branch heads in the log view.

I use `tig` to see where my branch is relative to another branch (typically main). I'd like to switch to `gitui` because it parses large git histories much faster, but I need to be able to see tags and branch heads in the log view.

Display tags and branch heads in the log view.

**Additional context**

The same repo in `gitui` and `tig`.

`tig` shows the tags and branch heads in orange.

image

alexmaco:
fixes #1252

It changes the following:
- adds fg colors to the theme for tags and branches
- displays local branches as labels on their respective commit heads

I followed the checklist:
- [ ] I added unittests


alexmaco:
@extrawurst This only displays local branches. Afaict asyncgit cannot report the upstream branch for a local branch, and that info would need to be added to `struct LocalBranch`, which would remember the name instead of a bool. Should we also do that change?

extrawurst:
Thank you

PR #1375, opened 2022-10-04
Add no-verify commit command

### LINKED ISSUE #1374: No verify option on commits
I am coming over from `magit`, I am really enjoying the application.  Thank you for this work!

I have a lot of pre-commit routines setup but sometimes I just need to commit in cases where I know they are going to fail, such as in the middle of a large refactor.

In my fork I have added `Commit no-verify` option in the commit workflow, which at least enables me to do this in a basic sense.  I will make a PR and see what you think.  However, I think it would be better to add a flag / option for `no-verify` so that it could be re-used by the other commit commands like amend. 

Cheers!
Dave

dam5h:
This Pull Request fixes/closes #1374.

It changes the following:
-  Adds a "Commit no-verify" option to the commit workflow.

I think there is likely a cleaner way that just adds a "no-verify" or "no-hooks" option, but maybe this works fine.  It seems the the `amend` workflow could probably still use this.

I followed the checklist:
- [ ] I added unittests


extrawurst:
yeah I think I would prefer an option that we can toggle with the new key and that resets back to the `verify` default after a commit.

dam5h:
I have been using this and it does work fine with the amend workflow.  I can look into adding it as an option, this makes for less "action" items on the menu bar, but does add a keystroke in the workflow.  You have to set the option and then hit commit.  As is you just hit `ctrl-f` to force a commit.  Maybe one use case for `--no-verify` is that you are is a rush anyway so I've found not having the extra step of a toggle (like in `magit`) to be kinda nice. 

That said adding options to this menu could well be extended into other flags as well, so that could be a good thing.

I'd have to dig through the code a bit more to see how you are handing options, but happy to do so.  Thanks for taking a look here.

Cheers!
Dave

dam5h:
Howdy @extrawurst , I've updated this to be a flag on the `CommitComponent` struct, which can be toggled on or off via `Ctrl-f`.  When verify is `true` (the default), it shows `Disable hooks [^f]` in the menu, once toggled the menu shows `Enable hooks [^f]`  

Let me know if this is what you had in mind, re: "toggle with new key".  I'll be taking it for a spin for the rest of the week.

Cheers,
Dave

extrawurst:
lgtm! Thanks you. will create the relevant changelog

PR #1376, opened 2022-10-04
Add notification when correctly copying hash commit

### LINKED ISSUE #1160: Provide some feedback after copying commit hash
I use the `Copy Hash [y]` feature in the log view quite often. Unfortunately, this feature offers no visual feedback when the key was hit.

It would be great to get some visual feedback when the hash was copied.
E.g. show a small overlay / popup, that the hash was successfully copied.

SergioRibera:
This Pull Request fixes/closes #1160.

It changes the following:
- When copying the hash of a commit, it displays a confirmation dialog if it has been copied correctly

I followed the checklist:
- [ ] I added unittests


alexmaco:
I also ran into these usability issues with copying.

Maybe it would be useful to display the text that was copied in the notification?

Additionally, copying might fail (e.g. if both xclip and xsel are missing or if gitui is being run via ssh). The notification shouldn't report success without checking the result of the copy operation. Maybe a good place to catch that error is at the call site of `copy_commit_hash` in `commitlist.rs`?

SergioRibera:
> I also ran into these usability issues with copying.
> 
> Maybe it would be useful to display the text that was copied in the notification?
> 
> Additionally, copying might fail (e.g. if both xclip and xsel are missing or if gitui is being run via ssh). The notification could report that failure instead when it happens. Maybe a good place to catch that error is at the call site of in ?`copy_commit_hash``commitlist.rs`

You are very right, I think it would be very useful to know what text is being copied, on the other hand I also agree with you, I think it is useful to show the error message if the copy fails, in that case I did nothing because I did not know what to do with that error, now I have an idea, so I will be making changes and will update the PR

src/components/commitlist.rs
```diff
@@ -26,6 +27,8 @@ use tui::{
 };
 
 const ELEMENTS_PER_LINE: usize = 9;
+const FAIL_COPY: &str = "Failed to copy the Text";
```
extrawurst, line 30: these should go into our separate strings.rs file

extrawurst (changes requested):
see above

src/components/commitlist.rs
SergioRibera: Done

SergioRibera:
@extrawurst That's it, I fixed all the clippy and formatting errors in the project (I think it should go in another PR but the CI was not working due to some base errors)

Cries in CI/CD *
image


If you need any changes, please let me know

extrawurst:
https://github.com/extrawurst/gitui/pull/1390 is merged now, so please update with current master again

SergioRibera:
@extrawurst Done, now a query please, what version of rust do you use?

extrawurst:
Please rebase again and resolve the conflicts so we can merge this

extrawurst:
please fix the clippy errors

SergioRibera:
@extrawurst Done 😃

extrawurst:
Thanks for sticking at this!

PR #1380, opened 2022-10-06
Display commit description in addition to hash in file view

alexmaco:
It changes the following:
- displays the commit name besides the hash in the file view

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/src/components/revision_files.rs
+++ b/src/components/revision_files.rs
@@ -11,10 +11,14 @@ use crate::{
 	AsyncAppNotification, AsyncNotification,
 };
 use anyhow::Result;
-use asyncgit::sync::{self, CommitId, RepoPathRef, TreeFile};
+use asyncgit::sync::{
+	self, get_commit_info, CommitId, CommitInfo, RepoPathRef,
+	TreeFile,
+};
 use crossbeam_channel::Sender;
 use crossterm::event::Event;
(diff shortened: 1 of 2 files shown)


extrawurst:
Screenshot 2022-10-15 at 12 25 39

im not convinced of this yet, this rarely fits the screen. maybe we just introduce a shortcut to open the current commit detail popup from the files component, this will contain the commit desc. swell

alexmaco:
> this rarely fits the screen

There are multiple things that can't be shown fully when the screen is too small to fit them. But most often it fits on a terminal for me.

The main use of this is remembering which commit this is exactly, and by proxy, sometimes which branch. It's easy to forget which commit does the current file view point at, and even the start of the message is usually enough to distinguish between several commits as a mnemonic.

Do you see any substantial downsides to having an at-a-glance disambiguation?

> maybe we just introduce a shortcut

Imo there are many shortcuts already, but no consistent model of interaction yet (hence, #1383).

Being able to switch to a commit view would be really nice if it would always work like a toggle, regardless of how the file view was opened, but that would introduce some overlap with the current use of Esc, and some unpredictable (i.e. special-cased) view switching. So off-hand it seems like something that's easier to get wrong than right without some more detailed UX consideration.

extrawurst:
ok lets have a changelog entry and it can be merged

PR #1381, opened 2022-10-07
Use ctrl+arrow to switch focus between panes in the status tab

### LINKED ISSUE #1295: [Help] Shortcut issues
How do I jump from one hunk to next without visiting each line? I'm SSH-ed into a Linux machine where `gitui` runs from a Mac.

alexmaco:
Currently, the only way to switch between the tree and stage is by scrolling through the entire list.
This adds a way of doing it quickly for very long lists.

This Pull Request fixes/closes #1295 .

It changes the following:
- adds 4 new key combos (focus_pane_$direction), defaulting to ctrl+arrow
- uses those combos in the status tab to switch views

Right now it doesn't add these combos as commands, since the available commands is already very crowded and confusing, but I can add the commands if necessary.

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/src/keys/key_list.rs
+++ b/src/keys/key_list.rs
@@ -112,6 +112,10 @@ pub struct KeysList {
 	pub view_submodule_parent: GituiKeyEvent,
 	pub update_submodule: GituiKeyEvent,
 	pub commit_history_next: GituiKeyEvent,
+	pub focus_pane_right: GituiKeyEvent,
+	pub focus_pane_left: GituiKeyEvent,
+	pub focus_pane_above: GituiKeyEvent,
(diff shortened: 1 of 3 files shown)


extrawurst:
> Currently, the only way to switch between the tree and stage is by scrolling through the entire list.

that’s not true there is a hot key for that already

extrawurst:
You can simply rebind that toggle to whatever you prefer

alexmaco:
> that’s not true there is a hot key for that already

Somehow I missed that, I thought it would send the current tree item to stage instead :laughing:.

alexmaco:
I think #1295 can probably be closed in that case.

PR #1390, opened 2022-10-14
Fix Clippy Lints

alexmaco:
This Pull Request fixes the current CI failures

It changes the following:
- applies latest nightly clippy lints
- disables `clippy::missing_const_for_fn` due to hight number of false positives on nightly

I followed the checklist:
- [ ] I added unittests
- [x] I ran `make check` without errors
- [x] I tested the overall application
- [ ] I added an appropriate item to the changelog

PR #1397, opened 2022-10-19
Generate rust toolchain file - to develop (and build)

SergioRibera:
This PR is to add the rust-toolchain file that allows all those who want to contribute or run the project to do it under the same version of Rust

## Inspiration
When I did my PR, I did it working under the version of rust that I use in my day to day (a nightly version), which caused confusions and problems when incorporating the changes, I have also seen something similar happen in the PR of [clippy fixes](https://github.com/extrawurst/gitui/pull/1390), so I think it is necessary to be able to work together in a safe and orderly way.

This Pull Request not fixes/closes any

It changes the following:
- Add rust-toolchain file

I followed the checklist:
- [ ] I added unittests
- [ ] I added an appropriate item to the changelog (For this change is it required?)

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,3 @@
+[toolchain]
+channel = "1.60.0"
+profile = "default"
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.60.0"
+channel = "stable"
 profile = "default"


extrawurst:
I am always using latest stable, is that something we can define in the file?

extrawurst:
ok please make it track "stable" like the CI does too

extrawurst:
Thank you!

extrawurst:
damn it seems to fuck up the CI and overwrite the compiler version chosen in there. so I have to roll it back

extrawurst:
see d92d43a

SergioRibera:
> damn it seems to fuck up the CI and overwrite the compiler version chosen in there. so I have to roll it back

Mmmmm, let's open the PR again then, now tell me, don't you think this is good? so we can define in this file the version of Rust that will be used everywhere? If not, then how about defining the channel with the version used in the IC?

On the other hand I am reading the CI logs and I don't see any problem, would you be so kind to explain it to me please?

extrawurst:
Here is the CI after I merged your change: https://github.com/extrawurst/gitui/actions/runs/3290393076

Here is the run after I Reverte your change:
https://github.com/extrawurst/gitui/actions/runs/3290616619

It breaks on linux-musl builds.

Unfortunately I did not run CI on your change first. my mistake.

Feel free to reopen and fix.

SergioRibera:
> Here is the CI after I merged your change: https://github.com/extrawurst/gitui/actions/runs/3290393076
> 
> Here is the run after I reverted your change: https://github.com/extrawurst/gitui/actions/runs/3290616619
> 
> It breaks on linux-musl builds.

Oooooh, great, I see what happened, yep, I can fix it, give me a moment and I'll open the PR again.

PR #1411, opened 2022-10-26
Minor refactors and deduplications

alexmaco:
It changes the following:
- performs some minor refactors

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/src/components/branchlist.rs
+++ b/src/components/branchlist.rs
@@ -115,6 +115,9 @@ impl Component for BranchListComponent {
 				out.clear();
 			}
 
+			let selection_is_cur_branch =
+				self.selection_is_cur_branch();
+
 			out.push(CommandInfo::new(
 				strings::commands::scroll(&self.key_config),
 				true,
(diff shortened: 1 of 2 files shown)


src/components/branchlist.rs
```diff
@@ -363,23 +364,16 @@ impl BranchListComponent {
...
 	/// fetch list of branches
 	pub fn update_branches(&mut self) -> Result<()> {
 		if self.is_visible() {
-			self.check_remotes();
 			self.branches =
 				get_branches_info(&self.repo.borrow(), self.local)?;
-			//remove remote branch called `HEAD`
+
 			if !self.local {
+				self.has_remotes = !self.branches.is_empty();
```
extrawurst, line 374: I think this will only set the `has_remotes` properly if in the local-branchlist, that's not what we want?!

extrawurst (changes requested):

src/components/branchlist.rs
alexmaco: Indeed, I missed that difference. I'll remove that change.

extrawurst:
superseded by 5b2b8c7

PR #1419, opened 2022-11-04
Add Linux targets for ARM, ARMv7 and AARCH64

### LINKED ISSUE #1343: aarch64 build?

Excitedly installed this on my ARM Linux box by following [these instructions](https://github.com/extrawurst/gitui/issues/1018#issuecomment-1102308583) and got:

```
$ gitui
bash: /usr/local/bin/gitui: cannot execute binary file: Exec format error
```

Please build for `aarch64`

N/A

**Additional context**

My Ubuntu install:

```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.4 LTS
Release:        20.04
Codename:       focal

$ uname -a
Linux mylinuxbox 5.4.0-92-generic #103-Ubuntu SMP Fri Nov 26 16:15:10 UTC 2021 aarch64 aarch64 aarch64 GNU/Linux

$ dpkg --print-architecture
arm64

$ arch
aarch64
```

adur1990:
This Pull Request fixes/closes #1343.

It changes the following:
- Adds compilation target for ARM, ARMv7 and AARCH64

As now code was changed, no tests beyond executing the binary are needed.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,0 +1,8 @@
+[target.aarch64-unknown-linux-gnu]
+linker = "aarch64-linux-gnu-gcc"
+
+[target.arm-unknown-linux-gnueabihf]
+linker = "arm-linux-gnueabihf-gcc"
+
+[target.armv7-unknown-linux-gnueabihf]
+linker = "arm-linux-gnueabihf-gcc"
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -10,9 +10,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest, ubuntu-22.04]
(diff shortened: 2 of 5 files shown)


extrawurst:
Thanks!
Please also add a changelog entry

extrawurst:
why is this specific to `ubuntu-22.04`?

adur1990:
It is not depending on ubuntu-22.04, but I use this more or less as a placeholder OS to build the ARM targets. The alternative I considered was to build the ARM targets when `matrix.os = 'ubuntu-latest'`, i.e., when the regular Linux target is built. But I thought that having the ARM targets built separately makes more sense. Unfortunately, however, there was no obvious and easy way (at least I can't see it) to achieve this using the current build chain.

extrawurst:
ok I think then we should make it use `ubuntu-latest`. you can always make this a separate step entirely or introduce another variable into the matrix build to check for. lets not misuse a pinned os version for it. this otherwise looks like a specific intent (like I thought in this case)

PR #1430, opened 2022-11-14
Edit file from file tree

### LINKED ISSUE #1409: Launch Editor from Compare Commits?
### Discussed in https://github.com/extrawurst/gitui/discussions/1407

<div type='discussions-op-text'>

<sup>Originally posted by **sez11a** October 24, 2022</sup>
Stumbled upon GitUI, and I think it's great! It's missing one thing (as far as I can tell) that prevents me from using it. My use case is reviewing pull requests from others. To do that, I check out their branch, rebase master on top of it, and then launch GitUI. 

In the Log view, I tag the top commit and the last commit from master, and then hit C to compare. GitUI shows a beautiful tree on the left with only the changed files in it and a diff on the right. If I could highlight one of those files and hit Enter or E to launch my default editor (neovim), GitUI would become my only Git client. 

I'm not sure if I'm missing something in the config or if this is a feature request. Is this possible? </div>

should be rather easy to implement. we even have such an edit-command in the file view/tab already.

extrawurst:
fixes #1409


extrawurst:
see new `edit` command on the bottom:

Screenshot 2022-11-14 at 14 22 50

PR #1499, opened 2023-01-07
Checkout commit

### LINKED ISSUE #983: [Feature] Allow checking out specific commits in log view
Hi! I just found this project and think it's pretty cool!

Anyway, I think it would be nice if the log view had an option to enter detached HEAD state and checkout the selected commit.

fralcow:
This Pull Request fixes/closes #983 

It changes the following:
- Adds a command to a log view to checkout a specific commit
- Added an entry to the log bottom hint bar with command name and a key to execute it.

I followed the checklist:
- [x] I added unittests
- [ ] I ran `make check` without errors -- sync::submodules::tests::test_smoke throws an error for me, both in master and my branch: `Error { code: -1, klass: 16, message: "the server did not provide a certificate" }`. I would appreciate some help fixing this.

The PR by the original author went stale, I hope we can add this function in this PR.


asyncgit/src/sync/branch/mod.rs
```diff
@@ -301,6 +301,36 @@ pub fn checkout_branch(
...
+	let cur_ref = repo.head()?;
+	let statuses = repo.statuses(Some(
+		git2::StatusOptions::new().include_ignored(false),
+	))?;
+
+	if statuses.is_empty() {
+		repo.set_head_detached(commit_hash.into())?;
+
+		if let Err(e) = repo.checkout_head(Some(
+			git2::build::CheckoutBuilder::new().force(),
```
extrawurst, line 321: this will overwrite changes made to the index, right? I am not sure we want `force` to be true

asyncgit/src/sync/branch/mod.rs
```diff
@@ -665,6 +695,37 @@ mod tests_checkout {
...
+
+	#[test]
+	fn test_smoke() {
+		let (_td, repo) = repo_init().unwrap();
+		let root = repo.path().parent().unwrap();
+		let repo_path: &RepoPath =
+			&root.as_os_str().to_str().unwrap().into();
+
+		write_commit_file(&repo, "test_1.txt", "test", "commit1");
+		write_commit_file(&repo, "test_2.txt", "test", "commit2");
```
extrawurst, line 713: `write_commit_file` returns the `CommitId`, you don't need to use `LogWalker` to find them

asyncgit/src/sync/branch/mod.rs
extrawurst: I tested, it in fact will fail with an `Err` if there are uncommitted changes. se more comments below

src/components/commitlist.rs
```diff
@@ -435,6 +440,15 @@ impl CommitList {
 		self.selection = position;
 	}
 
+	pub fn checkout(&mut self) -> Result<()> {
+		if let Some(commit_hash) =
+			self.selected_entry().map(|entry| entry.id)
+		{
+			checkout_commit(&self.repo.borrow(), commit_hash)?;
```
extrawurst, line 447: the fact that it fails if there are uncommitted changes makes this err bubble up here. we should use `try_or_popup` macro here. to show a proper error popup to the user instead of closing the app

extrawurst (changes requested):
small improvement on the error handling and this is good to go 💪

asyncgit/src/sync/branch/mod.rs
fralcow: done! thanks for the tip.

src/components/commitlist.rs
fralcow: done!

fralcow:
Thank you for your comments! I've fixed the code, let me know if more stuff needs fixing. :muscle:

extrawurst:
Thank you @fralcow

PR #1508, opened 2023-01-15
Use filetreelist crate

abergmeier:
No longer safe Status inside of `FileTreeItem`.
For Status storage, now introduced a new `Item`, which contains Status alongside `FileTreeItem`.

Note that `FileTreeItemKind` is no longer publicy available - thus we cannot store this enum anywhere anymore.

I followed the checklist:
- [ ] I added unittests


abergmeier:
So far change component shows no diff - not sure why

abergmeier:
I think for being able to write unittests I would need to extract a lot of traits for components.

extrawurst:
All good the tree component itself is pretty well tested

src/components/status_tree.rs
```diff
@@ -305,11 +303,12 @@ impl StatusTreeComponent {
 }
 
 /// Used for drawing the `FileTreeComponent`
-struct TextDrawInfo<'a> {
+struct TextDrawInfo {
 	name: String,
 	indent: u8,
 	visible: bool,
-	item_kind: &'a FileTreeItemKind,
+	status_item: Option<StatusItem>,
```
extrawurst, line 310: this looks like it will require a lot of allocations where only references where needed before

abergmeier: Will have a look.

extrawurst:
While reviewing noticed that it still has its own `FileTreeItems` instead of using the one from `filetreelist`-crate. We should make it use the crate one

extrawurst:
putting in draft till the next round of reviews

abergmeier:
> While reviewing noticed that it still has its own `FileTreeItems` instead of using the one from `filetreelist`-crate. We should make it use the crate one

The reason is that `filetreelist` crate knows nothing about status. If you want to use it I see 2 basic ways of doing it:
1. Introduce status to `filetreelist`
2. Have a separate "list" for status

How do you want handle that?

PR #1511, opened 2023-01-17
Refactor key_list and remove  key_list_file

yanganto:
It changes the following:
- refactor key_list 
- remove key_list_file

I followed the checklist:
- [X] No additional unittests are needed, one test case refactored
- [X] There is no feature change that need to descript in the changelog


yanganto:
Hi @extrawurst,
Thank you for accepting vim key binding before. I upgrade gitui on my old machine and feel sorry that so many people keep editing the `key_config.json` each time gitui upgraded.  I put some work into refactoring the KeyList, hope you like it.  Hope you have a good start of year 2023. :smile:

extrawurst:
@yanganto you have to tell me what this changes? the concept of the `KeysListFile` was already to create less breakage when changing/extending `KeysList`. you PR moves that manual code over to use a crate for it but does not improve UX for the user, or am I missing something?

yanganto:
Sorry, the removed file just goes back after solving the conflict master, so this PR makes you confuse.

The UX is not different because this is a refactor thing.  It is good to let a code focus on the main objective, so it will be easier to keep rolling out new features in the future. If a new developer with normal key binding wants to add a new feature here and needs to first modify more than one keybinding-related file, and fight compiling errors, the process will not straight forwarded anymore.  For example, we use `try_or_popup` also for the same purpose, we do want a contributor can easily focus on the feature itself.  Because the keybinding is the base but trivial thing for this project, it already took a lot of time from others to add new features and waste time on this.  

As an open-source project, it will become bigger and bigger in the future,  it will become impossible to let developers know everything, and it still can rock and roll.  (We all know no one knows every code in Linux, and it keeps rolling nowadays.)  If we have just 5 key bindings can we do this, it will be overkill.  Today, we have more than two-page keybindings, so I think this is a need to consider the DX.

The PR is not 100% needed, because this is a refactor thing.  For an open-source project, people easier to look back at what they introduce.  As the developer grows after 2~3 years, he can look back and will know to learn some limits by the vision itself.  This PR is I look back and know the limitation of myself 2 years before, and willing to have a better implementation on this.  I hope I have a chance to fix this.

extrawurst:
please rebase, so that d4d6fd2 get into this PR

yanganto:
Thanks for the review. :heart:

PR #1516, opened 2023-01-18
Copy file path

yanganto:
It changes the following:
- allow copy file path in revision files
- allow copy file path in status tree

I followed the checklist:
- [X] This feature is not a git operation, I think we do not need a test on copy feature.


src/components/status_tree.rs
```diff
@@ -302,6 +302,21 @@ impl StatusTreeComponent {
 			selection_offset_visible,
 		)
 	}
+
+	// Copy the real path of selected file to clickboard
+	fn copy_file_path(&self) {
+		if let Some(item) = self.selection() {
+			if crate::clipboard::copy_string(&item.info.full_path)
```
extrawurst, line 309: why not use `try_or_popup` here aswell?

src/components/status_tree.rs
```diff
@@ -481,6 +496,9 @@ impl Component for StatusTreeComponent {
 						}
 					}
 					Ok(EventState::Consumed)
+				} else if key_match(e, self.key_config.keys.copy) {
```
extrawurst, line 499: if we add a new event handler we should generally also add a new entry to the command bar

src/components/revision_files.rs
```diff
@@ -515,6 +516,15 @@ impl Component for RevisionFilesComponent {
 					);
 					return Ok(EventState::Consumed);
 				}
+			} else if key_match(key, self.key_config.keys.copy) {
```
extrawurst, line 519: if we add a new event handler we should generally also add a new entry to the command bar

extrawurst (changes requested):

src/components/status_tree.rs
yanganto: The data struct of this component is different from others, so we can not use this macro here.

Issue #1149, opened 2022-02-25
Sign git commits with SSH keys


skyfaller:
**Is your feature request related to a problem? Please describe.**
I would like to be able to sign git commits with SSH keys from gitui.

- https://blog.dbrgn.ch/2021/11/16/git-ssh-signatures/
- https://calebhearth.com/sign-git-with-ssh
- https://lobste.rs/s/l5h2ct/git_ssh_signing_add_commit_tag_signing

This is a new feature that seems to only work properly with OpenSSH 8.8 and above, which can make it challenging to test on more stable OSes; but if you are able to use it, it seems painless compared to GPG. Because it is new, it also has poor support on GitHub and other code forges, but I expect [that will change](https://github.com/github/feedback/discussions/7744#discussioncomment-1794438).

It may also seem premature to consider, given that [GPG signing isn't supported in gitui yet](https://github.com/extrawurst/gitui/issues/97); on the other hand, now may be a good time to consider the user interface to make it easy to support signing with SSH keys as well as GPG keys in the future.

skyfaller:
GitHub now supports verifying commits using SSH keys: https://github.blog/changelog/2022-08-23-ssh-commit-verification-now-supported/

extrawurst:
this feature needs to be supported by upstream `libgit2` first

dnaka91:
I did try to implement GPG signing in the past as part of #97, and it turned out to be very difficult, due to not having a native Rust crate for the GPG part. Basically, it complicated builds on Windows due to some extra C dependencies and licensing might be problematic too.

So I recently tried out to implement SSH signing instead, as I discovered the relatively new `ssh-key` crate from the RustCrypto group. After [a little forth and back](https://github.com/RustCrypto/SSH/issues/15) with them, they added an implementation for SSH signing to it, and with the current `0.5.0-rc.0`, it's very easy to sign commits.

The workflow is exactly the same as with GPG signing, so we wouldn't have to wait for `libgit2` to integrate that.  

Here is a very minimal example:

```toml
# Cargo.toml

[package]
name = "gitsign"
version = "0.1.0"
edition = "2021"

[dependencies]
anyhow = "1.0.66"
dirs = "4.0.0"
git2 = "0.15.0"
ssh-key = { version = "0.5.0-rc.0", features = ["ed25519", "encryption"] }
```

```rust
// src/main.rs

use std::{env, fs};

use anyhow::Result;
use git2::Repository;
[...]

PR #1544, opened 2023-02-13
Support git commit signing using OpenPGP

### LINKED ISSUE #97: GPG sign commits

hendrikmaus:
This Pull Request fixes/closes #97.

It changes the following:
- Add a new trait called `Sign`
- Implement `Sign` for `gpg` commit signing via shellout
  - Users with a working signing setup will find that it just works in gitui as well
- Add entrypoints for all natively supported methods
- Add potential entrypoint for pure rust implementations behind a gitconfig value of `gitui.signing_methods`

I followed the checklist:
- [x] I added unittests

Todo:
- [x] Test in Windows 🪟 
- [ ] ~Documentation~
- [ ] ~Draft an isolated integration test running in a container~

### Draft an isolated integration test running in a container

Since shellouts call out to actual system binaries, @extrawurst and I came up with the idea to create a test that runs in the isolation of a container. Inside of that environment, it could generate a GPG key and do an actual commit to verify the signature was created as expected.

**Update**: will be tracked in a follow-up issue.


hendrikmaus:
Will attend to the failures in CI within the next days.

ZCShou:
waiting for this feature ready

hendrikmaus:
Looking into the CI errors again now.

hendrikmaus:
The previous error was caused by a doctest I added. The other unit tests prepare their environment, the doc test does not. Since I added it as documentation only, I deactivated the doctest and the CI run in my fork was good now: https://github.com/hendrikmaus/gitui/actions/runs/4221076989

hendrikmaus:
The previous error was caused by a doctest I added. The other unit tests prepare their environment, the doc test does not. Since I added it as documentation only, I deactivated the doctest and the CI run in my fork was good now: https://github.com/hendrikmaus/gitui/actions/runs/4221076989

@extrawurst if you'd approve another CI run here, thanks.

melMass:
Hey just checking the status of this PR? 
I can test on macOS and windows easily if needed :)

hendrikmaus:
Both would be appreciated, thank you.

asyncgit/src/sync/sign.rs
```diff
@@ -0,0 +1,342 @@
...
+	/// let sign = SignBuilder::from_gitconfig(&repo, &config)?;
+	/// # Ok(())
+	/// # }
+	/// ```
+	pub fn from_gitconfig(
+		repo: &git2::Repository,
+		config: &git2::Config,
+	) -> Result<impl Sign, SignBuilderError> {
+		let signing_methods = config
+			.get_string("gitui.signing_methods")
```
hendrikmaus, line 99: @extrawurst this is the only spot where I invented a `gitui` config entry for the git config. I think we discussed to move that to the config files that `gitui` stores itself. Any preference on where to put it exactly? (exist file, new file, naming)

utkarshgupta137:
Tested it on macOS. Authoring commits works fine, but some simple operations like amending & rewording the last commit don't work. I've created a PR with workarounds: https://github.com/hendrikmaus/gitui/pull/1.

asyncgit/src/sync/sign.rs
extrawurst: checkout `options.rs` for `OptionsData` and its usage in various places

ZCShou:
> 👋 hello! Any movement on this by chance? Thanks!

I‘m also waiting for this feature

hendrikmaus:
Haven’t had time to continue, but will do so in the beginning of May. 

Will tackle these:
- Fix amending and rewording, thanks @utkarshgupta137 for already preparing a change-set
- Align configuration with the existing pattern
- Test on Windows
- Documentation
- Integration test

nrabulinski:
Hey, how's the progress on this? :)

hendrikmaus:
Can someone test a current build on a Windows environment? I'd appreciate that.

damccull:
To test this, would we need to download this PR's code, compile, and run? Or is there some auto-generated binary for testing I should download somewhere?

hendrikmaus:
At this point there are no pre-compiled binaries. So you have to download and compile it.

SirusCodes:
I compiled it with `cargo install gitui` and tried to run the binary from .cargo dir but it is showing me the same error. I might be doing something wrong...

PS: I don't have experience in working with Rust.

image

Sped0n:
works like a charm on ventura👌

just want this pr to be merged as soon as possible

image

Sped0n:
> I compiled it with `cargo install gitui` and tried to run the binary from .cargo dir but it is showing me the same error. I might be doing something wrong...
> 
> PS: I don't have experience in working with Rust.
> 
> image

maybe install with command below?

```shell
cargo install --git https://github.com/hendrikmaus/gitui --branch gpg-commit-signing
```

eclairevoyant:
@hendrikmaus works for me on windows 11 (MSVC toolchain, GnuPG as program for signing)

* Password prompt appears correctly when signing is enabled by default
* Commits are in fact signed (verified with `git log --show-signature`)

hendrikmaus:
Thank you for the feedback, that is good news.

hendrikmaus:
@extrawurst re configuration. 

As we were talking about this the other day, did you take another look at the available approaches? I.e. .gitconfig file versus the gitui options?

damccull:
Well, I installed it from the git repo and it definitely tries to sign now. However, I'm using my ssh key to sign rather than a gpg key and it says that's unsupported. Would this capability be difficult to add?

hendrikmaus:
@damccull out of scope here, but the change-set already prepares an entry point for that in the code. There is another open issue about signing with ssh keys in https://github.com/extrawurst/gitui/issues/1149

damccull:
Ah, thanks.

hendrikmaus:
> @extrawurst re configuration.
> 
> As we were talking about this the other day, did you take another look at the available approaches? I.e. .gitconfig file versus the gitui options?

I think it would be feasible to pass a new optional struct with the signing options down into the `asyncgit` crate from here: https://github.com/hendrikmaus/gitui/blob/24bb8ac57e2f7c8320eb4e6b75c8dc4739b6b7e0/src/components/commit.rs#L274

The `asyncgit::sync::commit` function has more than 40 call sites, mostly tests, that would need to be adjusted. The signatures of `asyncgit::sync::reword` and `asyncgit::sync::amend` would also be affected as they call `commit`.

I would propose something like `Option<CommitConfig>` to implement it open for extension. We might see the need to pass more configurations down to `commit` and I wouldn't want to limit it to only the signing options now. What do you think @extrawurst ?

legendary-cookie:
Any updates on this?

extrawurst:
an interesting catch: https://github.com/libgit2/libgit2/pull/6617 - seems like libgit2 starts supporting shell-out for ssh

h4rldev:
Can't wait for this, is there a workaround as of right now?

h4rldev:
> My current workaround is to use gitui for all operations it can do except commit and push. I quit it, use Connor via cli, then push real quick since I'm already there.

Ah, so what I've been doing currently, All I've really been doing is commiting which is quite easy, will probably dwelve further into gitui once this supports gpgsigning.

Scripter17:
I was able to compile this branch and have been using it for a while

I forget the exact commands but I assume some quick googling like "clone pull request from github" should help

extrawurst:
@hendrikmaus can you update this to the latest master? what exactly is still needed to push this over the finish line?

hendrikmaus:
Ack.

extrawurst:
@hendrikmaus can you please comment on https://github.com/extrawurst/gitui/pull/2047 - since it does not require shellout it would be important to me that we could get the abstraction aligned so that this PR becomes `just` another implementation (one that requires shellout)

blaggacao:
I'm exited! I just did the obvious resolution on the merge conflict built it from [there](https://github.com/blaggacao/gitui/tree/gpg-commit-signing) and it works like a charm. Adieu all the other tools!

hendrikmaus:
I resolved the conflict.

I was absent from this quite some time now due to other priorities.
Let's drive this home now.

hendrikmaus:
I will need to follow up on additional discussions that have been mentioned, but these are the remaining todo's specifically on this PR:

- [ ] Align configuration with the existing pattern
- [ ] Documentation
- [ ] Integration test

I think we got stuck on "Align configuration with the existing pattern" and I do not quite remember what we discussed @extrawurst. I think we can actually just get rid of that part for the time being and just get the `Sign` trait in and its impl. for GPG. If a user has GPG signing set up, it will seamlessly work in gitui as well.

extrawurst:
@hendrikmaus lets do that. this way we can move forward and set the right macherinery in place that allows implementing other signing methods too. the only important thing to please look over is: just making sure the way ssh singing is done in #2047 can be adopted to be "just" another implementation of your trait?

hendrikmaus:
I’ll have a look and verify that.

hendrikmaus:
Did not mean to close this. Was updating the PR.

extrawurst:
ok lets reopen :)

hendrikmaus:
We're back ;)

hendrikmaus:
Removed the testing files, as discussed, and will track them in a new issue later.

Now taking a look at the ssh signing PR to see if it could easily be refactored to use the `Sign` trait.

hendrikmaus:
The steps should roughly be:
- move the logic of [`fn fetch_ssh_key(config: &Config) -> Option<PrivateKey>`](https://github.com/extrawurst/gitui/pull/2047/files#diff-071ade48ffc93565e6e943f49b7ec30eb36067c1b0acf190d73b90ae1ae5d864R65-R96) to [`SignBuilder::from_gitconfig`](https://github.com/extrawurst/gitui/pull/1544/files#diff-cff35eb06935967639fae630f567c657957f172730981fc97be0152e5d5980acR148)'s match arm for ssh
- create smth like `pub struct SSHSign` and implement `Sign` for it using [`sk.sign`](https://github.com/extrawurst/gitui/pull/2047/files#diff-071ade48ffc93565e6e943f49b7ec30eb36067c1b0acf190d73b90ae1ae5d864R129-R131) which is all that is needed once the keypair was found
- the logic in the `commit` function does not need to change as it uses whatever the `SignBuilder` returns to sign the data. That is given the ssh signing also works on the commit's UTF-8 data
  - the `Sign` impl expects a UTF-8 encoded string slice, while the ssh impl seems to sign the buffer byte slice; will do a quick check to see I should change the trait to accept a byte slice as well

hendrikmaus:
According to git's documentation, commits are encouraged to be UTF-8, but any encoding can be used. So the raw bytes of the commit buffer should be signed rather than converting it to UTF-8 first.

Will implement the change in the `Sign` trait.

hendrikmaus:
Done.

hendrikmaus:
@extrawurst if you would approve the CI run, this is ready to be merged.

asyncgit/src/sync/commit.rs
```diff
@@ -82,16 +98,46 @@ pub fn commit(repo_path: &RepoPath, msg: &str) -> Result<CommitId> {
...
+		})?;
+
+		let sign = SignBuilder::from_gitconfig(&repo, &config)?;
+		let signed_commit = sign.sign(&buffer)?;
+		let commit_id =
+			repo.commit_signed(commit, &signed_commit, None)?;
+
+		// manually advance to the new commit ID
+		// repo.commit does that on its own, repo.commit_signed does not
+		repo.head()?.set_target(commit_id, msg)?;
```
extrawurst, line 126: this will probably break in an empty repo doing the first commit with signing, did you test this?

asyncgit/src/sync/commit.rs
```diff
@@ -82,16 +98,46 @@ pub fn commit(repo_path: &RepoPath, msg: &str) -> Result<CommitId> {
...
+		)?;
+
+		let commit = std::str::from_utf8(&buffer).map_err(|_e| {
+			SignError::Shellout("utf8 conversion error".to_string())
+		})?;
+
+		let sign = SignBuilder::from_gitconfig(&repo, &config)?;
+		let signed_commit = sign.sign(&buffer)?;
+		let commit_id =
+			repo.commit_signed(commit, &signed_commit, None)?;
```
extrawurst, line 122: if we want to support more than just gpg the third parameter `signature_field` needs to take something out of `sign` which in this case has to be `gpgsig` but something else for ssh

extrawurst:
another followup: this does not sign tags

asyncgit/src/sync/commit.rs
damccull: > if we want to support more than just gpg the third parameter `signature_field` needs to take something out of `sign` which in this case has to be `gpgsig` but something else for ssh

I very much wish for ssh signing. I don't PGP/GPG but i ssh all the time.

damccull: > if we want to support more than just gpg the third parameter `signature_field` needs to take something out of `sign` which in this case has to be `gpgsig` but something else for ssh

I very much wish for ssh signing. I don't PGP/GPG but i ssh all the time.

asyncgit/src/sync/commit.rs
MrPancakes39: About half a year ago approximately I tested out this version of `gitui` and it did indeed run into an error signing first commit in a repo. I would usually do first commit in the shell then open `gitui`. At least that's what I remember.

I haven't looked at the code so not sure if this line causes that issue but wanted to report my experience.

hendrikmaus: Yeah makes sense that is breaks as there isn't any HEAD at that moment. Will add code to default to the default branch name.

hendrikmaus: Fixed.

asyncgit/src/sync/commit.rs
hendrikmaus: Fixed.

@damccull this one only adds openpgp; there is another [PR](https://github.com/extrawurst/gitui/pull/2047/files#top) that wants to add ssh signing. @extrawurst wants them to adopt the `Sign` trait once this one here is merged though.

extrawurst:
wohooooo its merged! @hendrikmaus thanks for sticking with this through the end. could you please write an issue with the way you planned to do actual testing of this in CI via docker? also ideally pointing to the commits you reverted as inspiration, so that someone can take that up!

hendrikmaus:
> another followup: this does not sign tags

@extrawurst correct. it tags a signed commit just fine, but the tag itself isn't signed due to that. The way git signs tags is different from how it is cone for commits. I will open an issue with git2-rs to see if there is more information on it since I cannot seem to find anything in the docs that hints how to sign a tag.

I prepped the trait to differ between signing commits and tags, but I won't provide an implementation for tags with this PR.

PR #1547, opened 2023-02-15
Fix freeze on copy when xclip is installed on Linux

### LINKED ISSUE #1497: Copy hash freeze gitui
**Describe the bug**
When I copy the hash of a commit on gitui. It crashes.
But it copies the hash well.

**To Reproduce**
1. Open gitui 
2. go to log tab
3. Press y
4. and gitui freeze

 - Debian 11, kernel 5.10.0-20-amd64
 - gnome 3.38 (Wayland)
 - GitUI Version : 0.22.1
 - Rust version: 1.65.0
 - wl-clipboard : 2.0.0

louisbourque:
This Pull Request fixes/closes #1497 

It changes the following:
- It adds a boolean argument to determine if stderr should be piped, and sets it to false when running the xclip command.

I was able to reproduce the linked issue. If I run `killall xclip` then the UI resumes normal function. This issue started with e371153034db126235f342b52e2e6301e0880e70. Removing the `.stderr(Stdio::piped())` line fixes the freezing, but breaks the error reporting.

`xclip` waits for input on stdin. Something about piping stderr and stdin seems to cause the program to continue waiting for input. I tried different arguments to pass to xclip to tell it to exit, but nothing I tried seemed to work. I tried writing \x03 (End of Text) and \x04 (End of Transmission) and a few others that may have indicated to xclip to close. I tried passing `-l 1` (loop once then exit) which fixes the freezing, but results in the wrong thing being set on the clipboard. (It would always set the previous thing I copied, instead of the last)

In the end, passing a boolean to the function to indicate whether or not stderr should be piped seems to do the trick. If the command fails, we retry with `xsel` anyway, and report the error from that command. So the stderr output from `xclip` isn't needed.

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * crash on branches popup in small terminal ([#1470](https://github.com/extrawurst/gitui/issues/1470))
 * `edit` command duplication ([#1489](https://github.com/extrawurst/gitui/issues/1489))
 * syntax errors in `key_bindings.ron` will be logged ([#1491](https://github.com/extrawurst/gitui/issues/1491))
+* Fix UI freeze when copying with xclip installed on Linux ([#1497](https://github.com/extrawurst/gitui/issues/1497))
 * commit hooks report "command not found" on Windows with wsl2 installed ([#1528](https://github.com/extrawurst/gitui/issues/1528))
 
 ### Changed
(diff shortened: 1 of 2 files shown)


remlse:
Hello! Thanks for opening this. I'm having the exact same problem with `wl-copy` and your fix works with that as well:

```rust
if std::env::var("WAYLAND_DISPLAY").is_ok() {
	return exec_copy_with_args("wl-copy", &[], text, false);
}
```

extrawurst:
Thanks guys ❤️

PR #1566, opened 2023-02-27
Fix #1473 - correct msg limit in get_commits_intf

### LINKED ISSUE #1473: commit messages in the log page are sometimes missing
**Describe the bug**
Sometimes the commit messages are missing, see video below

**To Reproduce**
the steps to reproduce the error could not be clearly determined.
On some repos it functions correctly (showing the commit messages), on other repos it does not. The reason is not clear.
- use the `Log[2]` overview page notice the missing "commit messages"
- when pressing `2` the messages show up correctly

**Expected behavior**
The commit message should always show up.

**Screenshots**
The video first checks the commit messages on the [meiji163/gh-notify](https://github.com/meiji163/gh-notify) repo and then on [extrawurst/gitui](https://github.com/extrawurst/gitui) repo.

https://user-images.githubusercontent.com/92653266/207307694-16a01ea6-7a0b-4960-9921-931c5ca985e7.mp4

 - OS/Distro + Version: `macOS 10.15` (reproducible with default `Terminal` and `iTerm2`)
 - GitUI Version: `0.22.1`
 - Rust version: `stable 1.65.0 (bottled)` (installed via Homebrew)

**Additional context**
I tried to bisect the issue, but I ended up with the commit 9534e4c2f99111c928dcd2e91ca9fe49364a7ed5, that introduced the feature to save the last page the user viewed.

bisect

The problem may have been there for a while, but I never noticed it because I always had to press the `2` key to see the log page.

PS: Great tool you have made.

kting28:
When sync::get_commits_inf is called in revlog:::fetch_commits, the message limit used was the length of the list which can be zero at start up before items are fetched. This results in initial view of the Log view showing empty messages.

This Pull Request fixes/closes #{issue_num}.

It changes the following:
-
-

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/src/components/commitlist.rs
+++ b/src/components/commitlist.rs
@@ -88,9 +88,9 @@ impl CommitList {
 	}
 
 	///
-	pub fn current_size(&self) -> (u16, u16) {
-		self.current_size.get()
-	}
+	//pub fn current_size(&self) -> (u16, u16) {
+	//	self.current_size.get()
+	//}
 
 	///
 	pub fn set_count_total(&mut self, total: usize) {
--- a/src/tabs/revlog.rs
+++ b/src/tabs/revlog.rs
@@ -28,6 +28,7 @@ use tui::{
 };
 
 const SLICE_SIZE: usize = 1200;
+static MAX_MESSAGE_WIDTH: usize = 100;
 
 ///
 pub struct Revlog {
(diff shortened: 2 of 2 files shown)


kting28:
tentative fix, I don't understand the original code which use the list length as message limit.

LangLangBart:
:white_check_mark: after applying your patch, the problem described in the referenced issue is no longer reproducible.

extrawurst:
thanks for digging into this. your PR is kind of a hack. I created https://github.com/extrawurst/gitui/pull/1573 with a more stable solution I hope

kting28:
> thanks for digging into this. your PR is kind of a hack. I created #1573 with a more stable solution I hope

Awesome! FWIW, why is the message limit in "get_commits_info" needs to use current_size of the list? I thought the limit the horizontal limit of the draw screen. Cheers!

PR #1577, opened 2023-03-02
Fix dep struct patch

yanganto:
Sorry about was not fully compatible after merging others' feature PR in the crate. :pray: 

This Pull Request fixes the incorrect dependency after the incompatible one is yanked.

Due to revert the previous revert:
- [X] no additional unit tests are needed
- [X] no additional changelog is needed for refactoring


.github/workflows/ci.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        rust: [nightly, stable, '1.64']
+        rust: [nightly, stable, '1.67.1']
```
extrawurst, line 20: this is not warranted. Nija bumping the MSRV (which would require a bunch of other changes) as part of this I cannot accept

extrawurst:
> This branch cannot be rebased due to conflicts

PR #1578, opened 2023-03-02
ignore and remove flake.lock

yanganto:
remove the lock file not used in this project


extrawurst:
these changes are already in master. found PR too late

PR #1579, opened 2023-03-02
Fix external editor delay.

### LINKED ISSUE #1506: Delay when opening external editor
The delay before the external editor opens is bothering me, it's annoying at best and disruptive at worst.
It makes my normally snappy vim startup feel like emacs or some bloated, laggy electron-based IDE.

I read https://github.com/extrawurst/gitui/issues/157.
I don't know what can be done, maybe starting the editor in parallel with stopping input polling? 

In my opinion, adding loading spinners to operations that should take no time is really disappointing and not the kind of software I want to see or use in this day and age.

kosayoda:
The default polling rate of 1 second causes a 1 second delay when queuing the event to launch the external editor, causing borderline unusable latency.

However, a slower polling helps reduce CPU usage (not tested), so let's have a short polling duration as long as there are input events, and slow poll otherwise.

Since the external editor among other components (not tested) is always launched in response to an input event, we reduce the latency to ~100ms, which is the fast poll duration.

Fixes #1506.


kosayoda:
Potential questions:
- Do we still need the "opening external editor" popup if this PR fixes the delay? I tested not having the popup and it "feels faster" visually.
- Do we want a shorter "fast poll" duration? Personally I notice the latency drop when decreasing down to a 50ms poll duration but 100ms should be pretty snappy for most people.

extrawurst:
please rebase on current master. should fix the CI

kosayoda:
Hi! Just wondering if there's still anything to be done on my end.

PR #1616, opened 2023-03-21
chore(cargo): use crates.io sparse protocol

hendrikmaus:
It changes the following:
- Use crates.io sparse protocol
- Saves on compile time as downloading index is quicker
-https://blog.rust-lang.org/inside-rust/2023/01/30/cargo-sparse-protocol.html

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -6,3 +6,7 @@ linker = "arm-linux-gnueabihf-gcc"
 
 [target.armv7-unknown-linux-gnueabihf]
 linker = "arm-linux-gnueabihf-gcc"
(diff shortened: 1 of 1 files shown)


hendrikmaus:
Unfortunately there seems to be a problem in the macOS workaround steps.
Screenshot 2023-03-21 at 09 29 24

https://github.com/hendrikmaus/gitui/actions/runs/4476833175/jobs/7867621383

hendrikmaus:
Since this was recently merged, the sparse protocol will be the new default soon anyway. Will close this effort.

Issue #1316, opened 2022-09-07
Vim is not available as external editor on windows


ShrykeWindgrace:
Thanks for `gitui`, it makes my daily life easier=)

There is, however, one thing that seems break my current flow - I can not launch vim from within gitui.

**Describe the bug**
Vim is not available as external editor

**To Reproduce**
1. Navigate to a repo with modified files, run `gitui`
2. Chose a file with modifications
3. Hit 'e' button to edit this file
4. A "popup" appears:
```
failed to launch editor:
"vim": program not found
```
Same bug observed when trying to '^e'dit commit message in external editor.

**Expected behavior**
A `vim` instance with selected file loaded.

**Context (please complete the following information):**
 - OS/Distro + Version: win10 (build 21H2-19044.1889)
 - GitUI Version 0.21.0
 - Running in `WindowsTerminal` and `pwsh-7.2.6`
 - both `vim` and `gitui` are installed via `chocolatey`

**Additional context**
`vim` is present in path:
```
PS1> where.exe vim
C:\Windows\vim.bat
```
```console
git config  -l | rg editor
core.editor=vim
```

extrawurst:
need windows user support for this

ShrykeWindgrace:
I did a bit of digging
- "real" `vim` in my `$PATH` is, in fact, a `bat` script:
```console
PS> (Get-Command vim).Name
vim.bat
```
- there is a `%PATHEXT%` (`$env:PATHEXT` in `pwsh`) env variable that specifies the list of extensions to test if the command's extension is not provided. (see also https://learn.microsoft.com/en-us/previous-versions/cc723564(v=technet.10)#command-search-sequence and https://stackoverflow.com/a/1653492/3206165)
- For some reason the implementation in rust (I suppose the guilty party should be somewhere around `externaleditor.rs:115`) does not do these extension checks and only tries to run the provided executable as-is.

I see the following solutions:
- (workaround) explicitly provide extension; might lead to bugs if one uses `git`'s config in a cross-platform (or cross-installation) manner (that's my case, unfortunately)
- fix behavior of the library responsible for running external commands; that would be waaaay out of my depth, I can't write a helloworld in `rust`
- bikeshed a test in `gitui` "if the editor command has no extension and we are on windows, then try with the extensions form that env var..."
- some other option I might have missed?

ShrykeWindgrace:
Aha, it appears that things are not *that* easy upstream...

https://github.com/rust-lang/rust/issues/37519

[...]

PR #1641, opened 2023-04-10
make vim work

pm100:
This Pull Request fixes/closes https://github.com/extrawurst/gitui/issues/1316

on windows if the launch of the editor fails it tries to run it as a batch file. It executes

`cmd /C {editor setting} {file name}`


pm100:
i forgot to run clippy

PR #1646, opened 2023-04-11
fix test failure on nightly due to cef81dc

pm100:
simple fix in the end. Restores the behaviour - 

which is that 

        h.insert("\n", "\u{21b5}\n\r"); // old

	h.insert("\n", "\u{21b5}\r\n\n");  // new

this substitution string gets parsed as

    `arrow` and `""`

The new one works on stable and nightly

### DIFF
--- a/src/components/textinput.rs
+++ b/src/components/textinput.rs
@@ -251,7 +251,7 @@ impl TextInputComponent {
 
 		let cursor_highlighting = {
 			let mut h = HashMap::with_capacity(2);
-			h.insert("\n", "\u{21b5}\n\r");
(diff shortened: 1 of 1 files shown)


cruessler:
I was wondering whether we need the `\r` at all. Could it also be just removed from the new string?

pm100:
test passes with or without the \r. I left it there because somebody felt it was needed in the first place, maybe there is a non test code path that I cant see but that needs it, I dont think so, but i am not 100% sure

Seems that we need to fix this rather than fine tuning (given that all PRs are failing)

extrawurst:
Thanks for looking into this ❤️

PR #1648, opened 2023-04-12
Fix file history for sizes <= 1200 entries

### LINKED ISSUE #1560: wrong file history view
repro: 
* go to a commit
* select a file
* open file history
* select initial commit
* open commit details
* go back `close`
* only one entry shown in history

cruessler:
This Pull Request partly fixes #1560.

There is a lot of context in the comment I included in the change. As far as I can tell, this fix works for histories that have <= 1200 entries. If we want a solution that addresses the issue for larger histories as well, we can hopefully take this PR as a starting point. For the time being, I thought it would be best to open the PR in this state already, so we have more context and can start a discussion.

It changes the following:

- It brings `self.items.index_offset` always in sync with `self.table_state` by always using `0` as the items’ offset

I followed the checklist:

- [ ] I added unittests

### DIFF
--- a/src/components/file_revlog.rs
+++ b/src/components/file_revlog.rs
@@ -226,16 +226,40 @@ impl FileRevlogComponent {
 		if let Some(git_log) = &mut self.git_log {
 			let table_state = self.table_state.take();
 
-			let start = table_state.selected().unwrap_or(0);
-
 			let commits = get_commits_info(
 				&self.repo_path.borrow(),
-				&git_log.get_slice(start, SLICE_SIZE)?,
+				&git_log.get_slice(0, SLICE_SIZE)?,
 				self.current_width.get(),
 			);
 
 			if let Ok(commits) = commits {
-				self.items.set_items(start, commits);
+				// 2023-04-12
+				//
+				// There is an issue with how windowing works in `self.items` and
+				// `self.table_state`. Because of that issue, we currently have to pass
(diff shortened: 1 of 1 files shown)


cruessler:
Yes, the file history’s entries are loaded here: https://github.com/cruessler/gitui/blob/e40e095993dc2c7c652296a38cdb9ba2b487b7f4/src/components/file_revlog.rs#L231. The code reads `&git_log.get_slice(0, SLICE_SIZE)?`, so this will always load the first 1200 items, never items beyond that limit.

What we could also do is maybe introduce padding in `get_rows`, i. e. we could prepend the correct number of empty rows so that the offset used by `render_stateful_widget` would be valid. Downside: added complexity and a few bytes of additional memory used.

cruessler:
Since I wrote the initial code in `file_revlog.rs`, I can add a bit of context. :-) I copied the batching from `revlog.rs`, but I decided to use `tui`’s `Table` for rendering instead of `CommitList` (the latter is what `Revlog` uses). At that time, I was not aware that this would cause issues with batching.

From what I can tell, https://github.com/tui-rs-revival/ratatui/pull/12 would resolve the issue because it would give us access to `self.table_state`’s `offset` property which we need to adjust, but currently can’t.

extrawurst:
Ok I will merge this then. Let’s create a followup ticket to not forget about this Limitation and link it to the above ratatui issue  💪

PR #1650, opened 2023-04-12
make XDG_XXX_HOME work for Windows

### LINKED ISSUE #1498: Follow XDG spec on windows
I use windows and have set XDG_CONFIG dirs (XDG_CONFIG_HOME, XDG_CACHE_HOME) but gitui isn't honoring them

If XDG_CONFIG_HOME is set on any os, use XDG_CONFIG_HOME/gitui as directory for config file.   If it is not set, fall back to os default

pm100:
This Pull Request fixes/closes #1498.

- If XDG_CONFIG_HOME is set on Windows then use it instead of what `dirs_next` says
- if XDG_CACHE_HOME is set on Windows then use it instead of what `dirs_next` says

Note that `dirs_next` already honors XDG_XXX_HOME on Linux

It is arguable that not doing this for Mac is odd, but that not whats asked for in the bug.

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -145,9 +145,9 @@ fn get_app_cache_path() -> Result<PathBuf> {
 		env::var("XDG_CACHE_HOME")
 			.ok()
 			.map(PathBuf::from)
-			.or_else(|| dirs_next::config_dir())
+			.or_else(|| dirs_next::cache_dir())
 	} else {
-		dirs_next::config_dir()
+		dirs_next::cache_dir()
 	}
 	.ok_or_else(|| anyhow!("failed to find os cache dir."))?;
 
--- a/src/args.rs
+++ b/src/args.rs
@@ -145,7 +145,7 @@ fn get_app_cache_path() -> Result<PathBuf> {
 		env::var("XDG_CACHE_HOME")
 			.ok()
 			.map(PathBuf::from)
-			.or_else(|| dirs_next::cache_dir())
+			.or_else(dirs_next::cache_dir)
 	} else {
 		dirs_next::cache_dir()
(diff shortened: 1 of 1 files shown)


extrawurst:
Can we try and be good citizens and file this upstream and try making it a PR there?

pm100:
I am not sure what you are saying. 

Are you responding to my Macos comment, do you want me to add the same feature for Mac. In this PR? Or do you want a new issue for Mac to support XDG. In the past I got shot down here for doing more than was precisely asked in an issue.

Or did I do something wrong with this PR in terms of how it intersects with branches etc (I see the word 'upstream', which I am just now understanding). I am very limited in my git/github/PR experience and could have made a mistake.

extrawurst:
I m saying if you think it should be a Feature of `dirs_next` let’s try to add it as a PR over there (upstream) first before, WDYT?

pm100:
@extrawurst OK I understand.

No I dont thing that dirs_next should do it. Its a linux thing. Do what they do is fine.

Note that git explicitly describes the support of XDG, without qualifying it by platform. https://git-scm.com/docs/git-config#FILES. So we should too

I have verified that libgit2 obeys XDG cross platform. So in fact we do honor it. The code change I made makes it work for 2 things that we implement without libgit2, log file location and keyboard map location. And I do think we should do that for Mac too

However I am now surprised about the original issue. Its a blanket statement about not supporting XDG on windows (but with no mention of those 2 items, log and keyboard). I have asked the original poster to clarify, because maybe there is something else going on.

I am going to close this PR while awaiting the response. And I will add Mac support too

Issue #1654, opened 2023-04-14
External editor does not start with EDITOR set


darthmall:
**Describe the bug**
I have set `EDITOR=nvim` in my shell, but when I hit ctrl-e, nothing happens.

**To Reproduce**
1. Stage changes
2. Press 'w' to switch to staged changes
3. Press 'ctrl+e'
4. Nothing happens

**Expected behavior**
NeoVim should start up with the commit message template.

**Context (please complete the following information):**
 - OS/Distro + Version: Pop!_OS 22.04 LTS x86_64
 - GitUI Version: 0.22.1
 - Rust version: 1.67.1

**Additional context**
I’m using fish version 3.6.1 for my shell, and I have NeoVim version 0.8.2. I have tried running `EDITOR=vi gitui` and ctrl-e still does nothing.

pm100:
^e is only available in the commit popup. 'h' shows help

![image](https://user-images.githubusercontent.com/489231/232552319-0647f191-9eed-4f8d-ae15-8be62583b427.png)

in the staged change box, press 'c' to start a commit. You will get

![image](https://user-images.githubusercontent.com/489231/232552479-1667d5d6-a77b-4c1d-ba49-6b7002a22917.png)

now hit ^e

![image](https://user-images.githubusercontent.com/489231/232552903-a55cfd57-595f-4915-9823-6e9cbe72bd16.png)

darthmall:
Thanks, I misunderstood the help menu in the app. But now I’m getting an error:

```
Failed to launch editor:
"$EDITOR": no such file or directory (os error 2)
```

I’ve tried the following:

1. `EDITOR=nvim gitui`
2. `EDITOR=vi gitui`
3. `EDITOR=/home/evan/.local/bin/nvim gitui`

I have no trouble launching either `vi` or `nvim` from the same shell in which I then launch `gitui`.

darthmall:
> BTW the precedence is GIT_EDITOR, core.editor, VISUAL, EDITOR

Ok, that’s what it was. I had a bad value for `core.editor` in `~/.gitconfig`.

But for what it’s worth, I was not passing `gitui` as an argument to `nvim`. The commands I executed are:

1. `EDITOR=nvim gitui`
2. `EDITOR=vi gitui`
3. `EDITOR=/home/evan/.local/bin/nvim gitui`

[...]

PR #1656, opened 2023-04-17
Make .bat and .cmd editors work on windows

### LINKED ISSUE #1316: Vim is not available as external editor on windows

pm100:
This Pull Request fixes/closes #1316.

This has several parts
- first it adds code to enable .bat and .cmd commands to be invoked without the .bat or .cmd being given (core.editor='vim' will now run vim.bat)
- it then adds unit tests
- it also reorganized some code used for testing in asyncgit so that it can be used in the main, gitui, product

The actual fix is fairly straightforward.
- if on windows and the configured editor fails to launch then retry
- retry using `cmd /C <editor> <file>`
- if that fails to start or cmd reports an error then report that to the user

The reorg needs an explanation
- open_file_in_editor need a repo. 
- there is test plumbing in asyncgit that provides many repo utilities
- sadly things declared 'test' are not available in other products / subcrates. But there is a solution. see https://github.com/rust-lang/cargo/issues/8379. It involves creating a test-utils feature and making that a dev dependency.
- That now makes that test plumbing in asyncgit/sync/mod.rs no longer hidden under a test umbrella. The exhaustive cliuppy, lints, cargo-sort etc now picks up all sorts of issues there (that really dont matter) so now I had to add in a few things to stop those being picked up. Note that they are still only being compiled when doing tests, its just that clippy 'all-features' doesnt realize that

I also note that there are no other external editor test cases. I will add more once I get this in (I willl look at #1654 and put them in the fix for that). I did add one generic test, but there should be more 

I followed the checklist:
- [ x] I added unittests
- [x ] I ran `make check` without errors
- [x ] I tested the overall application

### DIFF
--- a/asyncgit/src/lib.rs
+++ b/asyncgit/src/lib.rs
@@ -23,7 +23,8 @@
 // #![deny(clippy::expect_used)]
 //TODO: consider cleaning some up and allow specific places
 #![allow(clippy::significant_drop_tightening)]
-#![allow(clippy::multiple_crate_versions)]
+
+//#![allow(clippy::multiple_crate_versions)]
 
 pub mod asyncjob;
 mod blame;
--- a/asyncgit/src/lib.rs
+++ b/asyncgit/src/lib.rs
@@ -23,8 +23,7 @@
 // #![deny(clippy::expect_used)]
 //TODO: consider cleaning some up and allow specific places
 #![allow(clippy::significant_drop_tightening)]
-
-//#![allow(clippy::multiple_crate_versions)]
+#![allow(clippy::multiple_crate_versions)]
 
 pub mod asyncjob;
 mod blame;
--- a/asyncgit/src/lib.rs
+++ b/asyncgit/src/lib.rs
@@ -23,6 +23,7 @@
 // #![deny(clippy::expect_used)]
 //TODO: consider cleaning some up and allow specific places
 #![allow(clippy::significant_drop_tightening)]
+#![allow(clippy::multiple_crate_versions)]
 
 pub mod asyncjob;
 mod blame;
--- a/deny.toml
+++ b/deny.toml
@@ -22,5 +22,6 @@ version = "1.0.3"
 multiple-versions = "deny"
 skip-tree = [
     { name = "windows-sys" },
-    { name = "hermit-abi" }
+    { name = "hermit-abi" },
+    { name = "syn" }
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -740,9 +740,9 @@ dependencies = [
 
 [[package]]
 name = "git2"
-version = "0.16.1"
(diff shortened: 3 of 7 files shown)


asyncgit/src/lib.rs
```diff
@@ -23,6 +23,7 @@
 // #![deny(clippy::expect_used)]
 //TODO: consider cleaning some up and allow specific places
 #![allow(clippy::significant_drop_tightening)]
+#![allow(clippy::multiple_crate_versions)]
```
extrawurst, line 26: this is a red-flag. why do we want that? it bloats the binary. did you investigate what introduces the multiple versions of the same crate?

deny.toml
```diff
@@ -22,5 +22,6 @@ version = "1.0.3"
 multiple-versions = "deny"
 skip-tree = [
     { name = "windows-sys" },
-    { name = "hermit-abi" }
+    { name = "hermit-abi" },
+    { name = "syn" }
```
extrawurst, line 26: see above

pm100:
Multiple versions do not get built into the product, they only happen on test builds

`cargo clippy --workspace --all-features` complains about the code that is in the test feature 'test_utils'. It does not realize that this is only for testing

Same for cargo deny

I tried real hard to untangle the multiple versions even though it only really happens with tests. I could not untangle some of them (people who version  their crate 0.xx are very annoying because 0.45 is not seen as a valid replacement for 0.44 by cargo. Yes windows-sys we are looking at you)

Then I saw that you already had the ignore dups in main.rs and in deny.toml and decided OK its not just me that could not un ravel these

TL;DR - the dups only happen when running tests

pm100:
I had to do a quick commit with that allow taken out because I could not remember what actually failed, I knew it was part of the CI build

pm100:
I just discovered squash. I apologize for my messy prs. I will clean this up

PR #1657, opened 2023-04-17
Default to tick-based updates

### LINKED ISSUE #1444: remove polling watcher
notify based PollWatcher does a ton of caching and can be expensive in giant repos. the new `--polling` arg should sidestep using notify entirely and go back to a regular status polling

cruessler:
This Pull Request fixes/closes #1444.

It changes the following:

- This commit reintroduces code that was previously removed in favor of a notify-based update trigger. It turned out that notify-based updates can cause issues in larger repositories, so tick-based updates seemed like a safer default.

PR that introduced the notify-based watcher: https://github.com/extrawurst/gitui/pull/1310

I followed the checklist:

- [ ] I added unittests

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -15,6 +15,7 @@ use std::{
 pub struct CliArgs {
 	pub theme: PathBuf,
 	pub repo_path: RepoPath,
+	pub notify_watcher: bool,
 	pub poll_watcher: bool,
 }
 
@@ -54,11 +55,14 @@ pub fn process_cmdline() -> Result<CliArgs> {
 		get_app_config_path()?.join("theme.ron")
 	};
 
+	let notify_watcher: bool =
+		*arg_matches.get_one("watcher").unwrap_or(&false);
 	let arg_poll: bool =
 		*arg_matches.get_one("poll").unwrap_or(&false);
 
 	Ok(CliArgs {
 		theme,
+		notify_watcher,
 		poll_watcher: arg_poll,
 		repo_path,
 	})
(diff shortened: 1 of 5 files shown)


src/args.rs
```diff
@@ -15,6 +15,7 @@ use std::{
 pub struct CliArgs {
 	pub theme: PathBuf,
 	pub repo_path: RepoPath,
+	pub notify_watcher: bool,
 	pub poll_watcher: bool,
```
extrawurst, line 19: `poll_watcher` should be deprecated by this, right? because it is the default

src/args.rs
```diff
@@ -95,6 +99,12 @@ fn app() -> ClapApp {
 				.long("logging")
 				.num_args(0),
 		)
+		.arg(
+			Arg::new("watcher")
+				.help("Use notify-based file system watcher instead of tick-based update. This is more performant, but can cause issues in larger repositories")
```
extrawurst, line 104: maybe we can add a entry in `FAQ` to sum up the reasonings behind this and link to the FAQ here

extrawurst (commented):
Thanks for taking this one ❤️

src/main.rs
```diff
@@ -83,11 +83,13 @@ use tui::{
 use ui::style::Theme;
 use watcher::RepoWatcher;
 
+static TICK_INTERVAL: Duration = Duration::from_secs(5);
```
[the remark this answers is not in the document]
cruessler: Yes, I got this line from #1310.

src/main.rs
```diff
@@ -114,6 +116,13 @@ pub enum AsyncNotification {
 	Git(AsyncGitNotification),
 }
 
+#[derive(Clone, Copy, PartialEq)]
+enum Updater {
+	Ticker,
+	PollWatcher,
```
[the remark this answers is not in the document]
cruessler: No, if it’s causing issues that are not fixable or unlikely to be fixed, it makes more sense to completely remove it.

src/args.rs
cruessler: I added an entry to `FAQ.md` and referenced it here.

src/args.rs
cruessler: I followed your suggestion below and removed `poll_watcher`.

After the removal, `clap` prints an error when `--poll` is passed as an argument: `error: unexpected argument '--poll' found`.

Issue #1350, opened 2022-09-21
Search branch


Wyfy0107:
**Is your feature request related to a problem? Please describe.**
When viewing all the branches in gitui, I would like to have the ability to search through these branches

**Describe the solution you'd like**
Perhaps in the branch view, press `/` to start searching, similar to vim

**Describe alternatives you've considered**

**Additional context**

extrawurst:
would be nice to reuse the fuzzy search popup we have for this already (see the file tab)

Wyfy0107:
Haven't used rust yet, but can I try to add this feature ?

ObiWanRohan:
I'm trying my hand at this.

ObiWanRohan:
@extrawurst I have added a key ("Branch Find Toggle") which would toggle the focus to the find box in the branch list popup.
Pressing the key (Currently `/` by default) would shift the focus to the text input.
I didn't go the route of making it similar to the file finder, as there are no other possible inputs in the file finder. Here there are possible inputs like `c` for create branch, `r` for rename and so on.
Otherwise there would have to be a completely separate finder popup for the branches (this is also an option).

Some feature questions come to mind:
- What should be the scope of the finder? (Just the open tab or both tabs). The open tab is easier to implement as it needs lesser changes. 
- Should the matches be highlighted or should only the matches be shown?
- How does the user exit the find mode? (The Escape key or backspacing the whole input)

I'll add more as they come.

ObiWanRohan:
@extrawurst Pinging for answers

PR #1658, opened 2023-04-18
add fuzzy finder in branch list

UUGTech:
This PR is related to #1350.

Hi @extrawurst,

Sorry for my poor English. But I have a small PR!
I implemented a fuzzy finder pop-up in the branch list.
This may make it easier to find the branch you want to select.

Instead of a pop-up, we could implement as an embedded finder in the branch list, but I don't think this is a good idea as it would change the look and feel a lot.
So I have implemented it as a pop-up.
The implementation is almost the same as file_find_popup.

It changes the following:
- add branch finder popup to branchlist component

Here is how it works in the implementation in this PR.
gitui_branch_find_popup

I followed the checklist:
- [ ] I added unittests


src/keys/key_list.rs
```diff
@@ -191,6 +192,7 @@ impl Default for KeysList {
 			abort_merge: GituiKeyEvent::new(KeyCode::Char('A'),  KeyModifiers::SHIFT),
 			open_file_tree: GituiKeyEvent::new(KeyCode::Char('F'),  KeyModifiers::SHIFT),
 			file_find: GituiKeyEvent::new(KeyCode::Char('f'),  KeyModifiers::empty()),
+			branch_find: GituiKeyEvent::new(KeyCode::Char('F'),  KeyModifiers::SHIFT),
```
extrawurst, line 195: why did you choose a different hotkey to the file finder?

src/strings.rs
```diff
@@ -693,6 +703,17 @@ pub mod commands {
 			CMD_GROUP_GENERAL,
 		)
 	}
+	pub fn scroll_popup(key_config: &SharedKeyConfig) -> CommandText {
```
extrawurst, line 706: why is there a new one needed here? did we not have that in other places for `up` and `down` already?

src/components/branch_find_popup.rs
```diff
@@ -0,0 +1,340 @@
...
+use std::borrow::Cow;
+use tui::{
+	backend::Backend,
+	layout::{Constraint, Direction, Layout, Margin, Rect},
+	text::{Span, Spans},
+	widgets::{Block, Borders, Clear},
+	Frame,
+};
+
+pub struct BranchFindPopup {
```
extrawurst, line 24: this looks like a lot of copy paste from `FileFindPopup`. can we extract a shared base here? we might want the same for other places too

src/components/branch_find_popup.rs
```diff
@@ -0,0 +1,340 @@
...
+					items,
+				);
+			}
+		}
+		Ok(())
+	}
+}
+
+impl Component for BranchFindPopup {
+	fn commands(
```
extrawurst, line 270: we should also add an entry for the `return` command that closes this finder based on the selection (technically the same as `esc` because we already selected the entry in the background)

extrawurst (changes requested):

src/keys/key_list.rs
UUGTech: In remote tab of branchlist component, `f` is already mapped for `fetch`. So I needed to make it different. 
If there is another good hotkey, we can choose it.

src/components/branch_find_popup.rs
UUGTech: Yes, the process is almost the same as `FileFindPopup`, so I will try to make one like a `FuzzyFindPopup` for general use.
Actually, I'm very new to Rust, so it might take a bit of time.

src/components/branch_find_popup.rs
UUGTech: Sorry, I forgot to add command info of `return`. I think it already could be closed by `return`. `Esc` also closes based on the selection. This is the same as `file_find_popup`. By the way, command info of `file_find_popup` also dose not have `return`, so I think we need to add it too.

src/components/branch_find_popup.rs
extrawurst: if you commit to following up with that I am fine if we make this a separate issue+PR. if you are up for that please create the issue to mark this as intended to not be part of this PR

src/components/branch_find_popup.rs
extrawurst: yeah lets add it in both places. exactly why we need to consolidate both into a common component

UUGTech:
Ok, when it's ready for the final review, I re-request review.

Before that, I have a small question.
Which is better command info?

`Scroll[...] Close[⎋] Close[⏎]`
`Scroll[...] Close[⎋⏎]`

extrawurst:
Thanks a lot!

extrawurst:
added changelog entry for you including gif showcase:

UUGTech:
I had completely forgotten about the changelog. 🙇‍♂️
Thank you so much!

PR #1686, opened 2023-05-11
fix double key input on windows due to crossterm 0.26

pm100:
crossterm 0.26 adds key release event on windows, this causes all key inputs to be duplicated in apps that dont explciitly look at the key event type

It changes the following:
- in input.rs is not key press then skip input

I followed the checklist:
- [ ] I added unittests


pm100:
notw that the CI is failing due to the 1.65 shift requirement

PR #1702, opened 2023-05-28
Indicate to the user that a stash save is in progress

### LINKED ISSUE #1495: Add a visual cue to indicate that stashing is in progress
When stashing it can take several seconds before the stash completes and the popup closes, but there is no visual cue that it is happening. I have on occasion wondered whether I actually pressed Enter and have pressed it again; fortunately, this seems to cause no harm.

In the Stash popup, overwrite the "type name (optional)" message or the name (if typed) with a message like "Stashing..." as soon as Enter is pressed. This will inform the user that something is actually happening.

nrempel:
This Pull Request fixes #1495.

It changes the following:
- Add AsyncStash
- Updates `StashMsgComponent` to use AsyncStash instead of sync::Stash
- Introduces disabled flag on `TextInputComponent`
- Change `StashMsgComponent` title to "Stashing..." while stash is pending and disable input

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/src/components/textinput.rs
+++ b/src/components/textinput.rs
@@ -31,11 +31,13 @@ pub enum InputType {
 }
 
 /// primarily a subcomponet for user input of text (used in `CommitComponent`)
+#[allow(clippy::struct_excessive_bools)]
 pub struct TextInputComponent {
 	title: String,
 	default_msg: String,
 	msg: String,
 	visible: bool,
+	disabled: bool,
 	show_char_count: bool,
 	theme: SharedTheme,
 	key_config: SharedKeyConfig,
@@ -57,6 +59,7 @@ impl TextInputComponent {
 		Self {
 			msg: String::new(),
 			visible: false,
+			disabled: false,
 			theme,
 			key_config,
 			show_char_count,
@@ -220,6 +223,16 @@ impl TextInputComponent {
 	///
 	pub fn set_default_msg(&mut self, v: String) {
 		self.default_msg = v;
+	}
+
+	///
+	pub fn disable(&mut self) {
+		self.disabled = true;
+	}
+
+	///
+	pub fn enable(&mut self) {
+		self.disabled = false;
 	}
 
 	fn get_draw_text(&self) -> Text {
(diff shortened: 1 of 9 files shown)


src/components/textinput.rs
```diff
@@ -31,11 +31,13 @@ pub enum InputType {
 }
 
 /// primarily a subcomponet for user input of text (used in `CommitComponent`)
+#[allow(clippy::struct_excessive_bools)]
```
nrempel, line 34: Note: I saw this elsewhere so I also allowed this rule. Refactoring this not to include 3 bools didn't make sense to me.

src/components/textinput.rs
```diff
@@ -419,7 +432,7 @@ impl Component for TextInputComponent {
 	}
 
 	fn event(&mut self, ev: &Event) -> Result<EventState> {
-		if self.visible {
+		if self.visible && !self.disabled {
```
extrawurst, line 435: this is pretty invasive. why do we need that? the text input for the stashing can just close IMO and does not need to learn a new state of `disabled`. what is the rational?

extrawurst (changes requested):

xcloudplatform:
yes we need that

src/components/textinput.rs
extrawurst: @nrempel

nrempel: Without disabled, it's possible to type in the field while the operation is pending.

I suppose we could close the model and continue the operation in the background. In this case, is there a pattern for handling stashing errors if they occur?

extrawurst: yeah that's what I think we should do, what we do in other cases already. just close the text input popup regularly and then the new popup will block any other input and visualise that the stashing is in progress

extrawurst:
@nrempel are you willing to finish this up?

nrempel:
Sorry, it's been a very busy year. I may have time to finish this over the holiday break. But I'll close this for now and revive it if needed.

PR #1738, opened 2023-06-28
Fix file history for all sizes

cruessler:
This Pull Request is a follow-up to #1648 and related to #1560.

It changes the following:

- It uses `offset()`, now exposed by `ratatui`, to keep `items`’ offset in sync with `table_state`’s offset.

I followed the checklist:

- [ ] I added unittests

### DIFF
--- a/src/components/file_revlog.rs
+++ b/src/components/file_revlog.rs
@@ -354,8 +354,14 @@ impl FileRevlogComponent {
 	}
 
 	fn set_selection(&mut self, selection: usize) {
-		let height_in_items = self.current_height.get() / 2;
-		let new_offset = selection.saturating_sub(height_in_items);
+		let height_in_items =
+			(self.current_height.get().saturating_sub(2)) / 2;
+
+		let offset = *self.table_state.get_mut().offset_mut();
+		let min_offset = selection
+			.saturating_sub(height_in_items.saturating_sub(1));
+
+		let new_offset = offset.clamp(min_offset, selection);
 
 		*self.table_state.get_mut().offset_mut() = new_offset;
 		self.table_state.get_mut().select(Some(selection));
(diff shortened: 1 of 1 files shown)


extrawurst:
thanks for tackling this. 
I see a different behaviour now where the first view of file history does not show any entry selected and it only shows once you go up or down once

cruessler:
The issue turned out to be more complex than I initially thought. I overhauled how selection as well as data fetching and refetching work in `FileRevlogComponent`. So far, I’ve found no issues. Behavior I tested:

- Opening the history for a file, scrolling up and down, opening commits and returning to the history view
- Closing the history for one file, opening it for another one, then repeating the tests done for the first file

extrawurst:
there seems something off:
Screenflick Movie 10

extrawurst:
@cruessler any updates?

cruessler:
I haven’t gotten to look at the PR again yet. If I don’t get to it this week, then I’ll find time end of next week, I think.

cruessler:
@extrawurst I think I found the issue. I rebased and manually tested the branch again, including the test case you provided.

extrawurst:
now the scrolling works but the diffs still do not update:

https://github.com/extrawurst/gitui/assets/776816/d9d10b44-23d1-425c-a721-15378d5b1360

extrawurst:
Visually it works now!

I am not sure if this implementation is working as intended though. I added a `log::info` in `fetch_commits` (line 221):

```
log::info!("revlog: fetch commits: {offset}");
```

if I reduce the batch size to force it to have to fetch commits (200 batch size) this shows that every `arrow down` leads to a new `fetch_commits` as soon as the `batchsize-elements in view` (or something else ^^):

```
revlog: fetch commits: 174
revlog: fetch commits: 175
...
```

this is not as efficient as it could be if we would fetch commit infos in batches and then only again once we run out.

side note: the selection should go up and down inside the view before the scroll position changes to keep the selection in view. right now if you scroll down it will stick to the bottom and the scroll position changes when scrolling up

extrawurst:
ok comparing with current master we should fix the scrolling up/down issue because on master it does it correctly with the scrolling window moving only when the selection reaches the upper/bottom border

cruessler:
I updated the PR, tested it quite a bit, hoping not to have missed anything, and ran `make check`. That took slightly longer than expected, so I’ll take a walk now. :-)

PR #1765, opened 2023-07-20
Switch from textwrap to bwrap

### LINKED ISSUE #1762: switch to bwrap
bwrap seems a better fit for the job of wrapping text: https://github.com/imichael2e2/bwrap

TheBlackSheep3:
This Pull Request fixes/closes #1762.

It changes the following:
- removed dependency textwrap
- added dependency bwrap 
- changed commit detail wrapping to work with bwrap api

I followed the checklist:
- [x] ~I added unittests~ The functionality was verfied by existing unittests


Cargo.toml
AmmarAbouZor, line 1: I think the feature `use_std` should be added here since gitui only runs with std support.

Letting the standard library manage memory is much safer option than external crates

src/components/commit_details/details.rs
```diff
@@ -115,10 +142,26 @@ impl DetailsComponent {
...
-		wrapped_title.len() + wrapped_message.len()
+		if let Some(ref data) = details {
+			if let Some(ref message) = data.message {
+				let title_len = message.subject.len();
+				let mut title_buffer =
+					vec![0; title_len + title_len / width + 1];
+				let message_len =
+					message.body.as_ref().map_or(0usize, String::len);
+				let mut message_buffer =
+					vec![0; message_len + message_len / width + 1];
```
AmmarAbouZor, line 153: Buffers length assumes that all the characters are ASCII so each char will occupy one byte only. But that's not the case and this will fail with non-ASCII chars since chars in UTF8 could occupy ab to 4 bytes. For such cases you could use `as_bytes()` method to get the actual number of bytes.

src/components/commit_details/details.rs
```diff
@@ -302,6 +352,27 @@ impl DrawableComponent for DetailsComponent {
...
+						(title_len, message_len)
+					},
+				)
+			});
+		let mut title_buffer =
+			vec![0u8; title_len + title_len / usize::from(width) + 1];
+		let mut message_buffer =
+			vec![
+				0u8;
+				message_len + message_len / usize::from(width) + 1
```
AmmarAbouZor, line 374: This is the same as the above and it should fail with none-ASCII chars

AmmarAbouZor (commented):
Using the struct `bwrap::Wrapper` will use manual memory management and ask to provide the bytes-buffers (which can go wrong easily), but they have the other struct `bwrap::EasyWrapper` which uses the standard library for memory management.

With the standard library feature you can actually use the macro `bwrap::wrap!` to wrap the text easily without having to change the method signatures since it has similar signature to `textwrap::wrap` methods and will save you all the headache from dealing with converting between utf8 and byte-slices because the current implementation will fail with none-ASCII characters.

TheBlackSheep3:
Closed in favor of #1792

PR #1783, opened 2023-07-29
switch back to `dirs`

### LINKED ISSUE #1782: switch back to `dirs`
since we switched to `dirs-next` the old but original `dirs` got maintained again and to simplify our dependencies we should switch back.

extrawurst:
closes #1782 

~~right now blocked because somehow `shellexpand` pins us back to also use `dirs` version 4.x which is weird because its toml allows dirs `>=4, <6`~~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ chrono = { version = "0.4", default-features = false, features = [ "clock" ] }
 clap = { version = "4.1", features = [ "env", "cargo" ] }
 crossbeam-channel = "0.5"
 crossterm = { version = "0.26.1", features = [ "serde" ] }
-dirs-next = "2.0"
+dirs = "5.0"
 easy-cast = "0.5"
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -30,7 +30,7 @@ chrono = { version = "0.4", default-features = false, features = [ "clock" ] }
 clap = { version = "4.1", features = [ "env", "cargo" ] }
 crossbeam-channel = "0.5"
 crossterm = { version = "0.26.1", features = [ "serde" ] }
-dirs-next = "2.0"
+dirs = "5.0"
```
TheBlackSheep3, line 33: Is there a reason for referencing version `5.0` instead of `5.0.1` like it is referenced in `Cargo.lock`?

extrawurst: Because I don’t need to pin to patch releases if semver is taken seriosly

PR #1792, opened 2023-08-05
Switch to bwrap

### LINKED ISSUE #1762: switch to bwrap

TheBlackSheep3:
This Pull Request fixes/closes #1762.

It changes the following:
- removed dependency textwrap
- added dependency bwrap 

I followed the checklist:
- [x] I extended unittests


TheBlackSheep3:
@AmmarAbouZor i thought it was easier to just approach it again from scratch. Now the macro `bwrap::wrap!` is used.

AmmarAbouZor:
It looks simple and elegant now :100:

src/components/commit_details/details.rs
```diff
@@ -86,11 +86,20 @@ impl DetailsComponent {
 		message: &CommitMessage,
 		width: usize,
 	) -> WrappedCommitMessage<'_> {
-		let wrapped_title = textwrap::wrap(&message.subject, width);
+		let width = if width < 1 { 1 } else { width };
```
extrawurst, line 89: lets use `width.max(1)` here. weird that clippy does not suggest this 🤔

src/components/commit_details/details.rs
```diff
@@ -86,11 +86,20 @@ impl DetailsComponent {
...
+		let wrapped_title = bwrap::wrap!(&message.subject, width)
+			.split('\n')
```
extrawurst, line 91: are we sure os independent that line breaks here will always be just `\n` and not any of the other potential encodings?

src/components/commit_details/details.rs
TheBlackSheep3: Thank you for the hint. It really sounds like something that should be recognized by clippy.

src/components/commit_details/details.rs
TheBlackSheep3: Valid point. I think `.lines()` should address this issue.

extrawurst:
@TheBlackSheep3 thank you ❤️

PR #1818, opened 2023-08-20
Jump to commit via sha

### LINKED ISSUE #1799: open specific commit sha view
A more simple case than #1791: open an input box and jump to a specific commit sha in the commit detail view. this can be done without traversing the whole log and is therefore a much simpler implementation

this picks up [this comment](https://github.com/extrawurst/gitui/issues/1791#issuecomment-1666747701).

AmmarAbouZor:
This Pull Request fixes/closes #1799

This PR stills work in progress

It changes the following:
- Adds new function the commits-log to jump to commit via its sha
- Default Keybinding is currently `Ctrl-j`
- The input pupop has live validation
- After closing the input box it'll jump to the commit with the matching SHA

The following point need clarification:
- The live validation retrieve the repo with each keystroke to check if the sha is valid via the method  `git2::revparse_single()`. I've tested it with big repos and didn't see any performance hit with it. But we still can make simple live validation (only hash chars, no spaces, length at most 40) If the current implementation's performance isn't good enough.
- The default keybinding 
- Validation error text
- Title and placeholder texts  

I'll add a demo for the current progress

I followed the checklist:
- [X] I added unittests


AmmarAbouZor:
Here is a small demo for the current progress

https://github.com/extrawurst/gitui/assets/35040531/619359fa-b4cf-4ba3-9d7d-aedcbccf6444

joyously:
Does it jump to it even if that SHA is not in the visible list?

AmmarAbouZor:
@joyously Yes it jumps to any commit is the repository since it uses the well-tested method `select_commit()` in `revlog`

extrawurst:
> Does it jump to it even if that SHA is not in the visible list?

yes it will only be able to go to commits in the current branch. otherwise it will show an error right now. #81 will introduce looking at all branches and fix this

extrawurst:
@AmmarAbouZor this looks great. 

To not add yet another command to the log view I would like this to be an option in the new search popup `LogSearchPopupComponent`. Logically it also makes sense to be in there. once this in place I will merge this.

AmmarAbouZor:
@extrawurst I think it would be better in the search popup too. 
I'd add a new option there `Commit SHA` Or  `Commit Hash` and move the implementation there. 
Could you please have a look at #1816 before. This would save me one merge conflict :wink:

extrawurst:
#1816 is merged

AmmarAbouZor:
@extrawurst  Jump to SHA is embedded in `LogSearchPopupComponent`.
- I added a new mode there with the name 'commit hash'
- No other option can be active along with the commit hash mode
- The validation still happens but it's not visible to the users. No execution can happened if the commit hash mode is active and no valid SHA exists


https://github.com/extrawurst/gitui/assets/35040531/5f37692a-90f4-48d3-9251-3ea1fd520b43

AmmarAbouZor:
Ok I'll get back to the previous implementation with a separate command (currently default `Ctrl-j`) and make the validation better visualized

extrawurst:
> Ok I'll get back to the previous implementation

but do not add a new popup. Reuse this one please. Make it use a mode neun or something to store what mode it is in at the moment

joyously:
For accessibility, it's important that information is not conveyed only with color. (7% of men are red/green color blind) 
In the demo, it looked like less than 3 characters was considered invalid, but it matched at 4. Perhaps a better UI would show the matching hashes, which the user could pick from, although that might be a bit much for a terminal... (but the whole point is that the hash is not visible)

AmmarAbouZor:
> For accessibility, it's important that information is not conveyed only with color. (7% of men are red/green color blind)
In the demo, 

We can show a validation message in the right bottom corner like in the commit popup. (first iteration of this PR used that)


Considering that the users won't type the hashes themselves in the most cases but they will paste from the clipboard, I don't think that implementing an auto-completion here would add more value.
I can though complete the SHA in the textbox once the users have typed enough characters to determine one single commit, but this can confuse the users in case they are typing a commit that doesn't exist in the repository

asyncgit/src/sync/commits_info.rs
```diff
@@ -31,6 +31,17 @@ impl CommitId {
 	pub fn get_short_string(&self) -> String {
 		self.to_string().chars().take(7).collect()
 	}
+
+	/// Tries to retrieve the `CommitId` form the revision if exists in the given repository
+	pub fn from_revision(
+		repo_path: &RepoPath,
+		revision: &str,
+	) -> Result<Self> {
+		let repo = repo(repo_path)?;
```
extrawurst, line 40: please add `scope_time` here so we can see if this is taking significant time

src/app.rs
```diff
@@ -904,10 +905,16 @@ impl App {
...
+				self.log_search_popup.open_search_mode()?;
+				flags
+					.insert(NeedsUpdate::ALL | NeedsUpdate::COMMANDS);
+			}
+			InternalEvent::OpenLogJumpCommitSha => {
+				self.log_search_popup.open_jump_commit_mode()?;
 				flags
 					.insert(NeedsUpdate::ALL | NeedsUpdate::COMMANDS);
 			}
+
```
extrawurst, line 917: why empty line here?

src/keys/key_list.rs
```diff
@@ -88,6 +88,7 @@ pub struct KeysList {
 	pub log_reset_comit: GituiKeyEvent,
 	pub log_reword_comit: GituiKeyEvent,
 	pub log_find: GituiKeyEvent,
+	pub log_jump_commit_sha: GituiKeyEvent,
```
extrawurst, line 91: `find_commit_sha`

src/queue.rs
```diff
@@ -136,6 +136,10 @@ pub enum InternalEvent {
 	RewordCommit(CommitId),
 	///
 	CommitSearch(LogFilterSearchOptions),
+	///
+	OpenLogJumpCommitSha,
```
extrawurst, line 140: not needed. make it feature of existing search pop

src/queue.rs
```diff
@@ -136,6 +136,10 @@ pub enum InternalEvent {
...
+	///
+	JumpToCommit(CommitId),
```
extrawurst, line 142: not needed. duplicate of `SelectCommitInRevlog()`

src/strings.rs
```diff
@@ -32,6 +32,7 @@ pub static PUSH_TAGS_STATES_DONE: &str = "done";
 pub static POPUP_TITLE_SUBMODULES: &str = "Submodules";
 pub static POPUP_TITLE_FUZZY_FIND: &str = "Fuzzy Finder";
 pub static POPUP_TITLE_LOG_SEARCH: &str = "Search";
+pub static POPUP_TITLE_LOG_JUMP_COMMIT: &str = "Jump to Commit";
```
extrawurst, line 35: no needed. reuse search popup

src/tabs/revlog.rs
```diff
@@ -556,6 +556,13 @@ impl Component for Revlog {
 					self.queue
 						.push(InternalEvent::OpenLogSearchPopup);
 					return Ok(EventState::Consumed);
+				} else if key_match(
```
extrawurst, line 559: make it part of search-popup

src/components/log_search_popup.rs
```diff
@@ -32,19 +35,28 @@ enum Selection {
 	AuthorsSearch,
 }
 
+enum PopupMode {
+	Search,
+	JumpToCommitHash,
+}
+
 pub struct LogSearchPopupComponent {
+	repo: RefCell<RepoPath>,
```
extrawurst, line 44: please compare in other places of code: we have `RepoPathRef` for this

extrawurst (changes requested):

extrawurst:
> back to the previous implementation

that was not the idea. the command for jump-to-commit-sha should be only available in the search-popup to prevent littering the revlog command list more. it contextually also makes sense to be a sub behaviour of `find`/`search`.

ideally you

src/strings.rs
```diff
@@ -1672,4 +1673,17 @@ pub mod commands {
 			CMD_GROUP_BRANCHES,
 		)
 	}
+
+	pub fn find_commit_sha(
+		key_config: &SharedKeyConfig,
+	) -> CommandText {
+		CommandText::new(
+			format!(
+				"Commit Sha [{}]",
```
extrawurst, line 1682: better `Search Hash` ?

AmmarAbouZor:
- Commit sha command is shown in LogSearchPopup with the default keybinding `Ctrl-j`
- Validation message is shown on the right bottom corner when the sha is invalid only
- `Esc` in Commit SHA will go back the search popup

https://github.com/extrawurst/gitui/assets/35040531/f8d2c6c5-c561-40ac-9161-59c3a2055003

extrawurst:
Thanks for sticking with this. Good work! ❤️

Issue #1791, opened 2023-08-05
search/filter commit(s)


extrawurst:
We want to substring / fuzzy search for 
* commit sha (or parts of it), 
* authors fields
* commit message
* filenames appearing in diff
* content of diffs

fixes #449, #429, #1790

Note: **Fuzzy** does not mean realtime in this case. It means we do not need perfect matching, mostly relying on `fuzzy_matcher`

AmmarAbouZor:
#### SHA

I think we can at first separate using sha from the rest because it's one to one search (each sha will give one commit only).
For that  I would suggest a "jump to commit" function which opens an input box with a keybinding to past from the system clipboard then when it's applied the commit will be selected if found or some message can be shown

#### Filter/Fuzzy Matches

I think there is a trade off we need to make here...
- Live fuzzy matches is more intuitive and easier to use but it needs all the data to be loaded in memory so it will not give misleading information.
- A "classic" Filter function: Filter popup with the wanted options >> apply the filter to show only the commits that meet the filter criteria. This approach isn't as intuitive as the fuzzy matches one but it comes with less overhead since we don't need to have all the data in memory and the filtering will be applied only once and not with each keystroke.

bewlay:
Listing all commits that modify a specific file would be very useful (maybe this is covered by diff filtering).

bewlay:
Indeed, I wasn't aware of that feature. Very nice.

extrawurst:
@AmmarAbouZor your comment lead me to split out a separate issue for this particular use case: https://github.com/extrawurst/gitui/issues/1799

Rodrigodd:
Recently (a month ago, actually) I have opened the PR #1753, that finish the work done by #506 and #672. That PR implements searching by sha, author, message and tag, including controls for each one, but not in a fuzzy way.

I am not sure if you @extrawurst missed it, or it just isn't relevant for this issue.

PR #1830, opened 2023-08-23
Use `nucleo` instead of `fuzzy-matcher` 

### LINKED ISSUE #1805: consider `nucleo` instead of fuzzy-finder
see https://github.com/helix-editor/nucleo. seems pretty new (release `0.0.0`) and not sure it will work as a drop-in replacement yet. but seems pretty fast

auvred:
This Pull Request fixes/closes #1805.

It changes the following:
- replace `fuzzy-matcher` with `nucleo` in `FuzzyFindPopup`

I followed the checklist:
- [ ] I added unittests

--- 

According to the `nucleo` [docs](https://docs.rs/nucleo/0.1.1/nucleo/struct.Matcher.html), the `Matcher` allocates some heap memory at the creation, that's why it should be reused

I'm not sure that this change will be merged, so for now I'll open a draft PR


src/components/fuzzy_find_popup.rs
```diff
@@ -88,17 +90,38 @@ impl FuzzyFindPopup {
 		self.filtered.clear();
 
 		if let Some(q) = &self.query {
-			let matcher =
-				fuzzy_matcher::skim::SkimMatcherV2::default();
+			let mut line_content_buf = Vec::new();
```
extrawurst, line 93: maybe we should also cache these buffers, right?

extrawurst: maybe just wipe them as soon as one closes the window?

Cargo.lock
```diff
@@ -649,15 +655,6 @@ dependencies = [
  "slab",
 ]
 
-[[package]]
-name = "fuzzy-matcher"
```
extrawurst, line 653: is this branch up to date with master? I don't see the removal of `fuzzy-matcher` in `asyncgit` which got introduced in #1791

extrawurst (changes requested):
please bring up to date with master and remove all usages of `fuzzy-matcher`

src/components/fuzzy_find_popup.rs
auvred: Thank you for the review! I moved them to the `FuzzyFindPopup` alongside the `matcher`. Maybe I should move `matcher` and `matcher_*` to a new struct to omit this prefix?

Cargo.lock
auvred: Sorry, this branch was really outdated, I merged latest changes from master

After a little digging in the [`asyncgit/src/sync/logwalker.rs`](https://github.com/extrawurst/gitui/blob/master/asyncgit/src/sync/logwalker.rs), I realised that it would require more changes than in `fuzzy_find_popup.rs`

The main problem I'm struggling with is the [`fuzzy_match`](https://docs.rs/nucleo/0.1.1/nucleo/struct.Matcher.html#method.fuzzy_match) borrows Matcher as mutable, this will require some changes in the current implementation

The shortest way I see is this: we can wrap `matcher` in `LogFilterSearch` into `RwLock`, this will allow us to borrow filter as mutable only during the actual matching. Otherwise we'll be required to handle the entire filter and closures that borrow it as mutable everywhere (maybe I'm wrong)

Could you please advise how best to implement this?

extrawurst: `Mutex` is fine here since we won't have multiple readers and any benefit of `RwLock` for this.

extrawurst:
btw. I would like to see nucleo mature a bit before we can merge this: https://github.com/helix-editor/nucleo/issues/15

Cargo.lock
pascalkuthe: I am not really aware of how your codebase works but I might mention here that `fuzzy-matcher` uses an implicit internal thread local together with interior mutability to hide the same internal mutability that nucleo has (and has similarly expensive allocations although they are lazy instead of eagerly). So if you are doing the matching in parallel (which I would recommend) the mutex will hinder nucleo and you would get better results by using one matcher instance per thread (the allocation is large but not that large either).  

Btw if you are looking to add an fzf like filter I would really recommend that you consider the high-level nucleo API. It performs the matching in parallel and can perform millions of matches with barely perceptible delay (on a beefy CPU). It's what we use in helix for the file picker (you can try it out by building this branch https://github.com/helix-editor/helix/pull/7814 and running it in your root directory). 

It also supports fully lock-free (and almost always wait-free) parallel streaming so if you are iterating revs in one (or multiple)  background threads that would well too

extrawurst:
@auvred please update to current master and new nucleo release: https://github.com/helix-editor/nucleo/issues/15#issuecomment-1697448141

auvred:
I'm a little busy in the next few days, I'll continue working on this PR soon, if it will still be relevant

It seems that it would be better to use the high-level api as pascalkuthe advised :)

extrawurst:
@auvred its not pressing to get it in. how about reworking it to use the high level API?

auvred:
I'm all for it!

auvred:
I will close this pull request for now so as not to interfere with the others who want to implement this feature. So if anyone wants to work on this, fill free to pick it up

PR #1852, opened 2023-08-28
Speedup CI

### LINKED ISSUE #1844: Room for potential CI speedups
Looking at a [CI run](https://github.com/extrawurst/gitui/actions/runs/5991565477), both `build-linux-arm` and `build-linux-musl` don't appear to be using a cache, is this intentional? If not, adding one should reduce the step times by a decent margin. They definitely won't be as quick as the native builds but a gain is a gain 😅 

Additionally, there is a wall of deprecation warnings, most of which should be coming from `actions-rs/toolchain@v1` which is not maintained anymore, a maintained alternative is https://github.com/dtolnay/rust-toolchain.

Another suggestion would be to perhaps look into [`rust-cache`](https://github.com/Swatinem/rust-cache) for caching instead of doing it manually as it comes with some more defaults that might be useful.

AntoniosBarotsis:
This Pull Request closes #1844.

### Fix Warnings

- use `actions/checkout@v3`
- switch from `actions-rs/toolchain` (not maintained) to [`dtolnay/rust-toolchain`](https://github.com/dtolnay/rust-toolchain)
- switch from `actions-rs/audit-check` (not maintained) to [`rustsec/audit-check`](https://github.com/rustsec/audit-check)

These changes resolved all the warnings.

### Caching

Used [`rust-cache`](https://github.com/Swatinem/rust-cache) as mentioned in the issue.

> **Note** 
> `rust-cache`:
> - caches `.cargo/` and `target/`
> - hashes `Cargo.lock` for the key automatically
>
> hence their removal from the ci script 

I kept the `cache-name` around and added a `target` variable as well.

## Is it any better?

I think the best way to figure that out would be to wait for another commit to be made to master which I can merge to my fork so we can compare the CI times between that and the original.

I forked at https://github.com/extrawurst/gitui/commit/b36ffc9481cbc30000f43e2a7d6c72cb834bb3bd and here's some numbers

- original repo: [27m 21s](https://github.com/extrawurst/gitui/actions/runs/5998812806/usage)
- fork with invalidated cache: [25m 35s](https://github.com/AntoniosBarotsis/gitui/actions/runs/6000180026/usage)
- fork with warm cache: [14m 5s](https://github.com/AntoniosBarotsis/gitui/actions/runs/5999723565/usage)

> In my experience, +- 2 minutes can often just be random noise when it comes to GA

I would guess that the warm cache time is more representative of the change since, well, the original repository should also have a warm cache. But again, a more serious comparison would be to wait for the next commit.

And also not having a wall of warnings makes it easier to spot any that are actually relevant when they appear 😅 

## Other

I also mention this in the `yml` itself but for some reason [there seems to be some issue with the `x86_64-unknown-linux-musl` target and the `dtolnay/rust-toolchain` action](https://github.com/dtolnay/rust-toolchain/issues/83).

Even though the rustup target should be installed, [the subsequent build fails](https://github.com/AntoniosBarotsis/gitui/actions/runs/5999098146/job/16268573800). [I instead install it manually](https://github.com/extrawurst/gitui/commit/8601e6411ab07c40409e65bb34f202f4d5a534f2) which works fine (it is also cached so subsequent runs don't suffer).

Let me know if you have any questions or want any changes made!

Also the last CI run after I changed a comment decided to fail due to a [timeout](https://github.com/AntoniosBarotsis/gitui/actions/runs/6001112261/job/16274588502), I'm guessing because I've been running it quite a few times in the last few hours. Anyway, seems unrelated + i ran the failed task again and it worked fine

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
     - name: Get version
       id: get_version
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -25,8 +25,7 @@ jobs:
       env:
         cache-name: ci
       with:
-        key: ${{ matrix.os }}-${{ env.cache-name }}-stable
-        shared-key: "shared"
+        shared-key: ${{ matrix.os }}-${{ env.cache-name }}-stable
 
     - name: Install Rust
       uses: dtolnay/rust-toolchain@stable
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -26,6 +26,7 @@ jobs:
         cache-name: ci
       with:
         key: ${{ matrix.os }}-${{ env.cache-name }}-stable
+        shared-key: "shared"
 
     - name: Install Rust
       uses: dtolnay/rust-toolchain@stable
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -21,11 +21,9 @@ jobs:
       run: echo ::set-output name=version::${GITHUB_REF/refs\/tags\//}
 
     - name: Install Rust
-      uses: actions-rs/toolchain@v1
+      uses: dtolnay/rust-toolchain@stable
       with:
         toolchain: stable
-        profile: minimal
-        override: true
         components: clippy
 
     - name: Build
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -20,6 +20,13 @@ jobs:
       id: get_version
       run: echo ::set-output name=version::${GITHUB_REF/refs\/tags\//}
 
+    - name: Restore cargo cache
+      uses: Swatinem/rust-cache@v2
+      env:
+        cache-name: ci
+      with:
+        key: ${{ matrix.os }}-${{ env.cache-name }}-stable
+
     - name: Install Rust
       uses: dtolnay/rust-toolchain@stable
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,7 @@ jobs:
         cache-name: ci
       with:
         key: ${{ matrix.os }}-${{ env.cache-name }}-${{ matrix.rust }}
+        shared-key: "shared"
 
     - name: MacOS Workaround
       if: matrix.os == 'macos-latest'
@@ -97,6 +98,7 @@ jobs:
         cache-name: ci
       with:
         key: ubuntu-latest-${{ env.cache-name }}-${{ matrix.rust }}
+        shared-key: "shared"
 
     - name: Install Rust
       uses: dtolnay/rust-toolchain@stable
@@ -143,6 +145,7 @@ jobs:
         cache-name: ci
       with:
         key: ubuntu-latest-${{ env.cache-name }}-${{ matrix.rust }}
+        shared-key: "shared"
 
     - name: Install Rust
       uses: dtolnay/rust-toolchain@stable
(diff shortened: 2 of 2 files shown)


extrawurst:
can you make sure `cd.yml` is kept in sync?

AntoniosBarotsis:
I can take a look at that sometime today/tomorrow as well

AntoniosBarotsis:
I merged https://github.com/extrawurst/gitui/commit/f639f4a0d03a5a0f1d0220320878a71e2aa76700 and https://github.com/extrawurst/gitui/commit/c38b1d1e1ce5008bb5a4055d84bd8df950591214 into my fork and here's some more numbers:

- original repo: [21m 26s](https://github.com/extrawurst/gitui/actions/runs/6008357126/usage) and [20m 27s](https://github.com/extrawurst/gitui/actions/runs/6011744678/usage)
- fork (ci branch): [10m 43s](https://github.com/AntoniosBarotsis/gitui/actions/runs/6012665209/usage)
- fork (master branch): [26m 58s](https://github.com/AntoniosBarotsis/gitui/actions/runs/6012659058/usage)

The fork's master branch was included as a sanity check to make sure that it has roughly the same times as the original repo (which it should as I have not modified it). Somehow it ended up being a bit slower, go figure.

The original repo shows 2 times as each commit was pushed separately unlike my fork where both were pushed together. 

The great news is that this also improves the `build` task times!

Overall, this seems to be quite a bit faster, at least in these 2 commits that I tested it on.

I won't be able to really test the CD workflow but I'll make the same changes there nontheless.

extrawurst:
would like to put this into action in `0.24.1` patch release tomorrow. will wait for the changes in `cd.yml`

AntoniosBarotsis:
I'll definitely try to have it ready by then so I can see it tested as well.

AntoniosBarotsis:
I checked the actions used in the CD workflow and made sure they are either on their latest version + switched the toolchain one as I did in the CI workflow earlier.

I also opted to use `${{ matrix.os }}-${{ env.cache-name }}-${{ matrix.rust }}` as the cache key everywhere instead of prefixing `build-linux-musl` and `build-linux-arm` with the target. This should not change the CI times, I'll run a test on an empty commit to make sure that that is the case.

Lastly, by getting CD to use the same cache key format, it should also get any speedups that CI got.

AntoniosBarotsis:
I changed from the `key` attribute to `shared-key` to achieve this. Manually inspecting the generated keys (which you can view in the logs of each action ` > Restore Cargo Cache > Cache Configuration > Cache Key`) reveals that for example `build-linux-arm (stable)` and `build (ubuntu-latest, stable)` have the same cache key which was not the case before. This should also guarantee that the CD workflow will use the existing cache generated by CI.

I'm gonna wait for the cache to warm up and run the workflow once more on an empty commit to make sure nothing changed in the times. After that, I think everything is done from my side so far, let me know!

AntoniosBarotsis:
This actually seems to worsen performance as sometimes the cache is not properly saved since multiple jobs are trying to access it at the same time

```
Failed to save: Unable to reserve cache with key v0-rust-ubuntu-latest-ci-1.65-e8f5d070-722cda0f, 
another job may be creating this cache. More details: Cache already exists.
```

To get around this, I will only keep the shared keys for one of the CI jobs (which is really all that was needed, in retrospect). Taking a look at [the latest CD run](https://github.com/extrawurst/gitui/actions/runs/5313022967) it seems that most of the time is actually spent building for windows and not arm which I would initially expect. This means that sharing the cache from the `build` job should be pretty effective in reducing CD times.

AntoniosBarotsis:
This last change has made CI run in 10 minutes again like last time so it seems to be working fine again.

This should also benefit CD heavily, to make sure that the cache is being utilized correctly when we do run CD we should check that the key for `build (windows-latest, stable)` for example is the same as `release (windows-latest)`.

One small issue with this is that CI currently does not run on `ubuntu-22.04`, just `ubuntu-latest` which, while they are technically the same I belive, will result in different cache keys and thus the CD task won't have a cache to utilize from the previous CI run. Each CD run however will update the cache meaning that the next run will have one, just not as up to date as the other OSes.

extrawurst:
Thank you!

PR #1914, opened 2023-10-03
add syntax highlighting for blame view (#745)

### LINKED ISSUE #745: use syntax highlighter in blame view
since #727 we have a way to visualise code in a syntax highlighted way, I think it could be added to the blame view aswell.

tdtrung17693:
This Pull Request fixes/closes #745.

It changes the following:
- Add syntax highlighting for blame view

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -399,7 +399,6 @@ impl App {
 			self.status_tab.update_git(ev)?;
 			self.stashing_tab.update_git(ev)?;
 			self.revlog.update_git(ev)?;
-			self.blame_file_popup.update_git(ev)?;
 			self.file_revlog_popup.update_git(ev)?;
 			self.inspect_commit_popup.update_git(ev)?;
 			self.compare_commits_popup.update_git(ev)?;
@@ -411,6 +410,7 @@ impl App {
 		}
 
 		self.files_tab.update_async(ev)?;
+		self.blame_file_popup.update_async(ev)?;
 		self.revision_files_popup.update(ev)?;
 		self.tags_popup.update(ev);
 
--- a/src/components/blame_file.rs
+++ b/src/components/blame_file.rs
@@ -10,27 +10,72 @@ use crate::{
 	queue::{InternalEvent, Queue, StackablePopupOpen},
 	string_utils::tabs_to_spaces,
 	strings,
-	ui::{self, style::SharedTheme},
+	ui::{self, style::SharedTheme, AsyncSyntaxJob, SyntaxText},
+	AsyncAppNotification, AsyncNotification, SyntaxHighlightProgress,
 };
 use anyhow::Result;
 use asyncgit::{
-	sync::{BlameHunk, CommitId, FileBlame},
+	asyncjob::AsyncSingleJob,
+	sync::{BlameHunk, CommitId, FileBlame, RepoPathRef},
 	AsyncBlame, AsyncGitNotification, BlameParams,
 };
+use crossbeam_channel::Sender;
 use crossterm::event::Event;
 use ratatui::{
 	backend::Backend,
 	layout::{Constraint, Rect},
 	symbols::line::VERTICAL,
-	text::Span,
+	text::{Span, Text},
 	widgets::{Block, Borders, Cell, Clear, Row, Table, TableState},
 	Frame,
 };
+use std::{convert::TryInto, path::Path};
 
 static NO_COMMIT_ID: &str = "0000000";
 static NO_AUTHOR: &str = "<no author>";
 static MIN_AUTHOR_WIDTH: usize = 3;
 static MAX_AUTHOR_WIDTH: usize = 20;
+
+struct SyntaxFileBlame {
+	pub file_blame: FileBlame,
+	pub styled_text: Option<SyntaxText>,
+}
+
+impl SyntaxFileBlame {
+	fn path(&self) -> &str {
+		&self.file_blame.path
+	}
+
+	fn commit_id(&self) -> &CommitId {
+		&self.file_blame.commit_id
+	}
+
+	fn lines(&self) -> &Vec<(Option<BlameHunk>, String)> {
+		&self.file_blame.lines
+	}
+}
+
+enum BlameProcess {
+	GettingBlame(AsyncBlame),
+	SyntaxHighlighting {
+		unstyled_file_blame: SyntaxFileBlame,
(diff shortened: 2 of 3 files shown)


tdtrung17693:
Hi @extrawurst, could you take a look into this PR? If there is anything need to change please let me know. Thank you!

src/app.rs
```diff
@@ -532,6 +533,7 @@ impl App {
 		}
 
 		self.files_tab.update_async(ev)?;
+		self.blame_file_popup.update_syntax(ev);
```
extrawurst, line 536: i find it misleading to call it `update_syntax` as you do not check for which event it is, so technically its just `update_async` and inside you make sure to react only on the right event

src/components/blame_file.rs
```diff
@@ -9,30 +9,40 @@ use crate::{
...
 
 static NO_COMMIT_ID: &str = "0000000";
 static NO_AUTHOR: &str = "<no author>";
 static MIN_AUTHOR_WIDTH: usize = 3;
 static MAX_AUTHOR_WIDTH: usize = 20;
 
+struct SyntaxFileBlame {
+	///
+	pub commit_id: CommitId,
+	///
```
extrawurst, line 41: why some empty comments and some without? consistency please

src/components/blame_file.rs
```diff
@@ -81,6 +93,13 @@ impl DrawableComponent for BlameFileComponent {
 			];
 
 			let number_of_rows = rows.len();
+			let title_with_highlight_progress = format!(
```
extrawurst, line 96: lets only show the percentage while its loading. no need to confuse with `100%` the whole time after the parsing happened

src/components/blame_file.rs
```diff
@@ -295,6 +315,10 @@ impl BlameFileComponent {
 			table_state: std::cell::Cell::new(TableState::default()),
 			key_config,
 			current_height: std::cell::Cell::new(0),
+			async_highlighting: AsyncSingleJob::new(
```
extrawurst, line 318: i would argue we should put it in an option and only keep it around while there is an async job ongoing and initialize it to `None` first

src/components/blame_file.rs
```diff
@@ -371,6 +402,40 @@ impl BlameFileComponent {
...
 
+	pub fn update_syntax(&mut self, ev: AsyncNotification) {
+		if let AsyncNotification::App(
+			AsyncAppNotification::SyntaxHighlighting(progress),
+		) = ev
+		{
+			match progress {
+				SyntaxHighlightProgress::Progress => {
+					self.syntax_progress =
+						self.async_highlighting.progress();
```
extrawurst, line 413: i have to double check but i think the progress retrieval is pretty lightweight, no need to cache it

src/components/blame_file.rs
```diff
@@ -371,6 +402,40 @@ impl BlameFileComponent {
...
+							if let Some(ref mut file_blame) =
+								&mut self.file_blame
+							{
+								if syntax.path()
+									== Path::new(&file_blame.path)
+								{
+									file_blame.styled_text =
+										Some(syntax);
+								}
+								// *(self.file_blame.s = Either::Left(syntax);
```
extrawurst, line 430: please clean up not needed comments

src/components/blame_file.rs
```diff
@@ -403,11 +468,18 @@ impl BlameFileComponent {
...
+				file_blame
+					.styled_text
+					.as_ref()
+					.map(std::convert::Into::into)
+			});
 		self.file_blame
 			.as_ref()
 			.map_or_else(Vec::new, |file_blame| {
-				file_blame
+				return file_blame
```
extrawurst, line 482: please keep it consistent with the non-return call style, last statement without `;` is the return value

src/components/blame_file.rs
```diff
@@ -417,19 +489,48 @@ impl BlameFileComponent {
...
+			.map(|l| l.1.clone())
+			.collect::<Vec<_>>();
+		let text = tabs_to_spaces(raw_lines.join("\n"));
+
+		self.syntax_progress = Some(ProgressPercent::empty());
+		self.async_highlighting.spawn(AsyncSyntaxJob::new(
+			text.clone(),
+			params.file_path.clone(),
+		));
+		self.async_highlighting.spawn(AsyncSyntaxJob::new(
```
extrawurst, line 520: why spawn the same job twice?

extrawurst (changes requested):

src/components/blame_file.rs
extrawurst: nvm i am blind :D

src/components/blame_file.rs
tdtrung17693: As I see in the codebase, this is currently the way that other components instantiate the async jobs. Also, keeping this way makes it easier to process the job's result, instead of multi-level pattern-matching.

extrawurst: in order to keep things maintainable i started using this pattern successfully in `Revlog` with `LogSearch`.
in our case here we can combine probably 4 member (`async_blame`, `syntax_progress`, `file_blame` and `async_highlighting`) variables into an enum depending on our state and it makes it easier to reason which ones are expected to be valid at what state.

extrawurst (changes requested):

src/components/blame_file.rs
tdtrung17693: Looks great. Let me try it.

extrawurst: What's the status here? Would be great to carry over the finish line!

extrawurst: Will put it in draft till reworked

extrawurst (changes requested):
Please update

src/components/blame_file.rs
tdtrung17693: I refactored it and combined all those 4 processes into the one below just as you suggested. 

```rust
enum BlameProcess {
	GettingBlame(AsyncBlame),
	SyntaxHighlighting {
		unstyled_file_blame: SyntaxFileBlame,
		job: AsyncSingleJob<AsyncSyntaxJob>,
	},
	Result(SyntaxFileBlame),
}
```

extrawurst (changes requested):
looks much better. please update to current master and then we can merge

extrawurst:
@tdtrung17693 thank you so much!

PR #1937, opened 2023-10-24
first PR for new textinput

pm100:
It has taken a long time. I had to work with tui-textarea to get new features added (that are useful to all users of the crate)

Rather than put a lot of text here please look at the big comment section in textarea.rs, its basically a read me. Read that before trying to follow the edits to the consumers of textinput, not many and they all follow the same pattern

The CI fails due to syn and bitflags have conflicting versions, but all else passes.

And as far as I can see everything works correctly, and there are no key meaning changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ shellexpand = "3.1"
 simplelog = { version = "0.12", default-features = false }
 struct-patch = "0.2"
 syntect = { version = "5.0", default-features = false, features = ["parsing", "default-syntaxes", "default-themes", "html"] }
-tui-textarea = {version ="0.2.3", default-features = false, features = ["ratatui-crossterm"]}
+tui-textarea = { version = "0.2.3", default-features = false, features = ["ratatui-crossterm"] }
 unicode-segmentation = "1.10"
 unicode-truncate = "0.2"
 unicode-width = "0.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,19 +24,19 @@ bytesize = { version = "1.3", default-features = false }
 chrono = { version = "0.4", default-features = false, features = ["clock"] }
 clap = { version = "4.4", features = ["env", "cargo"] }
 crossbeam-channel = "0.5"
-crossterm = { version = "0.26.1", features = ["serde"] }
+crossterm = { version = "0.27", features = ["serde"] }
 dirs = "5.0"
 easy-cast = "0.5"
 filetreelist = { path = "./filetreelist", version = "0.5" }
 fuzzy-matcher = "0.3"
 gh-emoji = { version = "1.0", optional = true }
-indexmap = "1.9"
+indexmap = "2.0"
 itertools = "0.11"
 log = "0.4"
 notify = "5.1"
 notify-debouncer-mini = "0.2"
 once_cell = "1"
-ratatui = { version = "0.21", default-features = false, features = ['crossterm', 'serde'] }
+ratatui = { version = "0.23", default-features = false, features = ['crossterm', 'serde'] }
 rayon-core = "1.11"
 ron = "0.8"
 scopeguard = "1.2"
@@ -46,6 +46,7 @@ shellexpand = "3.1"
 simplelog = { version = "0.12", default-features = false }
 struct-patch = "0.2"
 syntect = { version = "5.0", default-features = false, features = ["parsing", "default-syntaxes", "default-themes", "html"] }
+tui-textarea = {version ="0.2.3", default-features = false, features = ["ratatui-crossterm"]}
 unicode-segmentation = "1.10"
 unicode-truncate = "0.2"
 unicode-width = "0.1"
--- a/src/components/tag_commit.rs
+++ b/src/components/tag_commit.rs
@@ -83,8 +83,7 @@ impl<'a> Component for TagCommitComponent<'a> {
 					self.key_config.keys.tag_annotate,
 				) && self.is_valid_tag()
 				{
-					let tag_name: String =
-						self.input.get_text().into();
+					let tag_name: String = self.input.get_text();
 
 					self.input.clear();
 					self.input.set_title(
@@ -161,9 +160,9 @@ impl<'a> TagCommitComponent<'a> {
 
 	fn tag_info(&self) -> (String, Option<String>) {
 		match &self.mode {
-			Mode::Name => (self.input.get_text().into(), None),
+			Mode::Name => (self.input.get_text(), None),
 			Mode::Annotation { tag_name } => {
-				(tag_name.clone(), Some(self.input.get_text().into()))
+				(tag_name.clone(), Some(self.input.get_text()))
 			}
 		}
 	}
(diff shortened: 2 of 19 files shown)


Cargo.toml
```diff
@@ -46,6 +46,7 @@ shellexpand = "3.1"
 simplelog = { version = "0.12", default-features = false }
 struct-patch = "0.2"
 syntect = { version = "5.0", default-features = false, features = ["parsing", "default-syntaxes", "default-themes", "html"] }
+tui-textarea = { version = "0.2.3", default-features = false, features = ["ratatui-crossterm"] }
```
AmmarAbouZor, line 49: tui-textarea has release version 0.3.0 which doesn't specify which ratatui-version should be used with. This is very important to do before updating to ratatui 0.24.0 since they made a lot of breaking changes

src/components/tag_commit.rs
```diff
@@ -162,22 +160,15 @@ impl TagCommitComponent {
...
-		let gpgsign =
-			get_config_string(&self.repo.borrow(), "tag.gpgsign")
-				.ok()
-				.flatten()
-				.and_then(|val| val.parse::<bool>().ok())
-				.unwrap_or_default();
-		anyhow::ensure!(!gpgsign, "config tag.gpgsign=true detected.\ngpg signing not supported.\ndeactivate in your repo/gitconfig to be able to tag without signing.");
-
+	///
+	pub fn tag(&mut self) {
```
AmmarAbouZor, line 171: I didn't understand how this change is related to the textinput. Maybe I've missed something

src/components/textinput.rs
```diff
@@ -13,42 +15,183 @@ use crate::{
...
+					&& self.is_valid_tag()
+				{
+					....
+			// if not consumed here as 'special' pass to textinput
+			if self.input.event(ev)?.is_consumed() {
+				return Ok(EventState::Consumed);
+			}
+
+but that would have been a lot of change. So instead textinput ignores all keys that might be special for any UI. So , for example,
+^f never works now. (But ctrl-left works)
```
AmmarAbouZor, line 119: I totally agree, the suggested approach is more flexible and will keep the textinput more abstract since it can choose it's keybindings without fear of breaking another parts of the components. The current approach could be error prone one the long run since no one would think about testing all input places when changing a keybinding in the textbox

src/components/textinput.rs
```diff
@@ -13,42 +15,183 @@ use crate::{
...
+
+New line is shift-enter, ctrl-enter or ^m. These are all common new line editor commands (discord, emacs,...)
+
```
AmmarAbouZor, line 122: Those keys are tricky since they are different between terminal emulators and inside Tmux or Zellij. I think this will need more testing when the multi-lines feature get used somewhere in the code

src/components/textinput.rs
```diff
@@ -84,14 +225,17 @@ impl TextInputComponent {
...
 	}
 
 	/// Get the `msg`.
-	pub fn get_text(&self) -> &str {
-		self.msg.as_str()
+	pub fn get_text(&self) -> String {
+		self.textarea
+			.as_ref()
+			.map_or_else(String::new, |ta| ta.lines().join("\n"))
 	}
```
AmmarAbouZor, line 239: I would introduce a new method here as well that return a slice without allocating new memory, since the caller needs sometimes only to know if there are an input or not or how long the input is, without having to allocate new memory.

A method like this would be great
```rust
	pub fn get_text_lines(&self) -> &[String] {...}
```

src/components/textinput.rs
```diff
@@ -233,96 +306,16 @@ impl TextInputComponent {
...
 
 	fn draw_char_count<B: Backend>(&self, f: &mut Frame<B>, r: Rect) {
-		if self.char_count > 0 {
-			let w = Paragraph::new(format!(
-				"[{} chars]",
-				self.char_count
-			))
-			.alignment(Alignment::Right);
+		let count = self.get_text().len();
```
AmmarAbouZor, line 315: New string is allocated here just to to get its length. This could be written in a more verbose way but without allocating new memory
```rust
let count = self.textarea.as_ref().map_or(0, |txt| {
	txt.lines().iter().map(|line| line.len()).sum::<usize>()
	+ txt.lines().len()
	- 1
});

```

AmmarAbouZor (commented):
I think the implementation looks very good, I've left some comments about small details. 
I've compiled it and try it on my machine and noticed that the sizes of the input texts like the commit input have gotten smaller

pm100:
> I think the implementation looks very good, I've left some comments about small details. I've compiled it and try it on my machine and noticed that the sizes of the input texts like the commit input have gotten smaller

@AmmarAbouZor  when you say 'smaller' do you mean the font is smaller or that the display area is smaller. I dont see either issue but have only tested on windows. What platform did you see it on? I will work through the rest of your comments.

many dialogs use multiline, the most obvious is commit. Its one of the reasons I did this, the old textinput could show multiline commit messages but could not create them - you had to invoke the external editor. Having a choice of 'create new line' key seems to ensure that at least one will work

src/components/textinput.rs
pm100: i tried to not make significant changes to the caller code unless absolutely neccessary

src/components/textinput.rs
pm100: i tried to not make significant changes to the caller code unless absolutely neccessary

src/components/tag_commit.rs
pm100: it is a commit made since I last sync from upstream. I will merge it

Cargo.toml
pm100: tta 0.3 says `ratatui = { version = ">=0.23.0, <1", default-features = false, optional = true }`. Unless I am missing something, not impossible, cargo dependency versioning always surprises me

src/components/textinput.rs
pm100: many dialogs use multiline, the most obvious is commit. Its one of the reasons I did this, the old textinput could show multiline commit messages but could not create them - you had to invoke the external editor. Having a choice of 'create new line' key seems to ensure that at least one will work

Cargo.toml
pm100: size was a typo by me I set commit ui to single line rather than multi

AmmarAbouZor: Commit size issue is fixed by me too :+1: 

in tta 0.3 ratatui is specified as default in this section of cargo.toml
```toml
[features]
default = ["crossterm"]
ratatui = ["dep:ratatui"]
crossterm = ["ratatui", "dep:crossterm", "ratatui/crossterm"]
```
I've updated Ratatui to 0.24 and tui-textarea to 0.3.0, and it took ratatui as the default feature. The only problem was that I must change cargo.lock manually and remove all references to ratatui 0.23.0 in order to make compile again

src/components/textinput.rs
AmmarAbouZor: I meant that there are still no dialogs that accept typing new lines, where we can test how shift-enter, ctrl-enter or Ctrl-m would work. 
A potential problem here that on some terminal Ctrl-m could be sent as <CR> which will conform the commit in this case. I think there are another issue here where this point is discussed with more details

pm100:
closing this one, have worked with tui_textarea to improve things. New PR comin

Issue #1781, opened 2023-07-29
cargo update


extrawurst:
updating a few major dependencies is blocked right now by the ecosystem taking its time to follow the new major versions of `syn` and `bitflags`. and minor versions of `regex-syntax`, `redox_syscall`.

## `syn`
as it seems we have to wait for a few remaining crates to release new versions:
* `struct-patch-derive` (using old `syn`): https://github.com/yanganto/struct-patch/issues/19
  * it uses `proc-macro-error` which has no support for `syn2` merged&released yet: https://gitlab.com/CreepySkeleton/proc-macro-error/-/merge_requests/36
* `bugreport` which in turn requires `git-version` (using old `syn`): https://github.com/fusion-engineering/rust-git-version/pull/20 
  * or alternatively: https://github.com/sharkdp/bugreport/issues/13
* `chrono` -> `iana-time-zone` -> `iana-time-zone-haiku` (`v0.1.1`) which still depends transitively on syn1 via `cxx-build`: https://github.com/strawlab/iana-time-zone/pull/110

## `regex-syntax` ✅
* `syntect` (using old `regex-syntax`): https://github.com/trishume/syntect/issues/486

## `redox_syscall`
* `shellexpand` in turn using `redox_users` used by (using old `redox_syscall`): https://gitlab.redox-os.org/redox-os/users/-/merge_requests/36
* `notify` using `filetime` (`redox_syscall`): https://github.com/alexcrichton/filetime/issues/98
  * bump in `notify`: https://github.com/notify-rs/notify/pull/521
  * new `notify` release: https://github.com/notify-rs/notify/issues/513

## `bitflags`
* `ron` (`bitflags`): https://github.com/ron-rs/ron/issues/469
* `notify` (`bitflags`): https://github.com/notify-rs/notify/issues/513
* `crossterm` (`bitflags`): https://github.com/crossterm-rs/crossterm/issues/807
* `git2` (`bitflags`): https://github.com/rust-lang/git2-rs/issues/976
* `syntect`: https://github.com/trishume/syntect/pull/490

help needed: 
* `redox_syscall` itself does not support working with `bitflags` 2.x: [redox_syscall](https://gitlab.redox-os.org/redox-os/syscall/-/issues/34) (MR needed)

plus our very own:
* usage of `dirs-next`: https://github.com/extrawurst/gitui/pull/1783

yanganto:
`struct-patch-derive` already use syn2 after 0.3.0 release

[...]

PR #1963, opened 2023-11-27
replace textinput with tui-textarea

### LINKED ISSUE #1662: replace textinput with tui-textarea
The popup used for text input is fine in most cases but generates many requests for features in the multiline case, particularly the commit message.

There are several open requests on this. Rather than a piecemeal approach of adding bits and pieces I think it would be better to use a plugin component that offers 
- line split and join
- undo
- full multiline edit
- cut / paste
- .....

tui-textarea https://github.com/rhysd/tui-textarea is just such a component

I have done an experiment to see how easy it would be to drop in. Its pretty straightforward. Took a couple of hours

unless you think this is a completely stupid idea I will give it a less hacky go. You could then clone if from my repo and see what you think. The UI looks the same, its just that the multiline textinput supports full edit.

I also have a couple of points to discuss about it in terms of what keys should do what. But I want to see if you think this is at least a possibility before engaging is further debate.

PS - are you aware of the whole ratatui thing. https://github.com/tui-rs-revival/ratatui

pm100:
This Pull Request fixes/closes #1662 plus others mentioned in that issue.

Completely redone since my first PR. All changes are confined to a single file thanks to a lot of assistance from @ryhsd 

It changes the following:
- a complete rewrite of textinput.rs
- updates ratatui to 0.23

cargo deny check bans complains about
- itertools
- syn
- bitflags

Bitflags you know about
syn is compile time only
not sure what to do about itertools (you use 0.12 ratatui 0.23 uses 0.11)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * fix expansion of `~` in `commit.template` ([#1745](https://github.com/extrawurst/gitui/pull/1745))
 * fix hunk (un)staging/reset for # of context lines != 3 ([#1746](https://github.com/extrawurst/gitui/issues/1746))
 * fix delay when opening external editor ([#1506](https://github.com/extrawurst/gitui/issues/1506))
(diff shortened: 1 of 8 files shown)


extrawurst:
now that #1781 is closed please update and lets get this merged

PR #1979, opened 2023-12-15
ai commit summary creation using chat-gpt

extrawurst:
Right now this is not optimized for long diffs. 

ideas: 
- [ ] feed it less context if diff gets >4k
- [ ] feed list of files touched first and then diffs and cut at 4k point

### DIFF
--- a/git2-summarize/src/lib.rs
+++ b/git2-summarize/src/lib.rs
@@ -1,0 +1,85 @@
+//!	Uses Open API GPT-3 to summarize unified git diffs
+
+use openai_api_rs::v1::{
+	api::Client,
+	chat_completion::{self, ChatCompletionRequest},
+	common::GPT3_5_TURBO,
+	completion::{self, CompletionRequest},
+};
+
+/// Uses old GPT3_TEXT_DAVINCI_003 model to generate message
+///
+///  # Arguments
+///
+/// * `api_key` - open api key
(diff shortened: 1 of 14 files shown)


git2-summarize/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+	)
+	.max_tokens(200);
+
+	let result =
+		client.chat_completion(req).map_err(|e| e.message)?;
+
+	Ok(result
+		.choices
+		.get(0)
+		.ok_or_else(|| String::from("response.choises empty"))?
```
vladimir-salvador, line 49:
```suggestion
		.ok_or_else(|| String::from("response.choices empty"))?
```

git2-summarize/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+		.message
+		.content
+		.as_ref()
+		.ok_or_else(|| String::from("choise[0].message empty"))?
```
vladimir-salvador, line 53:
```suggestion
		.ok_or_else(|| String::from("choice[0].message empty"))?
```

vladejs:
Hi, I tested this on my Mac and noticed it works pretty good . How can we proritize this?

richgrov:
Any plans to continue work on this? I'd be open to helping if not

extrawurst:
I would appreciate that

Issue #1860, opened 2023-08-30
allow aborting search


extrawurst:
especially searching in the filenames of commits can take very long and can be annoying to have to kill `gitui` to abort

StemCll:
Hi @extrawurst :wave: 
I'm planning to take on this issue. Do you have any more details how to test this (maybe you have a certain repository to iterate over and search for "things"), what strategy should be undertaken when aborting etc.?

StemCll:
Hi @extrawurst 
I've pushed changes and created a PR. I haven't updated changelog yet - waiting for your CR :smile:

extrawurst:
@StemCll thanks for the groundwork you layed

PR #2000, opened 2024-01-07
feat(LogSearch): Allow aborting search

StemCll:
This Pull Request fixes/closes https://github.com/extrawurst/gitui/issues/1860
Tested on [Linux github repo](https://github.com/torvalds/linux) searching anything - it's big enough to make filtering take time.

It changes the following:
- Adds cancellation_flag to AsyncCommitFilterJob
- cancels LogSearch when pressed `escape`

I followed the checklist:
- [ ] I added unittests


asyncgit/src/filter_commits.rs
```diff
@@ -106,6 +112,10 @@ impl AsyncCommitFilterJob {
 				.collect::<Vec<(usize, CommitId)>>()
 				.par_chunks(1000)
 				.filter_map(|c| {
+					if self.cancellation_flag.load(Ordering::SeqCst) {
+						//cancel filter if cancellation_flag was set to true
+						return None;
+					}
```
extrawurst, line 118: this is not enough.

the (potentially) expensive filter function might still be called on a batch of 1k commits (which in my test filter also on the commit files can take up to 12s per batch) before realizing that the user aborted.

lets check the cancel flag before each filter function call and benchmark if that makes the normal search (without canceling) noticeably slower

extrawurst (changes requested):

asyncgit/src/filter_commits.rs
StemCll: Sure, will test it. What's your benchmark suite?

extrawurst: No suite, just observing after setting the abort flag and actually waiting for the cancel to happen to close the hi instead of closing the UI instantly. Then you see it with your own eyes

extrawurst: and to see whether the check inside the inner loop slows down search you can just use the search timing that is already shown in the UI after a search (it shows how long it took)

StemCll: @extrawurst in my search case the first version executed in 114.54053s. The latest one finished in  89.988s <- the check before `filter_map` does a better job (in my case at least...). I don't really like the two filter_maps solution, but couldn't come up with any better option to cancel rayon iterators

extrawurst: so its adds 20s to a normal non-aborted-search if we check the flag in the inner loop?

the problem is that this will close the UI before the search actually ends. i don't want the UI to hide the fact that there is still processing happening. we have to find a better solution then, it cannot be merged like this

extrawurst (commented):
can't be merged in this state

extrawurst:
finishing it via #2078

robinshen36:
@algora-pbc attempt

Just came across this and thought I'd reach out...

Been diving into CLI lately and this seems like a perfect fit.

Thinking of kicking things off by checking the vim keys - does that sound reasonable?

What's the budget looking like? Around $160-$280 feels about right, but flexible.

Talk soon hopefully.

PR #2023, opened 2024-01-24
Add ssh sign feature

### LINKED ISSUE #1149: Sign git commits with SSH keys

yanganto:
This Pull Request fixes #1149.

It changes the following:
- add `--ssh-key-path` and `--ssh-key-password` for the ssh signing key
- sign a commit when the ssh key is set

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,41 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "aead"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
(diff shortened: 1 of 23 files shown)


extrawurst:
i am working through a bunch of piled up issues here this week and come back to you @yanganto

extrawurst:
@yanganto based on your last [comment](https://github.com/extrawurst/gitui/issues/1149#issuecomment-1908414642) in the ref issue:

> There are things we need to think about before completing the PR. I have checked this signed commit with the ssh key will show verified tag in GitHub.
> 
> There will be a duplicate syn crate in the dependency, for 1.0 and 2.0

that should be fixed once you update to current master

> There will be a warning with the crate for RSA SSH key, but it breaks the current CI.

not sure what that means. can you fix the other ci compiler/clippy issues so that we can see that?

> The key path is passed from options, and we do not need additional TUI on it, because users do not change the signing key for different commits.

I am fine with no additional TUI for it in the first iteration. but I think we need some slightly simpler UX for providing this. Is there a standard way how this key is provided for example to git (the normal cli)? maybe env variables? or is it also purely via cli args?

yanganto:
This feature will work at #2047.

PR #2045, opened 2024-02-13
Replace textarea with tui-textarea

pm100:
PR for #1662 

now using tui-textarea v.04, that adds keyboard based text selection (hold down shift key while moving cursor with mouse)

reminder that a newline is made by pressing ctrl-enter or shift-enter

### DIFF
--- a/src/components/create_branch.rs
+++ b/src/components/create_branch.rs
@@ -104,7 +104,8 @@ impl CreateBranchComponent {
 				&strings::create_branch_popup_title(&env.key_config),
 				&strings::create_branch_popup_msg(&env.key_config),
 				true,
-			),
+			)
+			.with_input_type(super::InputType::Singleline),
 			theme: env.theme.clone(),
 			key_config: env.key_config.clone(),
 			repo: env.repo.clone(),
--- a/src/components/fuzzy_find_popup.rs
+++ b/src/components/fuzzy_find_popup.rs
@@ -42,7 +42,8 @@ impl FuzzyFindPopup {
 	///
 	pub fn new(env: &Environment) -> Self {
 		let mut find_text =
-			TextInputComponent::new(env, "", "start typing..", false);
+			TextInputComponent::new(env, "", "start typing..", false)
+				.with_input_type(super::InputType::Singleline);
 		find_text.embed();
 
 		Self {
--- a/src/components/log_search_popup.rs
+++ b/src/components/log_search_popup.rs
@@ -58,7 +58,8 @@ impl LogSearchPopupComponent {
 	///
 	pub fn new(env: &Environment) -> Self {
 		let mut find_text =
-			TextInputComponent::new(env, "", "search text", false);
+			TextInputComponent::new(env, "", "search text", false)
+				.with_input_type(super::InputType::Singleline);
 		find_text.embed();
 		find_text.enabled(true);
 
--- a/src/components/rename_branch.rs
+++ b/src/components/rename_branch.rs
@@ -98,7 +98,8 @@ impl RenameBranchComponent {
 				&strings::rename_branch_popup_title(&env.key_config),
 				&strings::rename_branch_popup_msg(&env.key_config),
 				true,
-			),
+			)
+			.with_input_type(super::InputType::Singleline),
 			branch_ref: None,
 			key_config: env.key_config.clone(),
 		}
--- a/src/components/stashmsg.rs
+++ b/src/components/stashmsg.rs
@@ -135,7 +135,8 @@ impl StashMsgComponent {
 				&strings::stash_popup_title(&env.key_config),
 				&strings::stash_popup_msg(&env.key_config),
 				true,
-			),
+			)
+			.with_input_type(super::InputType::Singleline),
 			key_config: env.key_config.clone(),
 			repo: env.repo.clone(),
 		}
--- a/src/components/tag_commit.rs
+++ b/src/components/tag_commit.rs
@@ -134,7 +134,8 @@ impl TagCommitComponent {
 				&strings::tag_popup_name_title(),
 				&strings::tag_popup_name_msg(),
 				true,
-			),
+			)
+			.with_input_type(super::InputType::Singleline),
 			commit_id: None,
 			key_config: env.key_config.clone(),
 			repo: env.repo.clone(),


pm100:
still would like to point out that one of the tests fails on windows due to a privilege issue. It would be nice if that test was cfg !windows

src/components/textinput.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(unused_imports)]
```
extrawurst, line 1: lets not

src/components/textinput.rs
```diff
@@ -14,39 +16,105 @@ use crate::{
...
+use std::{path::PathBuf, rc::Rc};
+use tui_textarea::{CursorMove, Input, Key, Scrolling, TextArea};
 #[derive(PartialEq, Eq)]
 pub enum InputType {
 	Singleline,
 	Multiline,
 	Password,
 }
+/*
+	completely rewritten using tui-textarea as it provides a ton of useful features
```
extrawurst, line 42: lets write comments as-is not as-compared-to-before. comments on the diff belong into the PR desciption

src/components/textinput.rs
```diff
@@ -14,39 +16,105 @@ use crate::{
...
+	- multi-char support
+
+
+
+
+=== key input ===
+
+All keys work as before.
+
+New line is shift-enter, ctrl-enter . These are all common new line editor commands (discord, emacs,...)
```
extrawurst, line 57: neither of the two work for me on mac-os. lets make them work & show up in the command list

src/components/textinput.rs
```diff
@@ -14,39 +16,105 @@ use crate::{
...
+
+There is no help for the editor window. The only thing a user really needs to know is the newline key stroke,
+but there are 10-15 ctrl key codes. I could add then to the general help popup, or make a special one for textinput
```
extrawurst, line 60: I like the idea of a specific help popup for the editor

src/components/textinput.rs
```diff
@@ -14,39 +16,105 @@ use crate::{
...
 	show_char_count: bool,
 	theme: SharedTheme,
 	key_config: SharedKeyConfig,
-	cursor_position: usize,
 	input_type: InputType,
 	current_area: Cell<Rect>,
 	embed: bool,
-	char_count: usize,
+	textarea: Option<TextAreaComponent>,
+	selecting: bool,
```
extrawurst, line 116: these two selection specific ones probably make sense to be an enum or struct and might fix the struct_excessive_bools-warning.

src/components/textinput.rs
```diff
@@ -57,20 +125,23 @@ impl TextInputComponent {
 		default_msg: &str,
 		show_char_count: bool,
 	) -> Self {
+		// this is here purely to make clippy happy
+		let eat_white_space = strings::symbol::WHITESPACE;
```
extrawurst, line 129: if this PR removes needing `strings::symbol::WHITESPACE` then lets remove it instead of this stunt

src/components/textinput.rs
```diff
@@ -84,14 +155,30 @@ impl TextInputComponent {
 
 	/// Clear the `msg`.
 	pub fn clear(&mut self) {
-		self.msg.clear();
-		self.update_count();
-		self.cursor_position = 0;
+		self.msg.take(); // = String::new();
```
extrawurst, line 158: comment not needed

src/components/textinput.rs
```diff
@@ -84,14 +155,30 @@ impl TextInputComponent {
...
+		if self.is_visible() {
+			self.create_and_show();
+		}
 	}
 
 	/// Get the `msg`.
+
+	// the fancy footwork with the OnceCell is to allow
```
extrawurst, line 166: lets put the comment (non doc comment) inside the fn

src/components/textinput.rs
```diff
@@ -109,120 +196,54 @@ impl TextInputComponent {
...
+			text_area
+		});
 	}
 
 	/// Set the `msg`.
 	pub fn set_text(&mut self, msg: String) {
-		self.msg = msg;
-		self.cursor_position = 0;
-		self.update_count();
+		self.msg.take();
```
extrawurst, line 241: instead of `take`+`set` we can just do `self.msg = msg.into();`, right?

extrawurst (changes requested):
a few nitpicks but ultimately the primary problem is: for me in none of the macos terminals (iterm2, Terminal, vscode embedded) the newline command works

pm100:
Its odd because I partially developed on mac so this shift-enter must be new-ish problem. Anyway there is a crossterm issue that I have reported (https://github.com/crossterm-rs/crossterm/issues/861). It is immediately resolvable by changing the startup code of gitui where it sets up crossterm. But that only works in some terminals (ITerm2 works, apple terminal does not).

I can also put back the ctrl-M code for newline that was there way back but one early reviewer said they didnt like it so I took it out.

I will wait a few days to see what crossterm has to say

src/components/textinput.rs
pm100: OK, I got fixated on keeping the changes to one file

src/components/textinput.rs
pm100: I thought there had to be a way to do it in one call.

pm100:
temp close while trying out fixes

pm100:
I have added ctrl-m back , meaing 'new line'

I have added the code to partially fix the mac issue, in my tests it works on iterm2, but not plain terminal. I dont think thats a good way to do it since its now not 100% clear to the user what will work when. The code is there so you can verify that it works.

Fixed everything else (pedantic is a pain, I think some things are less clear now, but whatever..)

Regarding adding help popup , extra key help etc... Can we get this big change in first. Everything works as before so nobody looses anything by not knowing the keystrokes

extrawurst:
@pm100 I went back through https://github.com/extrawurst/gitui/pull/509 which contains most of the considerations around this new-line command issue. lets go back to that original plan and use a new binding like `ctr+o` to commit and raw `enter` for the newline.

pm100:
I think thats a really bad idea, people will hate it

- its a breaking change
- its a breaking change to the most often used path through gitui ('gitui', 'a','c','xxxxx', 'enter')
- its a breaking change that is introduced solely to enable a feature that will not be used 99% of the time (multi line commit messages)
- its not obvious that ctrl-o means commit. It feels like 'open' to me
- its a breaking change caused by working around a bug in crossterm that only affects mac users using the native terminal

If you use my suggestion (ctrl-m for new line) then the feature is enabled, and nothing breaks, its an obscure key combination for use in a rare case. And maybe disable ctrl-enter on terminals where it doesnt work. 

Its your codebase but I guarantee you will annoy lots of users. Including me, because my muscle memory will still hit enter every time (just like I hit Esc every time to close a popup dialog because that seems natural - and it does close many popups, Annoys me every time it happens). I will be annoyed but at least I know the solution, others will not know what to do.

extrawurst:
i am unsure:

its been a long discussion prior. see old PR. 

its not a stable version of gitui yet, it will be documented in the cmd-bar. if anything can be broken its before 1.0.

Plus: your muscle memory will lead to no breaking change, you will insert a new-line, newcomers might be put of by the fact right now that enter is not a new-line and does a commit.

plus plus: you can re-bind the command if you want to.

pm100:
How about this
- I make it so that you can specify the key binding for new line (not doable at the moment )
-  I make it so there is only one binding for newline (there are multiple ones now, and your key config does not support that)
- I add to the bottom line of the text input box text saying something like 'ctrl-n=newline'
- I make the default ctrl-n (I dont like ctrl-M - its a default from tui-textarea but I override all those anyway, I think it comes from emacs)
- I think the commit = enter keystroke is also configurable, if not I will make it so.

Making it hard to do the no brainer thing is a really bad idea. But with the above changes people who want to make it hard to do the 99% case and easy to do the 1% case can do exactly that. Everybody happy.

pm100:
I have modified the PR so that all the changes I suggest are there, just so you can test it out and see

Here is the key config file for people who like it the hard way :-)
```
(
  multiline_enter: Some((code:Char('o') , modifiers:"CONTROL")),
  textbox_newline: Some((code:Enter, modifiers:"NONE")),
)
```

Includes fixing the textinput clients that should be single line (like branch name etc)

src/components/create_branch.rs
```diff
@@ -104,7 +104,8 @@ impl CreateBranchComponent {
 				&strings::create_branch_popup_title(&env.key_config),
 				&strings::create_branch_popup_msg(&env.key_config),
 				true,
-			),
+			)
+			.with_input_type(InputType::Singleline),
```
extrawurst, line 108: hm good catch, this was already an oversight before but now its getting to be a very small input field. let me see if I can fix that on master separately

extrawurst: Screenshot 2024-02-15 at 21 00 51

src/keys/key_list.rs
```diff
@@ -209,6 +211,8 @@ impl Default for KeysList {
 			view_submodule_parent: GituiKeyEvent::new(KeyCode::Char('p'),  KeyModifiers::empty()),
 			update_submodule: GituiKeyEvent::new(KeyCode::Char('u'),  KeyModifiers::empty()),
 			commit_history_next: GituiKeyEvent::new(KeyCode::Char('n'),  KeyModifiers::CONTROL),
+			multiline_enter: GituiKeyEvent::new(KeyCode::Enter,  KeyModifiers::empty()),
+			textbox_newline: GituiKeyEvent::new(KeyCode::Char('r'),  KeyModifiers::CONTROL),
```
extrawurst, line 215: lets call it just `newline`

src/main.rs
```diff
@@ -282,13 +286,29 @@ fn run_app(
...
+        )?;
+	}
 	io::stdout().execute(EnterAlternateScreen)?;
 	Ok(())
 }
 
 fn shutdown_terminal() {
 	let leave_screen =
-		io::stdout().execute(LeaveAlternateScreen).map(|_f| ());
+		stdout().execute(LeaveAlternateScreen).map(|_f| ());
```
extrawurst, line 311: lets leave it consistent with above, this change is not needed

src/main.rs
```diff
@@ -299,6 +319,15 @@ fn shutdown_terminal() {
 	if let Err(e) = leave_raw_mode {
 		eprintln!("leave_raw_mode failed:\n{e}");
 	}
+
+	let supports_keyboard_enhancement = matches!(
+		crossterm::terminal::supports_keyboard_enhancement(),
+		Ok(true)
+	);
+	if supports_keyboard_enhancement {
+		crossterm::queue!(io::stdout(), PopKeyboardEnhancementFlags)
```
extrawurst, line 328: i dont know yet how to reproduce this but I get this printed after closing gitui using this branch:

```
PopKeyboardEnhancementFlags failed%
```

how can we get rid of this?

src/components/textinput.rs
```diff
@@ -1,52 +1,55 @@
...
-
-/// primarily a subcomponet for user input of text (used in `CommitComponent`)
+#[derive(PartialEq, Eq)]
+enum SelectionState {
+	Selecting,
+	NotSelecting,
+	SelectionEndPending,
+}
+type TextAreaComponent = TextArea<'static>;
+#[allow(clippy::struct_excessive_bools)]
```
extrawurst, line 39: this allow is not needed anymore

src/components/textinput.rs
```diff
@@ -84,14 +86,30 @@ impl TextInputComponent {
...
-		self.update_count();
-		self.cursor_position = 0;
+		self.msg.take();
+		if self.is_visible() {
+			self.create_and_show();
+		}
 	}
 
 	/// Get the `msg`.
+
```
extrawurst, line 96: consistency: no empty line between doc comment and fn

src/components/textinput.rs
```diff
@@ -233,96 +182,44 @@ impl TextInputComponent {
...
-				self.get_msg(0..self.cursor_position);
-			let ends_in_nl = text_before_cursor.ends_with('\n');
-			txt = text_append(
-				txt,
-				Text::styled(text_before_cursor, style),
-			);
-			if ends_in_nl {
-				txt.lines.push(Line::default());
-			}
+		if self.is_visible() {
```
extrawurst, line 185: why recreate the whole textarea ?

src/components/textinput.rs
```diff
@@ -109,120 +127,51 @@ impl TextInputComponent {
...
 	}
 
 	/// Set the `msg`.
 	pub fn set_text(&mut self, msg: String) {
-		self.msg = msg;
-		self.cursor_position = 0;
-		self.update_count();
+		self.msg = msg.into();
+		if self.is_visible() {
+			self.create_and_show();
```
extrawurst, line 173: why recreate the whole textarea ?

src/components/textinput.rs
```diff
@@ -109,120 +127,51 @@ impl TextInputComponent {
...
-		while index > 0 {
-			index -= 1;
-			if !self.msg.is_char_boundary(index) {
-				continue;
-			}
-
-			let is_in_word = self.at_alphanumeric(index);
-			if was_in_word && !is_in_word {
-				return Some(last_pos);
+	fn create_and_show(&mut self) {
```
extrawurst, line 130: lets call it `create_inner_textarea`

src/components/textinput.rs
```diff
@@ -233,96 +182,44 @@ impl TextInputComponent {
...
 	}
 
-	fn draw_char_count<B: Backend>(&self, f: &mut Frame<B>, r: Rect) {
-		if self.char_count > 0 {
-			let w = Paragraph::new(format!(
-				"[{} chars]",
-				self.char_count
-			))
-			.alignment(Alignment::Right);
+	fn draw_newline_hint<B: Backend>(
```
extrawurst, line 212: Screenshot 2024-02-15 at 21 24 22

lets use the machinery that is already there to feed this new command into the cmd-bar on the bottom see `fn commands`

extrawurst (changes requested):
thanks for sticking with this. i think we are getting closer. i added a bunch of comments here and there.

src/main.rs
pm100: since we are not using ctrl/shift enter anymore I will remove all these changes, I dont really understand why they work, people can set it up using keyboard override if they want, but they will only do it if it works (ie not on a mac). Revisit if crossterm fixes their issue

src/components/textinput.rs
pm100: Dont go there :-). This code is the result of > 6 months of work arm wrestling with tui-textarea, your API into textarea (which I took as a given) and rust lifetime management.

src/components/commit.rs
```diff
@@ -565,7 +565,7 @@ impl Component for CommitComponent {
 			}
 
 			if let Event::Key(e) = ev {
-				if key_match(e, self.key_config.keys.enter)
+				if key_match(e, self.key_config.keys.multiline_enter)
```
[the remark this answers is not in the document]
pm100: Its a bit more fiddly than that . What should the behavior be for other multi-line edit boxes. For example 'reword commit'. I also wonder about stash name dialog, I changed it to be single-line but I think thats wrong, it can be a multi-line commit style message. Do all those behave the same way? If so (and I should check that they do work) them maybe 'commit' isnt the best name

src/components/textinput.rs
pm100: i missed that, will change

pm100:
OK. we are close :-)

I did everything you said. 

Notes: 

Should stash msg dialog be multi or single? I made it single but I think it should be multiple. Not sure tho. If multiple I need to change the plumbing around it too.

The defaults are Enter = commit, ctrl-r = newline. I have test both these

```
(
commit: Some((code:Char('o') , modifiers:"CONTROL")),
newline: Some((code:Enter, modifiers:"NONE")),
)
```
which is the one you asked for at the start

and

```
(
commit: Some((code:Enter , modifiers:"NONE")),
newline: Some((code:Enter, modifiers:"SHIFT")),
)
```

Which is what I originally had but doesnt work on a mac - as you found out.

PR #2047, opened 2024-02-13
Sign a commit if ssh secret key is provided

### LINKED ISSUE #1149: Sign git commits with SSH keys

yanganto:
This Pull Request fixes #1149

It changes the following:
- A ssh key path in `user.signingKey` of gitconfig  will be used to sign the commit 

I followed the checklist:
- [ ] I added unittests


extrawurst:
based on your last comment in the previous PR:

> Just hard to find another crate with the same function but using syn2. This is the same issue as the previous point.

still cannot follow what you refer to here. gitui now only depends on syn2, there is no old syn1 in the dependencies anymore.

what I am asking is: what is needed to make CI not fail?

>> I think we need some slightly simpler UX for providing this. Is there a standard way how this key is provided for example to git (the normal cli)? maybe env variables? or is it also purely via cli args?

> I will happy to use GITUI_SSH_KEY_PATH and GITUI_SSH_KEY_PASSWORD for the first iteration and moving forward from this.

I think we have some misunderstanding here. I am not saying lets introduce a bespoke env var that is gitui specific instead of a cli arg. My question is: how do other tools get fed the SSH key/pwd? Primarily the regular git cli and why can't we use the same way?

yanganto:
> based on your last comment in the previous PR:
> 
> > Just hard to find another crate with the same function but using syn2. This is the same issue as the previous point.
> 
> still cannot follow what you refer to here. gitui now only depends on syn2, there is no old syn1 in the dependencies anymore.
> 
> what I am asking is: what is needed to make CI not fail?

https://github.com/extrawurst/gitui/actions/runs/7794027383/job/21254692240
Yes, it was an issue in previous PR, but it is okay now.

> 
> > > I think we need some slightly simpler UX for providing this. Is there a standard way how this key is provided for example to git (the normal cli)? maybe env variables? or is it also purely via cli args?
> 
> > I will happy to use GITUI_SSH_KEY_PATH and GITUI_SSH_KEY_PASSWORD for the first iteration and moving forward from this.
> 
> I think we have some misunderstanding here. I am not saying lets introduce a bespoke env var that is gitui specific instead of a cli arg. My question is: how do other tools get fed the SSH key/pwd? Primarily the regular git cli and why can't we use the same way?

If the cli is good, I will pick these back and drop the `env::var` here.

The best practice for using an ssh password is to ask interactively with UI, such that the user can avoid writing down the password in a file or env.  And the default ssh key will be `~/.ssh/id_rsa` or `~/.ssh/id_ed25519`, which will depend on the way the user generates his key.  I think for the first iteration it is good to have a password here.  If you think this is bad, I can remove the password feature, because password is not a requirement for ssh key.

extrawurst:
> And the default ssh key will be ~/.ssh/id_rsa or ~/.ssh/id_ed25519, which will depend on the way the user generates his key. I think for the first iteration it is good to have a password here. If you think this is bad, I can remove the password feature, because password is not a requirement for ssh key.

Ok lets make things easy then and scan for the default keys, no arg or env for now.
Is there a git config that tells us that an ssh key is supposed to be used so we can show an error if we are supposed to use ssh and cannot find a key? 

also if we know we shall encrypt via ssh and we find a default key do we know if that key requires a pwd? so that we can fail for the MVP in that case because we don't know the pwd

extrawurst:
@yanganto but mostly we need to get the CI green

yanganto:
> > And the default ssh key will be ~/.ssh/id_rsa or ~/.ssh/id_ed25519, which will depend on the way the user generates his key. I think for the first iteration it is good to have a password here. If you think this is bad, I can remove the password feature, because password is not a requirement for ssh key.



> 
> Ok lets make things easy then and scan for the default keys, no arg or env for now. Is there a git config that tells us that an ssh key is supposed to be used so we can show an error if we are supposed to use ssh and cannot find a key?

You misunderstand the mechanism.  The commit can be signed with any ssh secret key, but only one.  Github will use verify the commit only if the corresponding pub key is already set in Github by user.  So we can not scan all the ssh key to select some one who is valid, because all key are valid and possible in use.  The default key path will be different because user can use different kinds of crypto mechanism.  We can have a default value as now, but we need a interface let user change it easier.
>
 
> also if we know we shall encrypt via ssh and we find a default key do we know if that key requires a pwd? so that we can fail for the MVP in that case because we don't know the pwd

No, if we do not have a password, we still can sign the commit but it will not verified by GITHUB, because the bytes will not be different from a real one.

extrawurst:
> You misunderstand the mechanism

then lets go back to square one: how does it work with the git-cli, how does it get configured and why cant we use the same method?

extrawurst:
as far as i can see it has to be configured like this: https://docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key#telling-git-about-your-ssh-key

so we can just use these git configs to know what signing key to use and whether signing is supposed to happen: so lets use that

yanganto:
git commit use following option to sign a commit with GPG
`git-commit -S[<keyid>], --gpg-sign[=<keyid>]`

And the ssh sign is using ssh secret key but the header still `gpgsig`, and GITHUB or other repository will using the pub key.

There is not exactly an interface in git-cli for the ssh key.  The way we are using now is `-s`.  I think it is good enough now.

yanganto:
> as far as i can see it has to be configured like this: https://docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key#telling-git-about-your-ssh-key
> 
> so we can just use these git configs to know what signing key to use and whether signing is supposed to happen: so lets use that

You can see the config is a pub key for verification not a secret key for a sign.  Do we have an assumption that the secret key is always the pub key path remove `.pub`?

yanganto:
By the way, the way to set git config with `gpg.format ssh` will not work as what you see on the web page. 

```
└─[$] <git:(ssh*)> g commit -m 'test ssh with git'
error: unsupported value for gpg.format: ssh
fatal: bad config variable 'gpg.format' in file '.git/config' at line 28
└─[$] <git:(ssh*)> git --version
git version 2.32.0

```

yanganto:
You may wonder why we need `Clone` trait on `CliArgs`
Because clippy will cry on passing too many fields of `CliArgs` in `run_app` function.
https://github.com/extrawurst/gitui/actions/runs/7888377158/job/21525757826

yanganto:
@extrawurst  It now is ready to review

extrawurst:
> You can see the config is a pub key for verification not a secret key for a sign.

this is exactly what every source I find about ssh-commit-signing tells you to do. to sign with your public key. what am I missing?

but don't take my word for it, see [Gitlab docs](https://docs.gitlab.com/ee/user/project/repository/signed_commits/ssh.html#configure-git-to-sign-commits-with-your-ssh-key), [tower (slightly popular git GUI app)](https://www.git-tower.com/blog/setting-up-ssh-for-commit-signing/) and this high ranking blog post [here](https://blog.dbrgn.ch/2021/11/16/git-ssh-signatures/)

yanganto:
> > You can see the config is a pub key for verification not a secret key for a sign.
> 
> this is exactly what every source I find about ssh-commit-signing tells you to do. to sign with your public key. what am I missing?
> 
> but don't take my word for it, see [Gitlab docs](https://docs.gitlab.com/ee/user/project/repository/signed_commits/ssh.html#configure-git-to-sign-commits-with-your-ssh-key), [tower (slightly popular git GUI app)](https://www.git-tower.com/blog/setting-up-ssh-for-commit-signing/) and this high ranking blog post [here](https://blog.dbrgn.ch/2021/11/16/git-ssh-signatures/)

A good sign mechanism is to use a secret key and pass the public key to verify whether it is true or not.  If you sign with a public key, then you need to pass the secret key to Github for verification. :roll_eyes: 
The commit in the current PR is signed by my SSH secret key with this code, and I set the public key on GitHub, so you can there is a green verified flag.

The blog post [here](https://blog.dbrgn.ch/2021/11/16/git-ssh-signatures/) is signed with a secret key, That is the same as what I did in this PR.

The ssh setting will need a git after `2.34`.  The legacy git will be broken when using it, also there is a lot of possible value in `signingKey`, the path to the pub key, the path to the secret key, the content of the SSH secret key, or possibly some content of gpg keys.  

Does current gitui handle a global git config and the local git config?  I dont think the first iteration need to handle these variations.

extrawurst:
> The blog post [here](https://blog.dbrgn.ch/2021/11/16/git-ssh-signatures/) is signed with a secret key, but he put the content of key in the signingKey, That is the same as what I did in this PR.

It literally says in the link above `Simply use your public key`

> The ssh setting will need a git after 2.34.

Fine with me.

Lets use `signingKey` config no cli param for now. if user wants to use ssh singing `signingKey` and `gpg.format ssh` have to be configured. 

> Does current gitui handle a global git config and overlayed with a local git config

yes `git2-rs` (libgit2) make that pretty easy

yanganto:
Hi @extrawurst 
I believe this is good to go.

asyncgit/Cargo.toml
```diff
@@ -14,6 +14,7 @@ keywords = ["git"]
 [dependencies]
 bitflags = "2"
 crossbeam-channel = "0.5"
+dirs = "5.0"
```
utkarshgupta137, line 17: You can use the `home` crate instead, it is a much smaller dependency.

extrawurst: its already a gitui dependency, switching to home instead would be a separate PR

extrawurst:
@yanganto can you rebase again to make it up to date?

yanganto:
Hi @extrawurst 
The master is merged into this branch, so the sign can be retained. :smile:

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ These defaults require some adoption from existing users but feel more natural t
 * support `prepare-commit-msg` hook ([#1873](https://github.com/extrawurst/gitui/issues/1873))
 * support for new-line in text-input (e.g. commit message editor) [[@pm100]](https://github/pm100) ([#1662](https://github.com/extrawurst/gitui/issues/1662)).
 * new style `block_title_focused` to allow customizing title text of focused frame/block ([#2052](https://github.com/extrawurst/gitui/issues/2052)).
+* support sign commit with ssh key by `user.signingKey` of gitconfig  ([#1149](https://github.com/extrawurst/gitui/issues/1149))
```
cmoog, line 25: Is my reading of the code correct that this will only support signing keys written to disk? If so, perhaps the change log should note that signing via ssh-agent/`SSH_AUTH_SOCK` isn't yet supported.

asyncgit/Cargo.toml
```diff
@@ -27,6 +28,7 @@ rayon = "1.8"
 rayon-core = "1.12"
 scopetime = { path = "../scopetime", version = "0.1" }
 serde = { version = "1.0", features = ["derive"] }
+ssh-key = { version = "0.6.4", features = ["crypto", "encryption"] }
```
extrawurst, line 31: is there any potential to safe on bloat by reducing features via `default-features = false` ?

asyncgit/src/sync/commit.rs
```diff
@@ -65,6 +67,38 @@ pub fn commit(repo_path: &RepoPath, msg: &str) -> Result<CommitId> {
 	scope_time!("commit");
 
 	let repo = repo(repo_path)?;
+	let ssh_secret_key = {
```
extrawurst, line 70: lets put that into a separate `fn` - lets also make sure we are not running this if the `gpg.format` is not `ssh`

extrawurst (changes requested):
this does not cover the other places that need signing yet:
* amend
* commit_revert
* merge_commit
* reword
* tag_commit

(hope I did not forget)

asyncgit/Cargo.toml
yanganto:
```
default = ["ecdsa", "rand_core", "std"]
crypto = ["ed25519", "p256", "p384", "p521", "rsa"]
encryption = [
    "dep:bcrypt-pbkdf",
    "alloc",
    "cipher/aes-cbc",
    "cipher/aes-ctr",
    "cipher/aes-gcm",
    "cipher/chacha20poly1305",
    "rand_core"
]
```
The only different is `std` and I do not think we need to disable default-feature here.

asyncgit/src/sync/commit.rs
yanganto: I will refuse to do this, because 
- Git version 2.32 with `gpg.format` `ssh` setting will be broken
- GPG will not use a path to a key file in that field, so it will not conflict here.

We can have more compatibility in this way without being a pain for people using different versions of git.

extrawurst: `gpg.format` was introduced in [2.19](https://github.com/git/git/blob/96c8a0712e569dd2812bf4fb5e72113caf326500/Documentation/RelNotes/2.19.0.txt#L57) (5 years ago)

`ssh` support was introduced in [2.34](https://github.com/git/git/blob/96c8a0712e569dd2812bf4fb5e72113caf326500/Documentation/RelNotes/2.34.0.txt#L81) (3 years)

I would challenge whether this is an important audience to cater for

extrawurst (changes requested):
i am open for the argument of not having to check the `gpg.format` if there is a unambiguous method to figure out what signing mechanism the user chooses (the most common ones being: ssh, gpg).

that being said please checkout my comment: https://github.com/extrawurst/gitui/pull/1544#issuecomment-1954431049

we should agree on some abstraction that we call to sign buffers that then allows us to delegate the signing to various paths: shelling out for gpg, using your implementation for ssh.

this abstraction should be called in all the places that require potentially signing so that whatever we do we never forget to support all relevant code paths for signing (like all the places missing in this PR) 

so actionpoints:
* lets sync up with @hendrikmaus on this
* potentially extract the abstraction that he build already in his PR into a separate PR where GPG is not implemented, just so we have the abstraction available also for this one

yanganto:
KISS (Keep it simple and stupid) will be the best minuteness way I think.

This already is extendable to another kind of sign and is also welcome to refactor in the future.
```rust
	let id = match (
		config.get_entry("gpg.format").ok(),
		fetch_ssh_key(&config),
	) {
		(Some(f), Some(sk)) if f.value() == Some("ssh") => {},
		_ => (),
	}
```

The `Sign` trait in #1544 is beautiful and much more abstract, but it already takes more than one year.  I hope we will not take another year from now.  Also, I like to adapt the implementation of the `Sign` trait is already in the mainstream.

We are building a daily tool for developers and not a rocket to Mars.  I believe having a good release pace, iterating, and easier customization for users will be great for this project.

extrawurst:
So far this project was developed with high standards and I believe this is what makes it successful. I won’t stop now.

merging signing in any shape is a big responsibility and I am open to do it without the abstraction in order to refactor later but then we still need to make it work in all places that require signing to not confuse users that trust the quality of gitui

yanganto:
Hi @thePanz 

Remove this setting that will work.
```
[commit]
    gpgsign = true
```

And I am surprised to see this restriction.  That means if a user uses other git tools with `gpgsign` at the same time, `gitui` will refuse to work. 🤨

Anyway, this is fixed now.

extrawurst:
@yanganto like you argued earlier we should not require the user to make changes in their config that works with vanilla git cli

yanganto:
Yes, so I fixed this.  I am just really surprise we have the disable alert, I through we did not have something like that.

yanganto:
Hi @thePanz 
After the fix, you don't need to change your gitconfig.  The previous commit is also signed with the same config as yours and I put it in the global.

yanganto:
If I can design the behavior for the alert,  I will just let the commit box in red with the title `unsigned commit` prompt the user but not restrict the user to do that.  If @extrawurst you think this is a good one, please open an issue on it.

extrawurst:
@thePanz please test it again

thePanz:
@yanganto @extrawurst tested again:
1. the committing is now working, no error message is reported
2. the signature is not attached the commit tho

See the screenshot:
image

- the first commit (`54d5f58`) is done with gitui
- the second commit (`86aba81`) was done with `git commit -m " Testing commit from git CLI"`

the graph is obtained with `git tree --show-signatures`

yanganto:
@extrawurst What is the next action item need to be done in this PR?

yanganto:
Hey there is no error here the sign is not attached because you are not setting the `allowedSignersFile`

if you do not set it, what you will see in the commit

```
$ git --no-pager log -n 2 --show-signature
commit 083294c5c6606f2e0e5d35f4707c8c268d6fc7fb (HEAD -> ssh-sign-2, origin/ssh-sign-2)
error: gpg.ssh.allowedSignersFile needs to be configured and exist for ssh signature verification
No signature
Author: Antonio Yang <yanganto@gmail.com>
Date:   Thu Feb 22 21:30:31 2024 +0800

    fix commit.gpgsign=ture allert when sshsign set

commit 1f53a1cfb848f5947715ff43f34e70379a84db0e
...
```

If you correctly setup the `allowedSignersFile` like this
```
[gpg "ssh"]
    allowedSignersFile = /home/yanganto/.config/git/allowed_signers
```
```
$ cat ~/.config/git/allowed_signers
yanganto@gmail.com ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQDQuRs44Ry//RaVddE3ZxR7wKMM7h8BAmUpud3urpGQ3LbxrhbCGKQoRV1OJEEdnYSoAk4bMvtEfkkXFZAioSJxZ0Flult983LHWCCB0pYkq/Hw0UqgLZV2TjyYIqLxAlhDh7vbivmlP0nDPQ3dV33GrIcPK9zShyI58JWJTna/Lf0XsLN2Iy72xo15R8FnwVL+NtCIVBSeP5huYVy8Z65Zv+Tcg8vXQCL2raNqwNE/B1NQary5vgfId76Q6w5gyvDXWZITskNOmmbmB5aZ/1mxWh1m16BsxO61upchopjIN/Wh6meipt3EY5FEibJp2GLUEQLEoDcZ6hvtZKmS7vhiroHAWb+u2MM/rJnNmIfVVRp2MlDd6wWfb9ZwpGb1YypS5qYAIocsjlMfGnonRW9ZgA/sWbWhp4aLKdncpyxw8qCPnEEf2E+mKmVCCRabhjs1yoZ/e99r2sJUruWF4udHjMOIVietMS166npRPIdOkz/23M/yL8Z7b7BfP0IkUq8= yanganto@yanganto-laptop
```

and you will see this correctly.
```
└─[$] <git:(ssh-sign-2*)> git --no-pager log -n 2 --show-signature
commit 083294c5c6606f2e0e5d35f4707c8c268d6fc7fb (HEAD -> ssh-sign-2, origin/ssh-sign-2)
Good "git" signature for yanganto@gmail.com with RSA key SHA256:DJqULdMeNLAu43jKUw50bB7b11X1lbu+CXkj15lqOZo
Author: Antonio Yang <yanganto@gmail.com>
Date:   Thu Feb 22 21:30:31 2024 +0800

    fix commit.gpgsign=ture allert when sshsign set

commit 1f53a1cfb848f5947715ff43f34e70379a84db0e
....
```

That is not my problem

thePanz:
@yanganto from my screenshot you can see the two top most commits that shows the issue.
That's not related to the `allowedSignersFile`, as the top most commit is showing that git is recognizing the "good" signature from my user.

In the screenshot, have a look at the commit hashes:

- `86aba81` was done with git commit -m " Testing commit from git CLI", the signature is there and valid for `thapanz+github@gmail.com`
- `54d5f58`  is done with gitui, but no signature is present

thePanz:
> What is the thing in your `allowedSignersFile`? Did it auto-append something there?

in my `.gitconfig` I have:
```
[gpg "ssh"]
        allowedSignersFile = ~/.ssh/allowed_signers
```

Is there anything I should check for you?

yanganto:
What is the content of ` ~/.ssh/allowed_signers`?  This is the thing you should provide to GITHUB, else it not really work on the internet I think.

yanganto:
> I'm fairly certain `allowed_signers` shouldn't need to exist. Other programs don't rely on it and can still properly sign commits.
Is the sign an ssh sign or gpg sign?
> 
> vscode nor any jetbrains product needs it in order to sign a commit.
We don't need `allowed_signers` to sign a commit, but it is for display.

thePanz:
I am using ssh signing

yanganto:
What is the content of `~/.ssh/allowed_signers` from your end? @thePanz

thePanz:
I guess the contents of the allowed_signers is not relevant to this issue, as I show in the screenshot: when the signature is created by the git CLI, it works as expected.

supleed2:
I'm glad to see the work done so far! Just wanted to add a datapoint.

Compiled from this commit: https://github.com/yanganto/gitui/commit/43bea2106d302241f522f03a1893a8d55468775b, I ran `git init`, created a file and added a remote, then started `gitui`, committed and pushed giving this signed commit on GitHub: https://github.com/supleed2/gitui-test/commit/62429dafbed4448b7a79688cf48b99072e9e2eff

Seems like it works fine?

thePanz:
I will test it again @yanganto , I forked @supleed2 test repo on my org and start pushing commits there

ps: I am not asking to chase and fix a bug, I am trying to check if this feature can be used with my workflow without altering any GIT config file

yanganto:
If you do not put the ssh key in the git config, there is no alert and no sign.

If you put the ssh_key settings in the git config (no allowedSignersFile), you can sign without an alert.

If you put the allowedSignersFile and the public key to the file, you can check the sign locally without an alert.

If you want something you need to provide the correct setting for the tool. It is reasonable.

yanganto:
Are these settings in your gitconfig?
```
[user]
	email = *
	name = *
	signingKey = /path/to/keyfile
[gpg]
	format = ssh
[commit]
	gpgsign = true
[gpg "ssh"]
	allowedSignersFile = /path/to/allowed_signers
```
Can you also provide this?

Aggreed, the `allowedSignersFile` will not impact on the signing process/

thePanz:
> Can you also provide this? Do you put your pub key on GITHUB?

yes, as you can see from https://github.com/thePanz/gitui-test/commits/master/, the commit I did with the git CLI is verified

yanganto:
Is There a key pair ` ~/.ssh/github.pub` and ` ~/.ssh/github` on your machine?

thePanz:
No, there is no `~/.ssh/github` on my machine, the private key is provided via the `git-credential-libsecret` to GIT.

I guess we spotted the issue then :+1: 

I believe that a message or warning should be shown in that case, as the git configuration is explicitly asking to add the signature, while gitui is ignoring the missing private key needed to sign it and respect the configuration.

This is not 100% clear from the changelog IMO

yanganto:
So now you are asking for an alert box with a warning message for this first implementation.
I think it should be another feature about talking to the store, but not an alert box at this PR. If you buy in the current PR, and list this as a known issue. 

Or, another solution is.
We can refuse to commit if I can not get the private key.  Just like current behavior refuse to commit, if there is a sign setting in gitconfig.

thePanz:
@yanganto I am not asking to do it, I am suggesting to have such behavior as it would help the users to better understand why the signing is required in the git config file, but gitui is adding the commit without the signature (thus ignoring the git config setting)

just my 2 cents

yanganto:
@thePanz I don't see any blocker at this moment.  Now anything gitconfig set but gitui cannot do will block which follows the current implementations. 

If you agree with my point and try to warn the user not to block the user you can open an issue and mention these https://github.com/extrawurst/gitui/pull/2047#issuecomment-1959452622 https://github.com/extrawurst/gitui/pull/2047#issuecomment-1959485605. These are already clear enough to solve in other PRs but not this one.

extrawurst (changes requested):
please update to latest master and integrate with the new signing trait

extrawurst:
closed via #2175

PR #2051, opened 2024-02-15
tui textarea

pm100:
new PR for  #1662 - the old one was getting cluttered

OK. we are close :-)

I did everything you said. 

Notes: 

Should stash msg dialog be multi or single? I made it (and I see you did too) single but I think it should be multiple. Not sure tho - isnt it another commit style message? If multiple I need to change the plumbing around it too.

The defaults are Enter = commit, ctrl-r = newline. 

I have tested both these key configs

```
(
commit: Some((code:Char('o') , modifiers:"CONTROL")),
newline: Some((code:Enter, modifiers:"NONE")),
)
```
which is the one you asked for at the start

and

```
(
commit: Some((code:Enter , modifiers:"NONE")),
newline: Some((code:Enter, modifiers:"SHIFT")),
)
```

Which is what I originally had but doesnt work on a mac - as you found out.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,2 +1,0 @@
-[advisories]
-ignore = []
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -200,13 +200,9 @@ jobs:
       run: |
         cargo install --locked cargo-deny
 
-    - name: cargo-deny licenses
+    - name: cargo-deny checks
       run: |
-        cargo deny check licenses
-
-    - name: cargo-deny bans
-      run: |
-        cargo deny check bans
+        cargo deny check
 
   udeps:
     name: udeps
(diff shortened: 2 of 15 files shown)


pm100:
One topic not discussed is the cut and paste behavior.

- tta today does not support clipboard. Long, long debate about how it should work is still open there. https://github.com/rhysd/tui-textarea/issues/50
- tta supports its own yank buffer. Which is probably not very useful here (we are not vim). So none of those key strokes are wired up at the moment in gitui
- paste from OS clipboard works in most cases - in fact its hard to get ctrl-v not to work - ie to get it passed to the app. The terminal just dumb types the keystrokes, tta does not see it as a paste.
- I know the internals of tta and they expose enough bits so that I can make cut and copy copy to the clipboard 
- you intercept ctrl-c to mean shutdown now. It might be better to suspend that while tta is shown in order to allow 'copy'

For now, lets wrap this up as is. I will raise a new issue for this

extrawurst:
@pm100 thanks for keeping on this one! its good to go. i might change the default, I still feel enter should be default-newline. but now everything can be configured however you like it

Issue #2057, opened 2024-02-18
commit text box does not support copy/cut /paste


pm100:
The new commit dialog does not support copy/cut/paste to and from the system clipboard.

PR #2058, opened 2024-02-19
Add support for copy/cut/paste to textinput box

pm100:
PR for #2057 

Notes:
- I added arboard for clipboard read (I need read) and write. I should probably replace the clipboard read code that you already have
- arboard uses a license that you dont approve in deny.toml
- I added key defs , they default to ctrl-c, ctrl-x, ctrl-v
- for ctrl-c I added code to turn off ctrl-c = exit while the textbox is displayed

Paste is complicated
- on windows no matter how hard you try you cannot get windows terminal to not capture ctrl-v, there are bugs open to MSFT on it but nobody has done anything about it
- windows then just types the text into the screen for you. Thats mainly OK but not in the case where you try to paste multiple lines. In the default gitui configuration, where Enter means commit, when windows types the newline (when pasting multiple lines) gitui commits . Which is a surprise and it took me a while to work out what was going on. 
- This is an issue without this PR. Pasting multiple lines using the OS functionality - same happens on a mac using command v
- this PR enters the text, including line breaks into the textarea, if it gets the key combination
- command - v still blindly types into the box on mac, ctrl-v works fine
- ctrl-v on windows does the same thing (blind type) - I have tested that ctrl-p (in key config) works fine.
- on linux ctrl-v seems to work fine, the PR sees the keystroke and does the right thing

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ keywords = ["git", "gui", "cli", "terminal", "ui"]
 
 [dependencies]
 anyhow = "1.0"
+arboard = "3.3.1"
 asyncgit = { path = "./asyncgit", version = "0.24", default-features = false }
 backtrace = "0.3"
 bitflags = "2.4"
--- a/src/app.rs
+++ b/src/app.rs
@@ -54,9 +54,10 @@ use std::{
 	cell::{Cell, RefCell},
 	path::{Path, PathBuf},
 	rc::Rc,
+	sync::atomic::AtomicBool,
 };
 use unicode_width::UnicodeWidthStr;
-
+pub static ENABLE_HARD_EXIT: AtomicBool = AtomicBool::new(true);
 #[derive(Clone)]
 pub enum QuitState {
 	None,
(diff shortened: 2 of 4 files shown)


extrawurst:
i wont merge this for now, this is has too many caveats for very little gain:

**pasting** works fine in terminal anyway, the default key for newline since 825935b is `enter`, so it wont commit on pasting newlines.

**copy** works too in most terminals by just selecting the text with the mouse.

pm100:
I messed up in my comments. The paste thing was totally off, its not an issue with the PR, its an issue without it, and only when you have non default settings (I forgot that I had gitui set up that way) - so nothing to do with the PR

PR summary

- edit_copy, edit_cut, edit_paste work on all platforms
- the keys are configurable, defaults are ctrl-c, ctrl-x, ctrl-v
- only odd one is that ctrl-v (the default paste) is intercepted by the OS on windows and it does the paste, which is fine. Only oddity is if its multiple lines with enter=commit non-default setting. But they are in that situation with or  without this PR .

2 things about not having this PR:
- Mouse driven  OS terminal copy does not work if you select multiple lines
- Keyboard users (everybody to start with given that gitui is a kb driven app) might expect to be able to ctrl-c to copy , given that they can select text. This will instantly exit the app

PR #2075, opened 2024-02-20
fix infinite spinner

extrawurst:
af9da95 introduced pending work to always be true


extrawurst:
cc @tdtrung17693

extrawurst:
cc @tdtrung17693

Issue #2007, opened 2024-01-12
Theme file not loading on macOS


maxhoffmann:
**Describe the bug**
When I override the default theme or try to add a new one, it doesn’t affect `gitui`’s theme.

**To Reproduce**
1. Install `gitui` via `brew`
2. Add a theme at `~/.config/gitui/theme.ron`
3. Example theme:
```ron
(
    line_break: Some(""),
    selection_bg: Some(LightBlue),
)
```
4. Open `gitui`
5. Colors are still the same and new line operator is still visible.

**Expected behavior**
I expect the theme to change.

**Context (please complete the following information):**
 - OS/Distro + Version: macOS `13.6.3`
 - GitUI Version `0.24.3`

**Additional context**
It would help to see where themes are loaded from, in case gitui doesn’t use `~/.config/gitui/`. Trying to add a custom theme and loading it via `-t` also doesn’t lead to any error. It silently fails, so I cannot debug where the `gitui` loads the themes from.

extrawurst:
just tried and on master aswell on the version you use via brew exactly your theme changes work as expected.

maxhoffmann:
@extrawurst Is there any way to tell gitui that it should log which themes it loads and in which folder it looks? This would really help debugging this on my machine.

extrawurst:
please run with `-l` to enable logging, see https://github.com/extrawurst/gitui#diagnostics

and share the log created. also helpful is checking if `echo $HOME` and `echo ~` return the right user folder where you store the config

extrawurst:
@maxhoffmann still an issue? any findings via the log option?

maxhoffmann:
@extrawurst unfortunately yes. there are no entries in the log file about loading a config. `echo $HOME` and `echo ~` return the correct path

extrawurst:
@maxhoffmann can you please try to build from source? in be10d9096a93b030ec3f1ab776a21225d974e272 I added logging theme loading errors to the logfile

extrawurst:
@cgahr please try again with current master (f0f90a8) which should also print the path of the theme that errors

cgahr:
I can confirm the `-t <path>` and `--theme <path>` both have no effect.

To reproduce:
1. remove last `)` from `~/.config/gitui/theme.ron` so that the resulting file is not a valid configuration
2. run `./target/debug/gitui -l -t doesnotexist`

The relevant line in the log file is 
```
[...]

PR #2077, opened 2024-02-20
better theme file handling

extrawurst:
related to #2007


extrawurst:
now help says this:

```
Usage: gitui [OPTIONS]

Options:
  -t, --theme <THEME_FILE>     Set color theme filename loaded from config directory [default: theme.ron]
```

plus it will not silently fallback to the default theme file name if the provided one does not work. it will try the one provided and print an appropriate error to the log

PR #2086, opened 2024-02-21
Remove references to bit flags

mo8it:
BTW: My old config with bit flags stopped working after the update to 0.25. This should be listed as a breaking change in the CHANGELOG and release info.


extrawurst:
good catch, thanks for adopting the docs

extrawurst:
@mo8it updated the changelog: https://github.com/extrawurst/gitui/commit/7bf65f483594c56aac883e9ac6e17d8d9ee33fd9

Issue #2098, opened 2024-02-23
`cargo install gitui` fails  without `--lock` or lockfile: gitui v0.25.1


nuke-web3:
**Describe the bug**
Install fails (expand for logs)
<details>

```sh
 Compiling gitui v0.25.1
error[E0308]: mismatched types
    --> /home/nuke/.cargo/registry/src/index.crates.io-6f17d22bba15001f/gitui-0.25.1/src/components/textinput.rs:151:28
     |
151  |                 .set_cursor_line_style(self.theme.text(true, false));
     |                  --------------------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `ratatui::style::Style`, found a different `ratatui::style::Style`
     |                  |
     |                  arguments to this method are incorrect
     |
     = note: `ratatui::style::Style` and `ratatui::style::Style` have similar names, but are actually distinct types
note: `ratatui::style::Style` is defined in crate `ratatui`
    --> /home/nuke/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ratatui-0.24.0/src/style.rs:192:1
     |
192  | pub struct Style {
     | ^^^^^^^^^^^^^^^^
note: `ratatui::style::Style` is defined in crate `ratatui`
    --> /home/nuke/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ratatui-0.26.1/src/style.rs:227:1
     |
227  | pub struct Style {
     | ^^^^^^^^^^^^^^^^
     = note: perhaps two different versions of crate `ratatui` are being used?
note: method defined here
    --> /home/nuke/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tui-textarea-0.4.0/src/textarea.rs:1770:12
     |
1770 |     pub fn set_cursor_line_style(&mut self, style: Style) {
     |            ^^^^^^^^^^^^^^^^^^^^^

error[E0308]: mismatched types
    --> /home/nuke/.cargo/registry/src/index.crates.io-6f17d22bba15001f/gitui-0.25.1/src/components/textinput.rs:154:5
     |
153  |               text_area.set_placeholder_style(
     |                         --------------------- arguments to this method are incorrect
154  | /                 self.theme
155  | |                     .text(self.selected.unwrap_or_default(), false),
     | |___________________________________________________________________^ expected `ratatui::style::Style`, found a different `ratatui::style::Style`
     |
[...]

PR #2101, opened 2024-02-25
ratatui 0.25 update

extrawurst:


extrawurst:
great now we are blocked by this: https://github.com/rhysd/tui-textarea/issues/60

extrawurst:
this will fix #2098

PR #2120, opened 2024-03-10
Make MsgPopup scrollable

### LINKED ISSUE #1138: View arbitrarily long multi-line annotations for git tags
Happily, [we can view git tag annotations](https://github.com/extrawurst/gitui/issues/1076), including annotations with multiple lines. Sadly, if the annotation is long and goes on for many lines, we can't see the whole thing. In my current terminal emulator, gitui only displays the first 23 lines.

This issue will increase in importance if we [support creating multiple-line annotations for git tags](https://github.com/extrawurst/gitui/issues/1137).

Adding a scrollbar seems like a good solution to me.

If adding scrollbars to a "popup window" is not desirable, perhaps refactoring the UI so that annotations can be viewed in a multi-panel interface, like the files list, might work better.

**Additional context**
Screenshot of gitui with a long git tag annotation and no way to scroll to read the rest of it

MichaelAug:
This Pull Request fixes/closes #1138.

It changes the following:
- Adds scrolling for tag annotations by adding scrolling to MsgPopups which are used to display tag annotations.

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 * re-enable clippy `missing_const_for_fn` linter warning and added const to functions where applicable ([#2116](https://github.com/extrawurst/gitui/issues/2116))
-* Make info and error message popups scrollable ([#1138](https://github.com/extrawurst/gitui/issues/1138))
+* Make info and error message popups scrollable [[@MichaelAug](https://github.com/MichaelAug)] ([#1138](https://github.com/extrawurst/gitui/issues/1138))
 
 ## [0.25.1] - 2024-02-23
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 * re-enable clippy `missing_const_for_fn` linter warning and added const to functions where applicable ([#2116](https://github.com/extrawurst/gitui/issues/2116))
+* Make info and error message popups scrollable ([#1138](https://github.com/extrawurst/gitui/issues/1138))
 
 ## [0.25.1] - 2024-02-23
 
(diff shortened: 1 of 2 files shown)


extrawurst:
somehow i cannot get this to work if I checkout your branch:


https://github.com/extrawurst/gitui/assets/776816/dd2a8f0a-808d-4fb9-96df-30ed8db97d76

MichaelAug:
The scrolling should work correctly now when shrinking the window

extrawurst:
it only uses the original messages lines count and not the number of lines that happen to be shown when the text wraps at places because it is wider than the message box:
Screenshot 2024-03-19 at 18 16 07

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 * re-enable clippy `missing_const_for_fn` linter warning and added const to functions where applicable ([#2116](https://github.com/extrawurst/gitui/issues/2116))
+* Make info and error message popups scrollable ([#1138](https://github.com/extrawurst/gitui/issues/1138))
```
extrawurst, line 13: also please add your github handle (see others)

src/popups/msg.rs
```diff
@@ -22,39 +27,70 @@ pub struct MsgPopup {
...
-		let mut width = u16::try_from(max)
-			.expect("can't fail due to check above");
-		// dont overflow screen, and dont get too narrow
-		if width > f.size().width {
-			width = f.size().width;
-		} else if width < 60 {
-			width = 60;
-		}
+			.max()
+			.expect("Expected at least one line in the message")
```
extrawurst, line 52: i know it should not happen but I'd rather not crash if it does that an empty message box would be shown.

CHANGELOG.md
MichaelAug: Fixed

src/popups/msg.rs
MichaelAug: Instead of panic will now return 0

MichaelAug:
> it only uses the original messages lines count and not the number of lines that happen to be shown when the text wraps at places because it is wider than the message box: Screenshot 2024-03-19 at 18 16 07

It should now use the correct width for text wrapping

extrawurst:
looks good now! thanks @MichaelAug

PR #2145, opened 2024-03-22
Git Config Commit Comments

### LINKED ISSUE #2136: Commit messages that begin with a hash (#) are ignored
**Describe the bug**
When using gitui to create a commit, commit messages that start with a # are ignored. This is the default behaviour of git however setting the config value `core.commentChar` makes the commit message work when using git on the command line, however commiting through gitui the commit message is still stripped out

**To Reproduce**
1. Go to a project and make some changes
2. Open git UI
3. Commit the changes
4. In the commit message dialog, use the message "#5123 test"
5. Go to the log and the message for that commit is blank
6. Make more changes
7. Run `git config --global core.commentChar ";"``
8. Repeat the above steps
9. Same outcome

**Expected behavior**
The behaviour should match commiting with the command line, where changing the core.commentChar is respected

**Screenshots**
image
image
image

 - OS/Distro + Version: Windows 11, 23h2, 22631.3296
 - GitUI Version: 0.24.3
 - Rust version: 1.75.0

Concelare:
This Pull Request fixes/closes #2136.

It changes the following:
- Commits will now follow the git config `core.commentChar` if one is not set it defaults to `#`

Screenshot Evidence
image

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/asyncgit/src/sync/commit.rs
+++ b/asyncgit/src/sync/commit.rs
@@ -4,7 +4,9 @@ use crate::{
 	error::Result,
 	sync::{repository::repo, utils::get_head_repo},
 };
-use git2::{ErrorCode, message_prettify, ObjectType, Repository, Signature};
+use git2::{
+	message_prettify, ErrorCode, ObjectType, Repository, Signature,
+};
 use scopetime::scope_time;
 
 ///
@@ -130,7 +132,7 @@ pub fn commit_message_prettify(
 		.get_string("core.commentChar")?
 		.as_bytes()
 		.first()
-		.unwrap()
+		.unwrap_or(&b"#"[0])
 		.to_owned();
 
 	Ok(message_prettify(message, Some(char))?)
--- a/asyncgit/src/sync/commit.rs
+++ b/asyncgit/src/sync/commit.rs
@@ -127,15 +127,13 @@ pub fn commit_message_prettify(
 	repo_path: &RepoPath,
 	message: String,
 ) -> Result<String> {
-	let char = repo(repo_path)?
+	let comment_char = repo(repo_path)?
 		.config()?
 		.get_string("core.commentChar")?
-		.as_bytes()
-		.first()
-		.unwrap_or(&b"#"[0])
-		.to_owned();
+		.chars()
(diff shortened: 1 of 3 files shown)


asyncgit/src/sync/commit.rs
```diff
@@ -119,6 +122,22 @@ pub fn tag_commit(
...
+pub fn commit_message_prettify(
+	repo_path: &RepoPath,
+	message: String,
+) -> Result<String> {
+	let comment_char_string =
+		repo(repo_path)?.config()?.get_string("core.commentChar");
+
+	let comment_char =
+		comment_char_string.map_or(b'#', |char_string| {
+			char_string.chars().collect::<Vec<char>>()[0] as u8
```
extrawurst, line 135: lets use a getter here instead of `[]` index access as this will panic if the index does not exist. its better to gracefully handle that by getting an `Option` back

extrawurst, line 135: lets use a getter here instead of `[]` index access as this will panic if the index does not exist. its better to gracefully handle that by getting an `Option` back

Concelare: Updated to use unwrap_or now so no longer panics

asyncgit/src/sync/commit.rs
```diff
@@ -119,6 +122,27 @@ pub fn tag_commit(
...
+	let comment_char_string =
+		repo(repo_path)?.config()?.get_string("core.commentChar");
+
+	let comment_char =
+		comment_char_string.map_or(b'#', |char_string| {
+			char_string
+				.chars()
+				.collect::<Vec<char>>()
+				.first()
+				.unwrap_or(&'#')
```
extrawurst, line 139: i think we can make this nicer, by not having to put in the default twice. you should be able to chain via `map` first and then `unwrap_or` at the end. surprising that this is not a clippy lint :)

Concelare: Updated to do it the cleaner way

extrawurst:
Thank you @Concelare 🥳

extrawurst:
@Concelare please add an entry to the changelog next time too

Issue #2024, opened 2024-01-24
feature request for support branch.sort


lucidash:
gitui seems to support default order(lex-asc) by branch name only.

It will be very helpful when support options to change  branch.sort config
(e.g. `git branch --sort -committerdate`)

extrawurst:
as for almost everything in git implementing this like git does is very complex - for reference: 
* https://git-scm.com/docs/git-config#Documentation/git-config.txt-branchsort
* https://git-scm.com/docs/git-branch#Documentation/git-branch.txt---sortltkeygt
* https://git-scm.com/docs/git-for-each-ref#_field_names

my suggestion is:

adding a sort option popup that allows sorting to be changed between the following keys:
* refname
* last commit timestamp
* last author

at the same time we should probably also show the last commit timestamp in the table

PR #2146, opened 2024-03-22
Add a sort option popup to branch list view

### LINKED ISSUE #2129: Allow sorting by last branch commit
I have quite a few branches and can not always remember the name of them. 

 It would be amazing if the branch list could be sorted my last-commit date.

Don't know of any.

UUGTech:
This Pull Request fixes/closes #2129, #2024

It changes the following:
- show last commit date and last commit author on branchlist popup.
- make it possible to sort by branch name, last commit date, and last commit author.

I followed the checklist:
- [ ] I added unittests

Screenshot 2024-03-22 at 19 48 35

https://github.com/extrawurst/gitui/assets/55311933/2d85c200-9746-46c7-9b28-c9568af14399


src/popups/branchlist.rs
```diff
@@ -554,6 +635,46 @@ impl BranchListPopup {
...
+		let branch_name_length: usize =
+			width_available as usize * 40 / 100;
+
+		(width_available as usize)
+			.saturating_sub(COMMIT_HASH_LENGTH)
+			.saturating_sub(COMMIT_DATE_LENGTH)
+			.saturating_sub(branch_name_length)
+			.saturating_sub(IS_HEAD_STAR_LENGTH)
+			.saturating_sub(three_dots_length)
+	}
```
UUGTech, line 654: Due to the length limitation of get_text(), some implementations are separated.

extrawurst:
Thanks for taking a go at this!

If you look at other popups like the options or search-in-log popup you will notice that the way we do these is: using the navigation cursors to pick an option instead of adding a bunch of new key bindings for each option uniquely. if you could convert this on to the same that would be awesome

UUGTech:
Thank you for pointing that out.
Now, it works with navigation keys 😄 

https://github.com/extrawurst/gitui/assets/55311933/e921cc8f-f017-49c2-b7d4-ee2e7c769da8

src/popups/branchlist.rs
```diff
@@ -587,6 +706,15 @@ impl BranchListPopup {
 			.take(height)
 			.enumerate()
 		{
+			let date_local = Local
+				.timestamp_opt(displaybranch.top_commit_time, 0)
+				.unwrap();
```
Concelare, line 711: Would it be possible to change this from an unwrap to an unwrap_or to protect from panics just incase

extrawurst: wtf `MappedLocalTime`: https://docs.rs/chrono/latest/chrono/offset/type.MappedLocalTime.html

lets do `earliest` and then unwrap the Option with `unwrap_or_else`

extrawurst: I am also debating whether to store this `DateTime<Local>` instead inside of `BranchInfo` which should be fine as it is `just` twice as big as the i64 that we we do not have to do this mess everytime the screen renders

extrawurst: but at the least the unwrap needs to go

src/popups/branchlist.rs
```diff
@@ -601,6 +721,16 @@ impl BranchListPopup {
 			.take(height)
 			.enumerate()
 		{
+			let date_text = displaybranch
+				.top_commit_time_local
+				.map_or("????-??-?? ".to_string(), |time| {
+					time.date_naive().to_string() + " "
+				});
```
UUGTech, line 728: Sorry, I'm not sure how to deal with None case.

asyncgit/src/sync/branch/mod.rs
```diff
@@ -92,6 +96,12 @@ pub struct BranchInfo {
 	///
 	pub top_commit: CommitId,
 	///
+	pub top_commit_time: i64,
+	///
+	pub top_commit_time_local: Option<DateTime<Local>>,
```
UUGTech, line 101: I added `top_commit_time_local` inside of `BranchInfo`.
But this could be `None`, I thought we should keep `top_commit_time: i64` here. What do you think?

src/popups/branchlist.rs
```diff
@@ -102,6 +104,8 @@ impl DrawableComponent for BranchListPopup {
 }
 
 impl Component for BranchListPopup {
+	// TODO: clean up
+	#[allow(clippy::too_many_lines)]
```
UUGTech, line 108: I don't have any idea to minimize this for now..

UUGTech:
@extrawurst Sorry for the delay 🙇‍♂️
I updated according to your feedback and left some comments.

extrawurst (changes requested):
when sorting it should also keep the selection: right now it only keeps the same index selected which will mostly be a different branch when sorting changes.

the popup list entries should also not use a focused (bold) style, make it more in line with other popups (search, fuzzy find)

UUGTech:
@extrawurst 
Thank you for your review.
>when sorting it should also keep the selection: right now it only keeps the same index selected which will mostly be a different branch when sorting changes.

Easier way to find new branch is required in #2129, also I think sorting is that kind of feature.
I thought keeping the selection makes it difficult. 🤔
I myself am not strongly attached to it, so If you still prefer to keep the selection, then I will do so.

>the popup list entries should also not use a focused (bold) style, make it more in line with other popups (search, fuzzy find)

Thank you. I'll fix this.

extrawurst:
> I myself am not strongly attached to it, so If you still prefer to keep the selection, then I will do so.

yes please. the use case you describe is better fixed by  providing quick scroll-to-top/bottom shortcuts like we also have in the commit log list view

UUGTech:
Other popups are in a bold style
Screenshot 2024-04-24 at 21 51 50

Which is better? 
not bold:
Screenshot 2024-04-24 at 21 51 34
bold:
Screenshot 2024-04-24 at 21 58 47

Currently, branch_sort popup is in bold style. This is same with other popups.
If they should not be bold, I will fix all popups.

UUGTech:
> the use case you describe is better fixed by providing quick scroll-to-top/bottom shortcuts like we also have in the commit log list view.

I made an issue for this.

extrawurst:
> Other popups are in a bold style Screenshot 2024-04-24 at 21 51 50
> 
> Which is better? not bold: Screenshot 2024-04-24 at 21 51 34 bold: Screenshot 2024-04-24 at 21 58 47
> 
> Currently, branch_sort popup is in bold style. This is same with other popups. If they should not be bold, I will fix all popups.

Let’s go non bold and make only the selection bold

UUGTech:
updated the style of these popups
#### branch-sort popup
https://github.com/extrawurst/gitui/assets/55311933/233382d6-b6e3-46c1-98d9-479042be8c1c

#### logs-search popup
https://github.com/extrawurst/gitui/assets/55311933/97ff4991-9edd-4b14-b107-090bbc59ecb5

#### fuzzy-find popup
Screenshot 2024-05-01 at 12 25 34

extrawurst:
@UUGTech are you still planning to push this over the finish line? will mark it as `needs_work`

UUGTech:
I'm sorry, but at this time I don't plan to continue working on this. It's okay to mark it as `needs_work` 🙇‍♂️

vztester94w:
This was amazing feature. IMO there is no real purpose for default "sort by name", git is for tracking change over time and should be default.

Issue #1093, opened 2022-01-23
pull on clone with multiple remotes yields Error: 'git: inconclusive remotes' and stuck on 'Pull: preparing ...'


RobertHana:
**Describe the bug**
I cloned a repo and added a 2nd remote.  Using gitui, I issued the Pull[f] and it yielded 'Error: git: inconclusive remotes' and eventually got stuck on 'Pull: preparing ...'.

**To Reproduce**
1. Pull[f]
2. See error, and hit return to close and get stuck on next error with no further way to close.

**Expected behavior**
I expected the pull to pull from the tracked remote, at least.

**Screenshots**
I don't think I need a screenshot, do I?

**Context (please complete the following information):**
#### Software version

gitui 0.19.0

#### Operating system

macOS 11.6 (Darwin 20.6.0)

#### Compile time information

- Profile: release
- Target triple: x86_64-apple-darwin
- Family: unix
- OS: macos
- Architecture: x86_64
- Pointer width: 64
- Endian: little
- CPU features: fxsr,sse,sse2,sse3,ssse3
- Host: x86_64-apple-darwin

#### Environment variables

```bash
SHELL=/bin/bash
EDITOR=vi
GIT_EDITOR=vi
VISUAL=vi
```

#### Command-line

```bash
gitui --bugreport 
```

**Additional context**
If you aren't familiar with how to add multiple remotes:
% git remote add remote1 {git-remote-url1}
% git remote add remote2 {git-remote-url2}

extrawurst:
hm interesting so you have no remote called `origin`? gitui expects there to be one currently. you are right, for pull/fetch we can figure it out by checking the tracking branch but for a push of a brand new branch we would have to ask what remote you want it to go to, this is another feature we currently miss.

does it work if you name one `origin`?

RobertHana:
okay i renamed it `origin` and i now get the error:
```
fetch failed:                                                                                                              ┃
┃git error:config value                                                                                                     ┃
┃'branch.ESPNG-696-Provide-info-in-error-message-about-which-profile-failed-in-multi-profile-report.remote' was not found;  ┃
┃class=Config (7); code=NotFound (-3)                                                                                       
```
Interestingly we do put slashes in our branch names, so the error, if it contained the real branch name, should have read:
```
feature/ESPNG-1471-Add-override-flag-indicators-to-tabs-in-Settings-and-Assumptions-when-something-is-overridden
```

RobertHana:
[...]

PR #2156, opened 2024-03-25
Get default push remote from configuration

cruessler:
This PR is in response to #1093. It changes how GitUI gets the remote to push to. It does not change how it gets the remote to pull from, so it doesn’t directly address #1093, but it is a first step towards making the code paths for getting push and pull configuration independent.

It changes the following:

- It adds `get_default_remote_for_push_in_repo` which reads git configuration related to remote repositories. It first reads `branch.<name>.pushRemote`, then `remote.pushDefault`, then `branch.<name>.remote`. This is [according to the docs][git-config-remote-push-default].

I followed the checklist:

- [x] I added unittests

[git-config-remote-push-default]: https://git-scm.com/docs/git-config#Documentation/git-config.txt-remotepushDefault

Test push behavior in status tab:

- [x] no upstream branch configured
- [x] upstream branch configured, `branch.<>.pushRemote` is set
- [x] upstream branch configured, `remote.pushDefault` is set
- [x] upstream branch configured, `branch.<>.remote` is set


asyncgit/src/sync/remotes/mod.rs
```diff
@@ -51,6 +51,65 @@ pub fn get_default_remote(repo_path: &RepoPath) -> Result<String> {
...
+	repo: &Repository,
+) -> Result<Option<git2::Branch>> {
+	for b in repo.branches(None)? {
+		let branch = b?.0;
+		if branch.is_head() {
+			return Ok(Some(branch));
+		}
+	}
+	Ok(None)
+}
```
cruessler, line 67: This function was copied verbatim from here: https://github.com/cruessler/gitui/blob/4ff0db9b80b816cd69a300877228dbab2c0986a3/asyncgit/src/sync/reword.rs#L37

cruessler:
In addition to `branch.<name>.pushRemote`, we should also try to read `branch.<name>.remote`. I’ll update the PR soon.

https://git-scm.com/docs/git-config#Documentation/git-config.txt-branchltnamegtremote

extrawurst:
looks good! yeah we should really try to get a unittest on this. i think local remotes to push to should work with libgit. so we can probably just simulate this that way.

cruessler:
> In addition to `branch.<name>.pushRemote`, we should also try to read `branch.<name>.remote`. I’ll update the PR soon.
> 
> https://git-scm.com/docs/git-config#Documentation/git-config.txt-branchltnamegtremote

I’ve updated the PR.

extrawurst:
@cruessler what's the status here? can this be merged? is it ready in your opinion? if not what TODOs do you still see?

cruessler:
> @cruessler what's the status here? can this be merged? is it ready in your opinion? if not what TODOs do you still see?

I found a way to start testing this PR. I’ll add a testing-related TODO list to the initial post soon-ish. The function I added is pretty solid I think, what’s left to do is wire it up in the right places and make sure all combinations of config values work as expected.

I didn‘t leave a more high-level comment related to the overall progress, thinking that my last comment was sufficient. Now I see that was not the case. :-)

cruessler:
@extrawurst I updated the PR. I think it’s now ready for testing! I tested different configurations for pushing and did not encounter any issues. I don’t have a good feeling for what I might have accidentally broken, so my testing was focused on pushing.

cruessler:
Also, there is now a bit of duplication, but I thought we could address it in the future, once we now this is the way to go.

asyncgit/src/sync/remotes/mod.rs
```diff
@@ -299,9 +374,44 @@ mod tests {
 			]
 		);
 
-		let res =
-			get_default_remote_in_repo(&repo(repo_path).unwrap());
-		assert_eq!(res.is_err(), true);
-		assert!(matches!(res, Err(Error::NoDefaultRemoteFound)));
+		let default_push_remote = get_default_remote_in_repo(&repo);
```
martihomssoler, line 377: why is this not `get_default_remote_for_push_in_repo(&repo)`? The other asserts match against the default push repo, except this one.

martihomssoler (commented):
I would add back the `test_default_remote_inconclusive()` test. Since the one that you added checks for push defaults and not for the inconclusive case for fetch repos. Apart from that, it look good. :+1:

asyncgit/src/sync/remotes/mod.rs
cruessler: Good catch! Splitting the test into two, restoring the previous `test_default_remote_inconclusive` also fixed this inconsistency.

cruessler:
I updated this PR’s description to reflect the fact that this only tackles getting push configuration yet. Given what I’ve learned about the subject, adding something similar for pull configuration should be rather straight-forward. I think it makes sense to create a follow-up PR, but I’ll be equally happy to add more functionality in this PR.

extrawurst:
@cruessler please add a changelog entry and then we are good to go

cruessler:
> No let’s wrap this one up and make a follow up for pull. Can you create that @cruessler ?

Will do, I’m aiming for sometime this weekend.

cruessler:
> @cruessler please add a changelog entry and then we are good to go

Done. (I didn’t find a specific issue this PR fixes, there’s one for pulling which I’ll link to in the follow-up PR.)

Issue #2173, opened 2024-04-09
Signing implementation currently not interchangeable (impl trait return)


yanganto:
The return type of SignBuilder is a bug.

`impl` as a return type only helps simplify your type signatures
https://doc.rust-lang.org/rust-by-example/trait/impl_trait.html

It will not be a generic thing as you think.
https://stackoverflow.com/questions/52001592/why-can-impl-trait-not-be-used-to-return-multiple-conditional-types

Your project uses tab not space, but the Sign trait implementation is mixed with tab and space here
https://github.com/extrawurst/gitui/blob/5d68ea0e397ea65dce6c8aeb653b1c43918c78e1/asyncgit/src/sync/sign.rs#L122-L134
 
You don't like a simple and working solution #2047, but implement a buggy trait with unnecessary complexity to force me to follow, or want me to solve your issue for free and in silence.

extrawurst:
> The return type of SignBuilder is a bug.
>
> impl as a return type only helps simplify your type signatures

Fair point, we can simply Box it and that's fixed. PR welcome.

> You don't like a simple and working solution https://github.com/extrawurst/gitui/pull/2047, 

I like simple but like I explained over there already: we cannot hardcode a solution that will only work with your branch, there was already the work gpg going for months prior that was conflicting with your work. PLus GPG was a far more requested feature than ssh-signing so it got priority.

> but implement a buggy trait with unnecessary complexity 

Seems like a fairly simple bug (see above)

> to force me to follow, or want me to solve your issue for free and in silence.

No one is forcing you to do anything, I would suggest you check your tone here, we all work on this for close to nothing. The time I invested into providing this tool to you and the whole community for free far exceeds any donations (less than 1% of time invested was paied)

PR #2175, opened 2024-04-10
sign with ssh key in disk

### LINKED ISSUE #1149: Sign git commits with SSH keys

yanganto:
This Pull Request fixes #1149 #2173 

ref: #2047 #2023

It changes the following:
- A disk ssh key path in `user.signingKey` of gitconfig  will be used to sign the commit 

I followed the checklist:
- [X] I added unittests

### DIFF
--- a/asyncgit/Cargo.toml
+++ b/asyncgit/Cargo.toml
@@ -14,6 +14,7 @@ keywords = ["git"]
 [dependencies]
 bitflags = "2"
 crossbeam-channel = "0.5"
+dirs = "5.0"
 easy-cast = "0.5"
 fuzzy-matcher = "0.3"
 git2 = "0.18"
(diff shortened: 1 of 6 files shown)


asyncgit/Cargo.toml
```diff
@@ -27,6 +28,7 @@ rayon = "1.10"
 rayon-core = "1.12"
 scopetime = { path = "../scopetime", version = "0.1" }
 serde = { version = "1.0", features = ["derive"] }
+ssh-key = { version = "0.6.4", features = ["crypto", "encryption"] }
```
jirutka, line 31: This added **dozens** of extra dependencies and increased the binary size by ~1 MiB!

yanganto: Wow, I did not realize this, what is your suggestion?  Is an optional feature good from your end?

PR #2187, opened 2024-04-15
Allow builds from 'git archive' generated tarballs

alerque:
The current release can be build from `git clone` sources because running `git rev-parse HEAD` at build time works, but it fails to build from an archive tarball such as is generated by GitHub when you download the source link on a release and generated with `git archive`.

This fixes that by first checking an env var that a user can set by using `export TAR_COMMIT_ID=$(git get-tar-commit-id <package>.tar.gz)`. It isn't a pancea and requires some intervention, but at least it makes it possible to build without a patch, etc.


build.rs
```diff
@@ -1,9 +1,14 @@
 fn get_git_hash() -> String {
 	use std::process::Command;
 
-	let commit = Command::new("git")
+	// Allow builds from `git archive` generated tarballs if output of `git get-tar-commit-id` is
+	// set in an env var.
+	if let Ok(commit) = std::env::var("TAR_COMMIT_ID") {
+		return commit[..7].to_string();
+	};
+	let commit = Command::new("foobar")
```
extrawurst, line 9: `foobar`?

alerque: Yes. That's a refference to my brain.

(I was having trouble *testing* this inside the Git context I was working in, so I did that to make sure the `git rev-parse` command failed. So it is leftover debug info and I pushed tho wrong rewrite of the commit. Corrected now.)

extrawurst:
@alerque please add a changelog entry matching other contributions

extrawurst:
thanks for your contribution. note that i will change the env name to `BUILD_GIT_COMMIT_ID`. hope that is ok

PR #2207, opened 2024-04-28
Add global config file support for SharedOptions

### LINKED ISSUE #2140: Configuration file support
Whenever I run GitUI in a new working directory, when comparing differences, I have to do `o` to get into Options, then under Diff, I set "Ignore whitespaces" to `true`.

I would like it so that within the config directory, I can place a `gitui.ron` file with the content:

```ron
(
    diff: (
        ignore_whitespace: true,
        context: 3,
        interhunk_lines: 0,
    )
)
```

Then when GitUI starts up, it uses these values as the new defaults, and any changes you make would be written to `.git/gitui` in the working directory.

remique:
This Pull Request fixes/closes #2140.

It changes the following:
-  If gitui fails to find a local config file, it falls back to global options under `$GITUI_HOME/gitui.ron` 

I followed the checklist:
- [ ] I added unittests

Also:
1) We should probably document this somewhere? If a person sets a global config and then sees that it does not change a thing in his gitui (due to the local config), they might be confused. 
2) I have renamed `gitui` file to `gitui.ron` so that both global and local configs match. I hope that's okay.

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -1,3 +1,5 @@
+use crate::args::get_app_config_path;
+
 use anyhow::Result;
 use asyncgit::sync::{
 	diff::DiffOptions, repo_dir, RepoPathRef,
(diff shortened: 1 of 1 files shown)


remique:
Oh, I was surprised by builds not passing but I see that [clippy::empty_docs](https://rust-lang.github.io/rust-clippy/master/index.html#/empty_docs) has been added in `1.78.0`.

Issue #2235, opened 2024-05-19
Loading theme, seems broken 


Alt-iOS:
**Describe the bug**
Gitui cant read ```theme.ron``` file on macos

**To Reproduce**
1 Place the following on your theme.ron
```ron
(
    selected_tab: Some(Reset),
    command_fg: Some(Rgb(205, 214, 244)),
    selection_bg: Some(Rgb(88, 91, 112)),
    selection_fg: Some(Rgb(205, 214, 244)),
    cmdbar_bg: Some(Rgb(24, 24, 37)),
    cmdbar_extra_lines_bg: Some(Rgb(24, 24, 37)),
    disabled_fg: Some(Rgb(127, 132, 156)),
    diff_line_add: Some(Rgb(166, 227, 161)),
    diff_line_delete: Some(Rgb(243, 139, 168)),
    diff_file_added: Some(Rgb(249, 226, 175)),
    diff_file_removed: Some(Rgb(235, 160, 172)),
    diff_file_moved: Some(Rgb(203, 166, 247)),
    diff_file_modified: Some(Rgb(250, 179, 135)),
    commit_hash: Some(Rgb(180, 190, 254)),
    commit_time: Some(Rgb(186, 194, 222)),
    commit_author: Some(Rgb(116, 199, 236)),
    danger_fg: Some(Rgb(243, 139, 168)),
    push_gauge_bg: Some(Rgb(137, 180, 250)),
    push_gauge_fg: Some(Rgb(30, 30, 46)),
    tag_fg: Some(Rgb(245, 224, 220)),
    branch_fg: Some(Rgb(148, 226, 213))
)
```

**Expected behavior**
The theme looks like the catpuccin mocha

**Context (please complete the following information):**
 - `macos 14.4.1`
 - `gitui nightly 2024-05-18`
 - `rustc 1.78.0 (9b00956e5 2024-04-29)`

**Additional context**
The log file says the following 
`.config/gitui/theme.ron"]: 2:24: Expected string`

Alt-iOS:
Thats the version brew installs, but even when moving to stable and installing with `cargo install`, it will give the same error

Alt-iOS:
Going back to gitui 0.26.1 fixes the issue

Alt-iOS:
[...]

PR #2243, opened 2024-05-22
unittest more of the color formats and document breaking change

extrawurst:


extrawurst:
this will close https://github.com/extrawurst/gitui/issues/2235

PR #2244, opened 2024-05-22
Document color serialization breaking change

cruessler:


THEMES.md
```diff
@@ -14,7 +14,18 @@ The file uses the [Ron format](https://github.com/ron-rs/ron) and is located at
 
 Alternatively, you can create a theme in the same directory mentioned above and use it with the `-t` flag followed by the name of the file in the directory. E.g. If you are on linux calling `gitui -t arc.ron`, this will load the theme in `$XDG_CONFIG_HOME/gitui/arc.ron` or `$HOME/.config/gitui/arc.ron`.
 
-Example theme override:
+Example theme override (>= 0.23, < 0.26.2):
```
extrawurst, line 17: lets make it a bit more clear that this is the old approach like I did in the changelog: `old` -> `new` so that it does not get confused to be what should work anymore.

cruessler: I changed this section, let me know what you think! I hope that I remembered the initial version of the config correctly.

extrawurst:
As discussed

PR #2260, opened 2024-06-07
refactor: replace `Default` for clarity

hamirmahal:
This pull request replaces `Default::default()` with T::default() where T is the type of the default value, to make it more clear which default we are using.

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/src/ui/reflow.rs
+++ b/src/ui/reflow.rs
@@ -248,7 +248,7 @@ mod test {
 		text: &str,
 		text_area_width: u16,
 	) -> (Vec<String>, Vec<u16>) {
-		let style = Default::default();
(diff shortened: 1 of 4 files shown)


src/ui/reflow.rs
```diff
@@ -248,7 +248,7 @@ mod test {
 		text: &str,
 		text_area_width: u16,
 	) -> (Vec<String>, Vec<u16>) {
-		let style = Default::default();
+		let style = ratatui::style::Style::default();
```
hamirmahal, line 251: I wasn't sure if `Style::default()`, with an additional import at the top, would be preferable here.

hamirmahal:
@extrawurst are you open to this change?

PR #2262, opened 2024-06-09
Add Go to line feature for the blame view

### LINKED ISSUE #2219: Quickly navigate to line number in blame view
When I open the blame view, I ahve to navigate to the piece of code I care about, often many lines down.

A way to type in the line number to jump to (e.g. so I could type `Shift-L`, `151`, `<Enter>`).

Specifying the file (and line) in the command line, as suggested in https://github.com/extrawurst/gitui/issues/1382, would also be great.

andrea-berling:
It's still probably a bit rough around the edges, but before I made it too pretty, I though I would get some early feedback on it.

This Pull Request fixes/closes #2219.

It changes the following:
- Add new stackable popup to get a line number to go to from the user
- Add a new internal event to change the selection in the most recent
  blame view as a result of using the go to line popup
- Add new keybinding for the go to line functionality (Shift-L)
- Add new command definition for the go to line functionality in BlameFilePopup
- Add clamping of the selected row in set_open_selection

I followed the checklist:
- [ ] I added unittests
- [ ] I added an appropriate item to the changelog'


cruessler:
@andrea-berling Thanks for the PR! I started testing on my machine, and I think this is looking good overall! One thing I noticed, though, is that, in a larger file with syntax highlighting enabled (I chose `CHANGELOG.md`), the blame view gets reloaded when I navigate to a line using the new popup. Is is possible to keep the blame view’s state when jumping to a line?

src/queue.rs
```diff
@@ -68,6 +68,8 @@ pub enum StackablePopupOpen {
 	InspectCommit(InspectCommitOpen),
 	///
 	CompareCommits(InspectCommitOpen),
+	///
+	GotoLine,
```
extrawurst, line 72: i think its a bit of a missed opportunity to not pass it some context: `enum Context {Blame, FileView}` and `PossibleRange(min,max)` to be able to reuse this in various places

andrea-berling:
@cruessler Added the functionality you requested, it makes a lot of sense for large files. Seems to work for me, can you check it out again?

src/queue.rs
andrea-berling: In my last commit I added the `Context` enum and used it to preserve the `BlameProcess` status across line jumps, and I added the two variants you mentioned in line comments. That's because cargo wouldn't let me compile if I didn't use those variants, which I didn't. Do you have a specific example in mind where those pieces of information could be used? If so, I could use them to implement those use cases as to make them more interesting, otherwise I can leave the comments around as placeholder for future functionality 🙂

cruessler (changes requested):
I left a few comments, mostly minor.

I found that there is an issue when you open a blame, then open the goto line popup before syntax highlighting finishes, then close the popup. The syntax highlighting that has completed in the background goes back to an in-progress state and doesn’t change anymore.

@extrawurst There’s a few added `#[derive(Debug)]` in this PR. Is that something we’d like to avoid? (I know that I’ve added similar `derive`s in the past as well.)

andrea-berling:
> I found that there is an issue when you open a blame, then open the goto line popup before syntax highlighting finishes, then close the popup. The syntax highlighting that has completed in the background goes back to an in-progress state and doesn’t change anymore.

@cruessler That is really weird, I was also worried that just cloning the blame as is could run into that issue, but I tested it multiple times and it seems to work fine on my machine. I could also just store the blame if it has completed
Screen Recording 2024-07-01 at 11 44 39

cruessler:
@andrea-berling What happens when you open the goto line popup while syntax highlighting is still running, but close it only once it has finished?

andrea-berling:
@cruessler progress goes back to 90% and remains there. I will look into it, perhaps I'm pushing the blame status too early

src/queue.rs
```diff
@@ -56,6 +56,13 @@ pub enum Action {
 	UndoCommit,
 }
 
+#[derive(Debug, Clone)]
+pub enum Context {
```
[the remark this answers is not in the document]
andrea-berling: I held off giving it a more specific name until I understood more clearly what @extrawurst meant in https://github.com/extrawurst/gitui/pull/2262#discussion_r1657693573. But yeah, ideally, once I have a better idea of where else this Context enum would be used, I will give it a specific name

andrea-berling:
Found a way to offer the same functionality without the need for a context data structure. Not a fan of the fact I had to make the `blame` field of `BlameFilePopup` `pub` (it sticks out like a sore thumb), but it works fine and doesn't run into the issues reported by @cruessler. What do you think?

cruessler:
@andrea-berling I can understand the concern. Is there any chance we can work around this issue by passing `blame: None` to `BlameFileOpen` and reusing the existing `self.blame` in this piece of code? https://github.com/andrea-berling/gitui/blob/76577fc88c3bd0ebe4f3a87742b90cbdc7390245/src/popups/blame_file.rs#L397-L405 Or by switching `self.blame: Option` for `self.blame: BlameRequest` with `enum BlameRequest { StartNew, KeepExisting, UseNew(BlameProcess) }`?

Apart from that, I tested the branch using `cargo run` and I found it to work well!

andrea-berling:
@cruessler yes, I tried the first solution you proposed before pushing my latest version, what I found was that when re-using `self.blame` if not `None` then it would work, but since there is a single `blame_file_popup` field in the app, the `self.blame` object used for the first file blamed would be used in any other blame going forward. For example, if I blamed `CONTRIBUTING.md`, it would keep showing me the blame for `CONTRIBUTING.md`, even if I blamed `CHANGELOG.md` afterwards.

The second one you propose is more interesting: if instead of reusing `self.blame` when it is not `None` we re-use it when explicitly requested (i.e. `KeepExisting`) when closing the Go To Line popup, and we request a new one when opening a new blame, it might just work 🙂 I will try that one and let you know

andrea-berling:
@cruessler I made the change using your enum idea, it seems to work and I like it better 🙂

cruessler (approved):
@extrawurst Just tested on Mac and Linux (Ubuntu, KDE). Looks good to me!

src/popups/goto_line.rs
```diff
@@ -0,0 +1,110 @@
...
+			if let Event::Key(key) = event {
+				if key_match(key, self.key_config.keys.exit_popup) {
+					self.visible = false;
+					self.line.clear();
+					self.queue.push(InternalEvent::PopupStackPop);
+				} else if let KeyCode::Char(c) = key.code {
+					if c.is_ascii_digit() {
+						// I'd assume it's unusual for people to blame
+						// files with milions of lines
+						if self.line.len() < 6 {
```
extrawurst, line 72: for me this is very cryptic? why hardcode to 6 digits? why is it called `line` wouldn't `digits` be more fitting and descriptive? instead of this we can just try to parse the result and only allow `enter` if the current input is valid and even show it red in case it is not.

src/popups/blame_file.rs
```diff
@@ -307,6 +326,15 @@ impl Component for BlameFilePopup {
...
 						));
 					}
+				} else if key_match(
+					key,
+					self.key_config.keys.goto_line,
+				) {
+					self.hide_stacked(true);
+					self.visible = true;
+					self.queue.push(InternalEvent::OpenPopup(
+						StackablePopupOpen::GotoLine,
```
extrawurst, line 336: here we can provide the context: `Blame, MaxLine`

extrawurst (changes requested):
re context: providing `Context, MaxLine` so we can even provide that as validation in the GotoLine popup and make the validation more powerful instead of just `5 valid digits that parse to a usize`.

Context allows us to use this also for the File View in the FileTree

src/popups/goto_line.rs
andrea-berling: The popup is called `goto_line`, thus the first name I came up with for the input was: `line` 😄 but I'm not married to it, `digits` works just as well for me, I trust your judgement for fitting-ness and descriptiveness 🙂 and yeah, I went for a very basic implementation at first based on the simplifying assumption that most people don't work with text files with hundreds of thousands of lines, but I like your idea of validating the input based on the max range and highlighting it red if not valid. I was even thinking of allowing to put a minus sign in front of the whole this to allow navigation to the last line, the second to last line, and in general, the `n`-th to last line. I will work something up in the next few days 🙂

andrea-berling:
Sorry it took me so long to push my latest changes, a bunch of life happened in between 😅 @extrawurst @cruessler what do you guys think?

extrawurst:
@andrea-berling could you update to recent master, would love to get this over the finish line

andrea-berling:
@extrawurst done, I've merged your `master`, made some last-minute adjustments, and added a Changelog entry 🙂  let me know if there is anything else you need from me

cruessler:
@extrawurst @andrea-berling I want to do some testing on my machines later in the day, provided it has not already been merged by then. :-)

extrawurst:
Can’t merge for now. It breaks the command bar when the popup is open and I think we should have a default text in the input

cruessler:
I just tested the changes, and I really like them! There’s one issue that might also not be an issue at all. I can open the popup before the file’s content have been loaded. If I do, the popup doesn’t accept any input because it still has state for a non-loaded file. I get the error message: `Invalid input: only numbers between -1 and 0 (included) are allowed (-1 means denotes the last line, -2 denotes the second to last line, and so on)`. (Also, one of `means denotes` probably can be removed.)

andrea-berling:
> I just tested the changes, and I really like them! There’s one issue that might also not be an issue at all. I can open the popup before the file’s content have been loaded. If I do, the popup doesn’t accept any input because it still has state for a non-loaded file. I get the error message: `Invalid input: only numbers between -1 and 0 (included) are allowed (-1 means denotes the last line, -2 denotes the second to last line, and so on)`. (Also, one of `means denotes` probably can be removed.)

@cruessler Thanks for catching this! The typo was the easiest to fix, but the issue you raised was also reasonably breezy: basically now the go to line popup won't open until it will be possible to determine the actual max line for the file (i.e. until the blame result is something). On most files, I'm not fast enough to try and open the blame popup before this is already the case, so there is no wait at all. On the larger ones (e.g. `Contributing.md`, `Cargo.lock`), you might need to wait a second or two before you can open the popup. Same amount of time you need to wait to even see the contents from the file in the UI, so I would say the UX should be reasonable and not too confusing/frustrating.

> Can’t merge for now. It breaks the command bar when the popup is open and I think we should have a default text in the input

@extrawurst hey 👋 so, when you say "It breaks the command bar", you mean that the popup makes it disappear, like in the following screenshot?
image
If so, I agree, it's not a great UX, although I didn't catch it at first. Curse of knowledge I guess, I suppose that because I knew you need to press Enter to go to a line and Esc to close the popup I didn't notice the lack of a hint.
What do you think of such a command bar for the popup?
image
I haven't pushed the change yet, it's still on my local machine. Let me know if you like it, so I can push that as well 🙂 

As per the default text in the input, I'm not so sure: I can't really imagine a bulletproof guess for a good line number for any file (Start of the file? End of the file? Middle of the file? etc.), and I would imagine that most people would end up deleting the default text anyway and put the line number they want to go to (although I guess one could implement something like HTML's input placeholders so the user doesn't need to delete the default text). Did you perhaps already have an idea for what to put as a default, or did you want to brainstorm it/discuss it?

extrawurst:
@andrea-berling sorry for the lag on this. can you resolve conflicts and update it?

andrea-berling:
@extrawurst Done :slightly_smiling_face:

cruessler:
Also, at this point, I’m tempted to say, let’s get this merged and polish in follow-up PRs. :-)

extrawurst:
@andrea-berling this needs another round of conflict resolves :(

andrea-berling:
@extrawurst Fixed :slightly_smiling_face:

extrawurst:
@andrea-berling one little clippy fix (plus a new conflict, the warnings will self heal after rebasing)

andrea-berling:
@extrawurst Fixed again :smile:

CHANGELOG.md
```diff
@@ -47,13 +48,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * use default shell instead of bash on Unix-like OS [[@yerke](https://github.com/yerke)] ([#2343](https://github.com/gitui-org/gitui/pull/2343))
 
 ### Added
-* add popups for viewing, adding, updating and removing remotes [[@robin-thoene](https://github.com/robin-thoene)] ([#2172](https://github.com/gitui-org/gitui/issues/2172))
```
extrawurst, line 51: probably some merge issue, there should not me any diff here other than your addition

src/strings.rs
```diff
@@ -1870,4 +1882,15 @@ pub mod commands {
 			CMD_GROUP_LOG,
 		)
 	}
+
+	pub fn goto_line(key_config: &SharedKeyConfig) -> CommandText {
+		CommandText::new(
+			format!(
+				"Go To Line [{}]",
```
extrawurst: lets just call it `Go To`

CHANGELOG.md
andrea-berling: Yeah that's weird, I don't remember ever solving a conflict with the `CHANGELOG.md` file, so I wonder where this came frome. At any rate, it's fixed now :+1:

src/strings.rs
andrea-berling: Done :+1:

src/popups/blame_file.rs
```diff
@@ -30,6 +30,8 @@ use ratatui::{
 };
 use std::path::Path;
 
+use super::goto_line::GotoLineContext;
```
extrawurst: please use `crate` import here to be aligned with the above ones

src/popups/goto_line.rs
```diff
@@ -0,0 +1,173 @@
...
+	widgets::{Block, Clear, Paragraph},
+	Frame,
+};
+
+use anyhow::Result;
+
+use crossterm::event::{Event, KeyCode};
+
+#[derive(Debug)]
+pub struct GotoLineContext {
```
extrawurst: lets not use a structured for a single member. we can pass around the `max_line` directly for now

extrawurst (approved):
Awesome, way less complicated now. Just two little nits and then we can merge this! Thanks for sticking with me here!

src/popups/blame_file.rs
andrea-berling: I got rid of `GotoLineContext`, so this problem went away by itself :smile:

src/popups/goto_line.rs
andrea-berling: Agreed, donezo :+1:

andrea-berling:
@extrawurst any remaining blockers? I would say let's try and close this PR before it gets old enough to vote :grin:

PR #2269, opened 2024-06-16
Use gitoxide for rev walk

cruessler:


cruessler:
## Flamegraph using `gitx`

flamegraph

## Flamegraph using `git2`

flamegraph

Byron:
That's all very exciting, thanks for getting started with `gix` in `gitui` :)!

I think it's worth noting that `gix` does not validate the objects it reads right now, which makes it less safe than `git2`, but as safe as Git (as far as I could tell).

This `git2` behaviour can be deactivated with [`strict_hash_verification`](https://docs.rs/git2/latest/git2/opts/fn.strict_hash_verification.html), and maybe more performance can be obtained by [disabling object caching](https://docs.rs/git2/latest/git2/opts/fn.strict_hash_verification.html).

extrawurst:
> This git2 behaviour can be deactivated with [strict_hash_verification](https://docs.rs/git2/latest/git2/opts/fn.strict_hash_verification.html), and maybe more performance can be obtained by [disabling object caching](https://docs.rs/git2/latest/git2/opts/fn.strict_hash_verification.html).

@cruessler did you ever benchmark this?

Byron:
That is interesting!

A note on `gix with use_commit_graph(true): about 13 s` - this would be about 10 times faster if a graph was actually used. It will be used by default (unless disabled in `git config`) if it was present as well. Was it created with `git commit-graph write --reachable`?

Something that seems strange here is that the numbers don't seem to match my own.

For instance, a simple commit traversal (that cannot use the commit-graph cache) can be done (with a hot FS-cache) at 138k commits/s on an M1 Pro.

```
❯ ein t hours
 09:50:50 traverse commit graph done 1.3M commits in 9.46s (138.4K commits/s)
 09:50:50        estimate-hours Extracted and organized data from 1309152 commits in 11.885291ms (110148920 commits/s)
total hours: 1243723.88
total 8h days: 155465.48
total commits = 1309152
total authors: 34641
total unique authors: 26510 (23.47% duplication)

linux ( master) +798 -408 [!] took 9s
❯ git rev-parse @
87d6aab2389e5ce0197d8257d5f8ee965a67c4cd
```

That code uses a the `Simple` iteration directly, which is used through abstractions here. I'd hope that these don't cause such a slowdown.

It would certainly be interesting, @cruessler, to see what `ein t hours` says on your machine. For a perfect comparison, one would certainly want to write a non-GUI program that does the traversal to be sure the right thing is measured.

Damaged pride aside 😅, I am glad that `git2` is this awesome, and that I could help.

cruessler:
To add a bit more context: I’ve created https://github.com/cruessler/gix-benchmarks in order to be able to more thoroughly compare history traversal speed of both `gix` and `git2`. It seems that `gix` is significantly faster, in particular in the Linux kernel. (I hope that I didn’t make a mistake in the benchmark code. :smile:)

Also: I did not know about `git commit-graph write --reachable` at the time, and the numbers most certainly reflect that. :smile:

The benchmark has `strict_object_creation(false); strict_hash_verification(false);` https://github.com/cruessler/gix-benchmarks/blob/73711297dbad890da146f113c3d9f7e92f0afac7/src/main.rs#L63-L65.

PR #2285, opened 2024-07-01
Add support for recursive blame

### LINKED ISSUE #2194: Support for recursive blame
For example in tig (and similarly in magit) you can blame recursively like this: Start by blaming some file, go to some line. Then press `,` to show the blame view of the same region of the same file but at the parent commit. This can be repeated to travel further back in time, always showing the same region of file. Press `<` to go the other way (forward in time).

This is very handy when trying to understand the history of a particular piece of code. It would be great to have it supported in gitui!

kanielrkirby:
This Pull Request closes #2194.

It changes the following:

- Replace `BlameFilePopup.blame: Option<BlameProcess>` with `BlameFilePopup.blame_stack: Vec<BlameProcess>`. This involved changing any references to it, which happened to be a fair amount.
- Add keybinding for going back to a previous blame (`b`).
- Make keybinding for `blame` (`Shift+B`) work in the `popups/blame_file.rs` view, where it will take the commit and open a blame from that.
- Add hints for `blame` and `back` keybindings in blame_file.

I followed the checklist:

- [ ] I added unittests (can add if you think it's necessary)
- [x] I ran `make check` without errors (this reached the same point `master` does, which fails on `test_pre_commit_py`. Does this need an issue?)
- [x] I tested the overall application (used it to make these commits! Haha)


src/popups/blame_file.rs
```diff
@@ -93,7 +93,7 @@ pub struct BlameFilePopup {
 	table_state: std::cell::Cell<TableState>,
 	key_config: SharedKeyConfig,
 	current_height: std::cell::Cell<usize>,
-	blame: Option<BlameProcess>,
+	blame_stack: Vec<BlameProcess>,
```
extrawurst, line 96: I would like us to not keep this stack around. @cruessler and I agreed on working out an asynchronous blaming that will deliver results iteratively so the worst case scenarios are not as slow anymore and this stack will save a lot of data when going only in one direction through the blame history and only optimize for going back fast and not going back and forth.

 Furthermore I have a gut feeling that it might lead to weird stacks when jumping to a file history from a blame and then to another blame again which will not properly reset the stack

extrawurst:
Thanks for tackling this. I only had one comment or request to rework. Would love to add this feature

src/popups/blame_file.rs
kanielrkirby: Ah, that makes sense. So, to clarify, do you want to `a)` have somewhat of a non-popping stack, `b)` just nix the `back` functionality altogether, or `c)` did you have something else in mind?

kanielrkirby: Just a friendly ping :)
@extrawurst

extrawurst:
lets fix the ci, and i noticed that when the popup is open the command bar goes blank, that seems wrong

kanielrkirby:
> lets fix the ci, and i noticed that when the popup is open the command bar goes blank, that seems wrong

Will get on that when I get a chance! Appreciate the update.

kanielrkirby:
> Ah, that makes sense. So, to clarify, do you want to `a)` have somewhat of a non-popping stack, `b)` just nix the `back` functionality altogether, or `c)` did you have something else in mind?

Also @extrawurst I'm not sure if you saw my other comment, but I needed some more info.

PR #2287, opened 2024-07-06
Support `options` in `git_status.fetch()`

### LINKED ISSUE #2263: Stashing tab shows not stage/ not tracked files in view
Hi,

I'm not sure it's even a bug.

When I was doing stashing in Stash[4] tab, you can switch what to stash by toggling staged/untracked files with keys `i/u` respectively.

At first I thought it's noop, since there was no visual indication or cue that anything changed. It works as it should by omitting untracked or staged files when toggling with respective keys.

I figured out that it behaves correctly without any obvious clue only by accident.
I had a single untracked directory with different user. The directory had `root` user and I couldn't stash it cause of permissions. Than the window popped up showing when I was attempting to stash with stashing configuration `{StashUntracked: true, ...}`. When pressing `u`, the config changed and I could stash only tracked files as expected.

**To Reproduce**
Create staged or untracked files with elevated permission like `root` in project folder. Notice error, than press u or i and see it stashes as expected.

**Expected behavior**
I expect to have visual cue for what will be stashed. It should be reflected in the file tree view upon pressing `i/u` keys.

**Screenshots**
Screenshot from 2024-06-10 11-21-16
The below screenshot actually show correct behavious, as there's nothing to stash, but files are still listed in file tree view
Screenshot from 2024-06-10 11-21-30

 - OS/Distro + Version: `Fedora Linux 39 (Workstation Edition) x86_64`
 - GitUI Version `nightly 2024-06-09 ()`
 - Rust version: `1.74.0-x86_64-unknown-linux-gnu`

kanielrkirby:
It looks like support was available/built-in, but possibly left out for some reason?

Happy to rework, this is just to start conversations :)

This Pull Request fixes/closes #2263.

I followed the checklist:
- [ ] I added unittests

It would also be good to know your stance on the `?` idea, or any other additions you might want to this to make this mergeable.

### DIFF
--- a/src/tabs/stashing.rs
+++ b/src/tabs/stashing.rs
@@ -75,14 +75,14 @@ impl Stashing {
 	pub fn update(&mut self) -> Result<()> {
 		if self.is_visible() {
 			let status_type = if self.options.keep_index {
-				StatusType::WorkingDir
-			} else {
 				StatusType::Both
+			} else {
+				StatusType::WorkingDir
 			};
 			let show_untracked = if self.options.stash_untracked {
-				Some(ShowUntrackedFilesConfig::No)
-			} else {
 				Some(ShowUntrackedFilesConfig::All)
+			} else {
+				Some(ShowUntrackedFilesConfig::No)
 			};
 			self.git_status.fetch(&StatusParams::new(
 				status_type,


kanielrkirby:
I just realized I got it backwards! That is totally my bad :sweat_smile:

extrawurst:
thanks for looking into this. i wonder what blame tells us how this got overseen - I assume no one ever used it so it went unnoticed 🙈

kanielrkirby:
> thanks for looking into this. i wonder what blame tells us how this got overseen - I assume no one ever used it so it went unnoticed 🙈

It seems to be! To be completely fair, most people (myself included) don't know how Git works in most cases. All you usually need is the good ol' `clone <repo>; add <things>; commit -m <changes>; push;`

I remember seeing this was from Stephan Dilly (though I'm not sure who that is, he's in a lot of the old code. Is that your old handle / name perhaps? Or an old maintainer?)

After blaming (using the recursive blame I just added! haHA!!), the project at that point was (as far as I can tell) fully able to support it (the options were there already). Maybe it was just planned to be done in a separate commit.

ondrej-ivanko:
I'm glad someone is working on this. Thank you guys. I use stashing a lot, so this would definitely help me.

 I suppose this solution doesn't really change anything in the functionality, when user swaps different values in `Options` -> `Status` -> `Show untracked` on Stash tab [4]. In my case switching to different values in the `Show untracked` option doesn't change what files are displayed on Stash tab [4], only what files are displayed on Status tab [1].

src/tabs/stashing.rs
```diff
@@ -71,9 +74,20 @@ impl Stashing {
 	///
 	pub fn update(&mut self) -> Result<()> {
 		if self.is_visible() {
-			self.git_status
-				//TODO: support options
-				.fetch(&StatusParams::new(StatusType::Both, None))?;
+			let status_type = if self.options.keep_index {
+				StatusType::Both
+			} else {
+				StatusType::WorkingDir
```
naseschwarz, line 80: My expectation here would be, that this file list shows all those files, which would be included in a stash.

Here's `--keep-index`'s description:

```
      -k, --keep-index, --no-keep-index
           This option is only valid for push and save commands.

           All changes already added to the index are left intact.
```

I.e. "keep index" has no effect on whether a change in the index is stashed or not. It just affects whether that change is removed from the index and work tree afterwards.

Thus, I would have expected that whether "keep index" is active or not has no effect on whether that file is listed here.

naseschwarz (changes requested):
Hi!

I've noticed a minor snag with the --keep-index logic in your PR.

If you were to fix that and please add an entry to CHANGELOG.md, I'd say we're good to go, though.

extrawurst:
@kanielrkirby are you committed to help push this over the finish line?

Issue #2298, opened 2024-07-19
[BUG]: Does not support custom git config path via GIT_CONFIG_GLOBAL


utilyre:
**Describe the bug**
I tried to commit some changes and got the following error

```
commit error:
git error:config value 'user.name' was not found; class=Config (7); code=NotFound (-3)
```

It turns out it cannot find `user.name` or any other config because I have my git configuration in a different path from `~/.gitconfig`, which I achieved by setting `GIT_CONFIG_GLOBAL` to `~/.config/git/config.ini`.

**To Reproduce**
1. `export GIT_CONFIG_GLOBAL=~/.config/git/config.ini` (or any other location)
2. Go to a git repository and make some changes
3. Commit through `gitui`

**Expected behavior**
It should check for this environment variable and use it if set.
Also there is another environment variable that is related: `GIT_CONFIG_SYSTEM`.

**Workaround**
```bash
ln -s $GIT_CONFIG_GLOBAL ~/.gitconfig
```

**Context (please complete the following information):**
 - OS/Distro + Version: `Fedora Linux 40`
 - GitUI Version: `gitui 0.26.3 2024-06-02 ()`
 - Rust version: NA (installed through `dnf`)

extrawurst:
That is interesting. upstream supports this since last year: https://github.com/libgit2/libgit2/pull/6544

extrawurst:
@utilyre if you have any chance to build from #2299 it would be great to test if this fixes your problem

koffydrop:
@extrawurst i was having the same problem and the build from #2299 did fix it

extrawurst:
#2299 is merged now and a new nightly build triggered

PR #2301, opened 2024-07-23
Fix log crashing in subdirectories

cruessler:
When opened in a sub-directory of a git repository, the log view crashes. Replacing `gix::open` by `gix::discover` fixes this issue. `gix::open` errors when the passed directory is not a git repository, `gix::discover` searches parent directories for a git repository. The code causing the crash has not been released yet, so no changelog entry is necessary.

### DIFF
--- a/asyncgit/src/error.rs
+++ b/asyncgit/src/error.rs
@@ -94,8 +94,8 @@ pub enum Error {
 	Sign(#[from] crate::sync::sign::SignError),
 
 	///
-	#[error("gix::open error: {0}")]
-	GixOpen(#[from] Box<gix::open::Error>),
+	#[error("gix::discover error: {0}")]
+	GixDiscover(#[from] Box<gix::discover::Error>),
 
 	///
 	#[error("gix::reference::find::existing error: {0}")]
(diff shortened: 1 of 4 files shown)


extrawurst:
@cruessler can you add a teest for that so we don't regress on this again?

extrawurst:
@cruessler we also need to make sure it does respect the env vars, see https://github.com/extrawurst/gitui/issues/2298

cruessler:
> @cruessler can you add a teest for that so we don't regress on this again?

@extrawurst I just added a test for running `AsyncLog` in a sub-directory (the test fails on `master`, but passes on this branch). The test works, but it feels a bit like testing implementation details. I think we could keep it for now, but might want to come up with a cleaner test in the future. What do you think?

I’ll have a look at environment variables next.

extrawurst:
@cruessler please make a followup with a test for the env, this broke in the past and it would be good to know when it does :)

cruessler:
> @cruessler please make a followup with a test for the env, this broke in the past and it would be good to know when it does :)

Will do, I put it on my todo list. :-)

cruessler:
@extrawurst I would really like to write an integration test for env variables, similar to what I’ve tried in #2368. That way, we ideally would have a test that read like “set env variable, start app, expect screenshot showing the loaded app to look like snapshot”. I will also try adding a test to `app.rs`.

extrawurst:
Sounds great!

PR #2343, opened 2024-08-23
Use default shell instead of bash on Unix-like OS

### LINKED ISSUE #2317: Use default shell for git-hooks instead of bash
Would it be possible to run the hooks with the shell specified by `$SHELL` instead of defaulting to bash here? If I understand correctly, it matters when running from a non-login (e.g. gui application) shell since bash won't inherit desired environment variables.

https://github.com/extrawurst/gitui/blob/4ef633a75a3581936d0da9aade7af8ed7aad98ab/git2-hooks/src/hookspath.rs#L116-L117

yerke:
This Pull Request fixes/closes #2317.

It changes the following:
- Add `find_default_unix_shell()`.
- Use default shell instead of bash on Unix-like OS.

Why we need it? macOS uses zsh by default now.

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/git2-hooks/src/hookspath.rs
+++ b/git2-hooks/src/hookspath.rs
@@ -3,7 +3,7 @@ use git2::Repository;
 use crate::{error::Result, HookResult, HooksError};
 
 use std::{
-	path::Path, path::PathBuf, process::Command, str::FromStr,
+	env, path::Path, path::PathBuf, process::Command, str::FromStr,
 };
 
 pub struct HookPaths {
@@ -113,9 +113,10 @@ impl HookPaths {
 
 		log::trace!("run hook '{:?}' in '{:?}'", hook, self.pwd);
 
-		let git_bash = find_bash_executable()
-			.unwrap_or_else(|| PathBuf::from("bash"));
-		let output = Command::new(git_bash)
+		let git_shell = find_bash_executable()
+			.or_else(find_default_unix_shell)
+			.unwrap_or_else(|| "bash".into());
+		let output = Command::new(git_shell)
 			.args(bash_args)
 			.current_dir(&self.pwd)
 			// This call forces Command to handle the Path environment correctly on windows,
@@ -190,4 +191,9 @@ fn find_bash_executable() -> Option<PathBuf> {
 	} else {
 		None
 	}
+}
+
+// Find default shell on Unix-like OS.
+fn find_default_unix_shell() -> Option<PathBuf> {
+	env::var_os("SHELL").map(PathBuf::from)
 }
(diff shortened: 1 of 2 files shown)


yerke:
@extrawurst @mtsgrd Do you mind taking a look? Thanks.

extrawurst:
I suggest you run `make check` locally - it will run everything the CI runs (and currently fails)

yerke:
@extrawurst thanks for letting me know. It passes now both locally and in CI.

git2-hooks/src/hookspath.rs
```diff
@@ -113,9 +113,11 @@ impl HookPaths {
 
 		log::trace!("run hook '{:?}' in '{:?}'", hook, self.pwd);
 
-		let git_bash = find_bash_executable()
-			.unwrap_or_else(|| PathBuf::from("bash"));
-		let output = Command::new(git_bash)
+		let git_shell = find_bash_executable().unwrap_or_else(|| {
```
Byron, line 116: A nit: Could this not be `find_bash_executable().or_else(find_default_unix_shell).unwrap_or_else(|| "bash".into())`?

yerke: Thanks! That looks way better.

extrawurst:
@yerke thank you for your contribution ❤️

cruessler:
@extrawurst There’s a recent commit in `gitoxide` that seems somewhat related at first sight: https://github.com/GitoxideLabs/gitoxide/commit/51bbb8646287f0a6dfa9ff0fd3852c25e001aaf0. I don’t have enough context, though, to know whether it’s actually related.

Byron:
I think it is related. My [initial PR](https://github.com/GitoxideLabs/gitoxide/pull/1752) was based on what I saw in `git2-hooks`, but it was rightfully criticised which led to the change linked [above](https://github.com/extrawurst/gitui/pull/2343#issuecomment-2589223991).
But in the absence of `gix-hooks` I also don't know how Git exactly is running hooks, and to what amount that would apply to making `gitui` more portable.

Joshix-1:
My login-shell isn't posix compliant. git can successfully call the hooks, while gitui fails to do so, because of this patch.
I think `sh` should be used to call the hooks instead of `$SHELL` or `bash`.

If I'm not mistaken (i hate reading c code and i don't know the git codebase) it seems like git uses /bin/sh per default.
https://github.com/git/git/blob/757161efcca150a9a96b312d9e780a071e601a03/run-command.c#L411

```
$ git var GIT_SHELL_PATH
/bin/sh
```

PR #2368, opened 2024-09-23
Add UI snapshot test for simple case

cruessler:
This is a PR for a branch that I have been working on a couple of weeks ago. I did not want this to get lost, so I decided to open a PR, in particular because I discovered there were some open questions with respect to UI tests.

I think this works really well for cases where the program exits with an error message.

I found it quite challenging, though, when it comes to testing an application interactively because `snapbox`, the library that powers the snapshot test, isn’t designed for that use case. There’s potentially ways to work around that via environment variables or command line arguments, but before I go down that route, I first wanted to get your opinion on whether you think it’s worth putting more effort into this. I still think if we can get this to work, we’d have the option to write truly useful high-level regression tests of which I’m a big fan. What do you think? @extrawurst

2024-09-24: I’ll add more context and fix the failing test soon.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -116,6 +116,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "038dfcf04a5feb68e9c60b21c9625a54c2c0616e79b72b0fd87075a056ae1d1b"
 
 [[package]]
+name = "anstyle-lossy"
+version = "1.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


extrawurst:
we are going with `insta` right, so this is superseded by https://github.com/extrawurst/gitui/pull/2411 ?

cruessler:
@extrawurst Yes, unless we’re specifically interested in having SVG snapshots for non-interactive use cases. And I’m thinking this would not be worth the effort currently.

PR #2369, opened 2024-09-24
filetreelist: exclude demo.gif from published crates

decathorpe:
This reduces download size for the crate from ~900 kB to ~10 kB.

I didn't file an issue because that feels redundant for the single-line metadata change proposed here (and most of the checklist in the PR template doesn't apply either).


extrawurst:
thank you!

extrawurst:
pushed a new patch release

decathorpe:
Great, thanks!

PR #2388, opened 2024-10-15
feat: help popup display viewing progress

wugeer:
It changes the following:
- help popiup page show viewing progress

I followed the checklist:
- [ ] I added unittests

before
[gitui_help_popup_before.webm](https://github.com/user-attachments/assets/628feb94-16b9-41cb-9610-c3a0f307c301)

after
[gitui_help_popup_after.webm](https://github.com/user-attachments/assets/97f736b1-9ac3-4bd0-bfa1-21d2537fb510)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ Checkout [THEMES.md](./THEMES.md) for more info.
 ### Added
 * due to github runner changes, the regular mac build is now arm64, so we added support for intel x86 apple build in nightlies and releases (via separat artifact)
 * support `BUILD_GIT_COMMIT_ID` enabling builds from `git archive` generated source tarballs or other outside a git repo [[@alerque](https://github.com/alerque)] ([#2187](https://github.com/extrawurst/gitui/pull/2187))
+* support help popup display viewing progress[[@wugeer](https://github.com/wugeer)](https://github.com/extrawurst/gitui/pull/2388))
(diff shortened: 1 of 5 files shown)


extrawurst:
Thanks for looking into this. The gitui idiomatic way to show progress on a vertical scroll though is using a scrollbar. so I would suggest we integrate that here too. feel free to join our discord to discuss there things upfront in the future to not waste any effort.

wugeer:
In my personal view, having a percentage-based progress might make it a bit easier to understand. :)
It seems that when scrolling up and down on this page, I didn’t notice a scrollbar on the right side.
image
Also, I’ll create a Discord account a bit later.

wugeer:
@extrawurst  I'll try to add a scrollbar on the right, similar to the one in a browser.:)

extrawurst:
Browser?

Just use the same code we use everywhere else do scroll bars in for example the log tab

wugeer:
Yes, I will try it.
image

wugeer:
@extrawurst Have Done!

extrawurst:
looking good! i wonder why you have different formatting - lets see what the ci says about that

wugeer:
My bad, my local environment was on an unstable version. I switched to the stable version of Rust, and cargo clippy works fine.

extrawurst:
Thanks for your contribution

PR #2392, opened 2024-10-17
Fix code formatting

ofek:
Hey there, great tool! I noticed this while checking out the defaults. FYI the rest of the file also uses tabs, not sure if that's intentional.

image


extrawurst:
this was not the only line in that file that had that issue. so i fixed it for all the following lines in 1866bf5

Issue #2155, opened 2024-03-25
Refactor: Create ScrollableTextView component


MichaelAug:
Multiple UI elements have a text area with a vertical scroll bar. This can be refactored into a reusable component to reduce code duplication and improve UI consistency.

Tasks:

1. Create `ScrollableTextView` component
2. Refactor `MsgPopup` to use new component
3. Refactor `DetailsComponent` to use new component

MichaelAug:
Looks like we could use ratatui's built in [Scrollbar widget](https://docs.rs/ratatui/latest/ratatui/widgets/struct.Scrollbar.html) instead of our custom one. This could simplify [scroll_vertical.rs](https://github.com/extrawurst/gitui/blob/master/src/components/utils/scroll_vertical.rs) and [scroll_horizontal.rs](https://github.com/extrawurst/gitui/blob/master/src/components/utils/scroll_horizontal.rs) and probably remove [scrollbar.rs](https://github.com/extrawurst/gitui/blob/master/src/ui/scrollbar.rs) entirely. Would you like this done?

Here's a comparison of the current scrollbar:
![old_scrollbar](https://github.com/user-attachments/assets/1962c6f8-8a46-4071-85b8-8fa5d01f49a4)
And the ratatui built in (default look):
![new_scrollbar](https://github.com/user-attachments/assets/46a2c53f-5383-4b1a-a51d-9ef61e4561c4)

MichaelAug:
> I havent looked into the impact of this but i am all for simplifying and using a upstream solution where we can if it does not degrade performance

I think swapping the custom scrollbar with the ratatui one shouldn't be too difficult, I'll make a PR so we can test the changes

MichaelAug:
I've made a small PR to test the scrollbar. You can see the new scrollbar in e.g. tag description or the file list. Looks like a lot of scrollable components don't use VerticalScroll.
https://github.com/extrawurst/gitui/pull/2393

PR #2393, opened 2024-10-20
WIP - Use upstream scrollbar instead of custom one

MichaelAug:
It changes the following:
- Use upstream scrollbar instead of custom implementation

Here is an example of new scrollbar:
image

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/src/components/utils/scroll_vertical.rs
+++ b/src/components/utils/scroll_vertical.rs
@@ -1,6 +1,6 @@
-use crate::{
-	components::ScrollType,
-	ui::{draw_scrollbar, style::SharedTheme, Orientation},
+use crate::{components::ScrollType, ui::style::SharedTheme};
+use ratatui::widgets::{
+	Scrollbar, ScrollbarOrientation, ScrollbarState,
 };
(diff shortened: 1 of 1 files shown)


extrawurst:
This shows that we are not doing a good job of using our own abstraction everywhere, as most places draw scrollbars themselfs by directly calling `draw_scrollbar`.

Depending on how much time you have we should clean that up. Minimum is to also make `draw_scrollbar` use the ratatui widget.

Ideally though we also migrate all places directly calling `draw_scrollbar` to use our VerticalScroll abstraction

MichaelAug:
yeah i've got time :) I can refactor VerticalScroll and HorizontalScroll and then replace all usages of draw_scrollbar with one of those.

Do you want this work done as part of https://github.com/extrawurst/gitui/issues/2155 or should we make a new issue? Do you have a preference for how this work should be split up? A single giant PR or split it up into multiple?

extrawurst:
Ideally we open a new PR with just the migration to our existing two types and then we followup with this

MichaelAug:
Can do! Will close this PR for now then

PR #2413, opened 2024-10-31
Support "Copy Path" operation in WSL

johnDeSilencio:
This PR adds support for the "Copy Path" operation in the "Files" tab when running in WSL.

For those like myself who are accessing WSL from a terminal, there is no X server, so `xclip` won't work. However, my Ubuntu WSL image at work still has `xclip` installed, which `gitui` detects and uses, causing `gitui` to freeze. I have to kill `gitui` at that point.

This PR checks if we are running in WSL by examining the description of the kernel in the OS release file, `/proc/sys/kernel/osrelease`. If `gitui` is running in WSL, we copy the path to the Windows clipboard using the `clip.exe` utility.

I followed the checklist:
- [ ] I added unittests (I didn't see any other unit tests in `clipboard.rs`, so I didn't try to add any here)
- [X] I ran `make check` without errors (proof: [output.txt](https://github.com/user-attachments/files/17581691/output.txt))
- [X] I tested the overall application (see below)
- [X] I added an appropriate item to the changelog (put in the `Added` subsection under `Unreleased`)

gitui

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Set `CREATE_NO_WINDOW` flag when executing Git hooks on Windows ([#2371](https://github.com/extrawurst/gitui/pull/2371))
 
 ### Added
+* support for "Copy Path" action in WSL [[@johnDeSilencio](https://github.com/johnDeSilencio)] ([#2413](https://github.com/extrawurst/gitui/pull/2413))
 * help popup scrollbar [[@wugeer](https://github.com/wugeer)](https://github.com/extrawurst/gitui/pull/2388))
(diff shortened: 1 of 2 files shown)


cruessler:
> @cruessler could you test that on windows?

I only have Linux and Mac machines available, unfortunately.

Concelare:
I can test it tonight if you want @extrawurst

johnDeSilencio:
@extrawurst Sorry! Meant to mention that I was going to test it when I got back into work today and had access to my WSL instance. Looks like everything is working as expected! :)

gitui_wsl

Concelare:
I tested this and it worked as expected,I didn't have the most time this evening to test it in every way but I had no issues using it as intended. 👍

johnDeSilencio:
> I tested this and it worked as expected,I didn't have the most time this evening to test it in every way but I had no issues using it as intended. 👍
> 
> Tested on Windows 10 WSL 2 Ubuntu 24.04, Used it on few different git repos of different size with no issues

Thank you, @Concelare!

extrawurst:
@johnDeSilencio thank you for the contribution!

PR #2464, opened 2025-01-04
Support ssh-agent and external sign binaries for commit signing with ssh

### LINKED ISSUE #2188: Cant use commit signing
**Describe the bug**
When performing a commit I get the error 
`sign builder error: Failed to retrieve 'user.signingkey' from the git configuration: Currently, we only support a pair of ssh key in disk.`

This is likely because of my git configuration for commit signing.

**To Reproduce**
1. Add the following to your `~/.gitconfig`

```
[user]
...

[gpg]
	format = ssh
[gpg "ssh"]
	program = /Applications/1Password.app/Contents/MacOS/op-ssh-sign
[commit]
	gpgsign = true
[push]
	autoSetupRemote = true
```
2. try to perform a commit
3. See the error above

**Expected behavior**
This should work without error. `git` on the CLI works fine without error

**Screenshots**
Screenshot 2024-04-15 at 11 38 03 AM

 - M2 MacOS 14.4.1
 - GitUI Version gitui nightly 2024-04-14
 - Rust version: cargo 1.77.1 (e52e36006 2024-03-26)

**Additional context**

chirpcel:
This Pull Request fixes/closes #2188.

It changes the following:
- Respect configuration for alternative signing binary (for example 1Password ssh agent)
- Using ssh-keygen for ssh signing operation to support ssh-agent
  - additional: sharing same implementation for default & configured binary

I followed the checklist:
- [x] I added unittests

### DIFF
--- a/asyncgit/src/sync/sign.rs
+++ b/asyncgit/src/sync/sign.rs
@@ -1,6 +1,5 @@
 //! Sign commit data.
 
-use ssh_key::{HashAlg, LineEnding, PrivateKey};
 use std::path::PathBuf;
 
 /// Error type for [`SignBuilder`], used to create [`Sign`]'s
@@ -72,11 +71,11 @@ pub trait Sign {
 
 	/// only available in `#[cfg(test)]` helping to diagnose issues
 	#[cfg(test)]
-	fn program(&self) -> &String;
+	fn program(&self) -> String;
 
 	/// only available in `#[cfg(test)]` helping to diagnose issues
 	#[cfg(test)]
-	fn signing_key(&self) -> &String;
+	fn signing_key(&self) -> String;
 }
 
 /// A builder to facilitate the creation of a signing method ([`Sign`]) by examining the git configuration.
@@ -156,33 +155,50 @@ impl SignBuilder {
 				String::from("x509"),
 			)),
 			"ssh" => {
-				let ssh_signer = config
+				let program = config
+					.get_string("gpg.ssh.program")
+					.unwrap_or_else(|_| "ssh-keygen".to_string());
+
+				let signing_key = config
 					.get_string("user.signingKey")
-					.ok()
-					.and_then(|key_path| {
-						key_path.strip_prefix('~').map_or_else(
-							|| Some(PathBuf::from(&key_path)),
-							|ssh_key_path| {
-								dirs::home_dir().map(|home| {
-									home.join(
-										ssh_key_path
-											.strip_prefix('/')
-											.unwrap_or(ssh_key_path),
-									)
-								})
-							},
+					.map_err(|err| {
+						SignBuilderError::SSHSigningKey(
+							err.to_string(),
 						)
 					})
-					.ok_or_else(|| {
-						SignBuilderError::SSHSigningKey(String::from(
-							"ssh key setting absent",
-						))
-					})
-					.and_then(SSHSign::new)?;
-				let signer: Box<dyn Sign> = Box::new(ssh_signer);
-				Ok(signer)
+					.and_then(|signing_key| {
+						Self::signing_key_into_path(&signing_key)
+					})?;
+
+				Ok(Box::new(SSHSign {
+					program,
+					signing_key,
+				}))
 			}
 			_ => Err(SignBuilderError::InvalidFormat(format)),
+		}
+	}
+
+	fn signing_key_into_path(
+		signing_key: &str,
+	) -> Result<PathBuf, SignBuilderError> {
+		let key_path = PathBuf::from(signing_key);
+		if signing_key.starts_with("ssh-") {
+			use std::io::Write;
+			use tempfile::NamedTempFile;
+			let mut temp_file =
+				NamedTempFile::new().map_err(|err| {
+					SignBuilderError::SSHSigningKey(err.to_string())
+				})?;
+			writeln!(temp_file, "{signing_key}").map_err(|err| {
+				SignBuilderError::SSHSigningKey(err.to_string())
+			})?;
+			let temp_file = temp_file.keep().map_err(|err| {
+				SignBuilderError::SSHSigningKey(err.to_string())
+			})?;
+			Ok(temp_file.1)
+		} else {
+			Ok(key_path)
 		}
 	}
 }
@@ -191,16 +207,6 @@ impl SignBuilder {
 pub struct GPGSign {
 	program: String,
 	signing_key: String,
-}
-
-impl GPGSign {
-	/// Create new [`GPGSign`] using given program and signing key.
-	pub fn new(program: &str, signing_key: &str) -> Self {
-		Self {
-			program: program.to_string(),
-			signing_key: signing_key.to_string(),
-		}
-	}
 }
 
 impl Sign for GPGSign {
@@ -261,55 +267,20 @@ impl Sign for GPGSign {
 	}
 
 	#[cfg(test)]
-	fn program(&self) -> &String {
-		&self.program
+	fn program(&self) -> String {
+		self.program.clone()
 	}
 
 	#[cfg(test)]
-	fn signing_key(&self) -> &String {
-		&self.signing_key
+	fn signing_key(&self) -> String {
+		self.signing_key.clone()
 	}
 }
 
-/// Sign commit data using `SSHDiskKeySign`
+/// Sign commit data using `ssh-keygen`
 pub struct SSHSign {
-	#[cfg(test)]
 	program: String,
-	#[cfg(test)]
-	key_path: String,
-	secret_key: PrivateKey,
-}
-
-impl SSHSign {
-	/// Create new `SSHDiskKeySign` for sign.
-	pub fn new(mut key: PathBuf) -> Result<Self, SignBuilderError> {
-		key.set_extension("");
-		if key.is_file() {
-			#[cfg(test)]
-			let key_path = format!("{}", &key.display());
-			std::fs::read(key)
-				.ok()
-				.and_then(|bytes| {
-					PrivateKey::from_openssh(bytes).ok()
-				})
-				.map(|secret_key| Self {
-					#[cfg(test)]
-					program: "ssh".to_string(),
-					#[cfg(test)]
-					key_path,
-					secret_key,
-				})
-				.ok_or_else(|| {
-					SignBuilderError::SSHSigningKey(String::from(
-						"Fail to read the private key for sign.",
-					))
-				})
-		} else {
-			Err(SignBuilderError::SSHSigningKey(
-				String::from("Currently, we only support a pair of ssh key in disk."),
-			))
-		}
-	}
+	signing_key: PathBuf,
 }
 
 impl Sign for SSHSign {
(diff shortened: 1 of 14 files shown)


chirpcel:
There's one drawback. Like before, we don't support encrypted private keys which are not loaded to agent. The error message isn't as clear as before, but from a functionality point of view it isn't different. I would make a new issue for this behavior and take a look into supporting encrypted private keys.

extrawurst:
> The error message isn't as clear as before

extrawurst:
> The error message isn't as clear as before

cant we make it as clear as before and then followup 

> supporting encrypted private keys.

chirpcel:
Actually I didn't figured out how to capture the passphrase prompt, as it's spawned directly on the tty.
But I just have an Idea, maybe we could provide an empty passphrase as arg to skip the password prompt and map the "wrong passphrase" error to the previous error message.
I'll take a look at it today.

chirpcel:
@extrawurst Worked, we now have a meaningful error message. I tested with 1Password SSH-Agent, Default SSH-Agent, non encrypted keys on disk and encrypted keys on disk.

DaRacci:
Any movement on this?

This would allow me to use gitui without needing to fallback to shell to actually commit.

chirpcel:
@DaRacci Think maintainer is currently busy as asking for co-maintainers and migrating his repo to an org for better support by more people.
For now you can compile yourself with applied pr as I do or wait.

asyncgit/src/sync/sign.rs
```diff
@@ -261,79 +267,99 @@ impl Sign for GPGSign {
...
+		let mut child = cmd
+			.spawn()
+			.map_err(|e| SignError::Spawn(e.to_string()))?;
+
+		let mut stdin = child.stdin.take().ok_or(SignError::Stdin)?;
+
+		stdin
+			.write_all(commit)
+			.map_err(|e| SignError::WriteBuffer(e.to_string()))?;
+		drop(stdin);
```
naseschwarz, line 320: This is fine and simple for now, let's keep it like this. I just want to voice one thought so it can be referenced later: This will need to be changed if we ever support password prompts without a custom askpass. The canonical way for this is to write the commit to a temporary file, sign that with ssh-keygen and read the signature from `tempfilename` + `".sig"`.

asyncgit/src/sync/sign.rs
```diff
@@ -261,79 +267,99 @@ impl Sign for GPGSign {
...
 	#[cfg(test)]
-	fn signing_key(&self) -> &String {
-		&self.signing_key
+	fn signing_key(&self) -> String {
+		self.signing_key.clone()
 	}
 }
 
-/// Sign commit data using `SSHDiskKeySign`
+/// Sign commit data using `SSHSign`
```
naseschwarz, line 280:
```suggestion
/// Sign commit data using `ssh-keygen`
```

asyncgit/src/sync/sign.rs
```diff
@@ -261,79 +267,99 @@ impl Sign for GPGSign {
...
+			.stdout(Stdio::piped())
+			.stderr(Stdio::piped())
+			.arg("-Y")
+			.arg("sign")
+			.arg("-n")
+			.arg("git")
+			.arg("-f")
+			.arg(&self.signing_key);
+
+		if &self.program == "ssh-keygen" {
```
naseschwarz, line 305: Sorry, I don't really understand why you only pass `-P` if this condition is fulfilled.

naseschwarz (changes requested):
Hey @chirpcel. Thank you for this implementation. I support this fully and agree that this should be the way commit signing is implemented at the moment. It is almost in compliance with `git commit`'s implementation. Where changes will be required for currently unsupported `git commit` signing features, I have remarked them in this review. I wouldn't say that these are blockers, though.

I also have minor questions and nitpicks, mostly about corner cases, here and there and some FYIs.

On which platforms have you tested this? I have tested it on Linux and it works as promised. It would be great if someone tested this on Windows and Mac, too, though. (A test should consist of at least signing a commit and checking whether that commit's signature can be validated.)

asyncgit/src/sync/sign.rs
chirpcel: Cause there are some implementations which doesn't support empty -P with fallback to agent. For example the 1Password SSH-Agent and it's cli `op-sign` throws an error in that case. On default `ssh-keygen` its tested to work but it hasn't to be on 3rd parties.

asyncgit/src/sync/sign.rs
```diff
@@ -261,79 +267,99 @@ impl Sign for GPGSign {
...
+			cmd.arg("-P").arg("\"\"");
+		}
```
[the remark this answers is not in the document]
chirpcel: Seems like a more clean solution for me. I'll like to change it to that way

chirpcel:
Thanks @naseschwarz for your cr! I commited your suggestions and commented on questions. I'll take a look at the remaining annotations of edge cases/optimizations. I'll update the pr soon. 

I've tested the implementation on linux and macOS. Windows currently missing, maybe I can test it in a VM. Currently don't know about license/activation. Keep you updated.

hazzuk:
Would simply like to add that I've been been running a compiled version of gitui with this pull request since August. So far I've had no issues on my Linux setup.

For me, ssh-agent support was critical, so I'd also like to say thank you for this work. I hope we see this merged one day.

chirpcel:
@hazzuk Me too and I totally forgot that I have this PR open :face_in_clouds:  
You just bring back my attention to it. I'll sort out the last open notes and the we should get it merged.

asyncgit/src/sync/sign.rs
```diff
@@ -156,35 +155,52 @@ impl SignBuilder {
...
+			let mut temp_file =
+				NamedTempFile::new().map_err(|err| {
+					SignBuilderError::SSHSigningKey(err.to_string())
+				})?;
+			writeln!(temp_file, "{signing_key}").map_err(|err| {
+				SignBuilderError::SSHSigningKey(err.to_string())
+			})?;
+			let temp_file = temp_file.keep().map_err(|err| {
+				SignBuilderError::SSHSigningKey(err.to_string())
+			})?;
```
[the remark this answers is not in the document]
chirpcel: I hold the instance now until drop of SSHSign struct

asyncgit/src/sync/sign.rs
```diff
@@ -156,35 +155,52 @@ impl SignBuilder {
...
 			}
 			_ => Err(SignBuilderError::InvalidFormat(format)),
 		}
 	}
+
+	fn signing_key_into_path(
+		signing_key: &str,
+	) -> Result<PathBuf, SignBuilderError> {
+		let key_path = PathBuf::from(signing_key);
+		if signing_key.starts_with("ssh-") {
```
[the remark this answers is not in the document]
chirpcel: I've handled this case now. If there's a `key::` prefix I remove it for further processing.

asyncgit/src/sync/sign.rs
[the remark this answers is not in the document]
chirpcel: I switched to the documented `-U` flag.

chirpcel:
Open notes should now been solved. 
I'll test tomorrow on some machines with different configs and then request an follow-up review.

chirpcel:
I didn't encounter problems on my machines. Tested on linux with openssh agent and pgp based ssh agent.

asgeirn:
Tested on MacOS using 1Password signing, worked perfectly!

chirpcel:
@extrawurst 
As this initial PR is some months old and it happened much since them, who is currently able to do a review on changes?

cruessler:
I’ll have a look!

cruessler:
(Just wanted to let you know that I haven’t forgotten this PR. I hope to be getting to it over the weekend.)

asyncgit/src/sync/sign.rs
```diff
@@ -156,35 +155,62 @@ impl SignBuilder {
...
-						SignBuilderError::SSHSigningKey(String::from(
-							"ssh key setting absent",
-						))
-					})
-					.and_then(SSHSign::new)?;
-				let signer: Box<dyn Sign> = Box::new(ssh_signer);
-				Ok(signer)
+					.map(|value| {
+						if value.starts_with("key::") {
+							value.replacen("key::", "", 1)
```
cruessler: What do you think about `value.strip_prefix("key::")` instead (seems more idiomatic)?

asyncgit/src/sync/sign.rs
```diff
@@ -156,35 +155,62 @@ impl SignBuilder {
...
+					let named_temp_file =
+						Self::signing_key_into_temp_file(
+							signing_key.as_str(),
+						)?;
+					signing_key = format!(
+						"{}",
+						named_temp_file.path().display()
+					);
+					pub_key_temp_file = Some(named_temp_file);
+				}
```
cruessler: What do you think about changing to something like the following:

```
let (signing_key, pub_key_temp_file) = {
  …
};
```

This could make the data flow easier to follow.

asyncgit/src/sync/sign.rs
```diff
@@ -156,35 +155,62 @@ impl SignBuilder {
 				String::from("x509"),
 			)),
 			"ssh" => {
-				let ssh_signer = config
+				let program = config
+					.get_string("gpg.ssh.program")
+					.unwrap_or_else(|_| "ssh-keygen".to_string());
+
+				let mut signing_key = config
```
cruessler: What do you think about renaming to `signing_key_path` since that’s the field name in `SSHSign`?

asyncgit/src/sync/sign.rs
```diff
@@ -156,35 +155,62 @@ impl SignBuilder {
...
+					program,
+					signing_key_path: signing_key,
+					_pub_key_temp_file: pub_key_temp_file,
+				}))
 			}
 			_ => Err(SignBuilderError::InvalidFormat(format)),
 		}
 	}
+
+	fn signing_key_into_temp_file(
```
cruessler: What do you think about prefixing this with a verb so that it becomes `write_signing_key_to_temp_file` (or something similar)?

asyncgit/src/sync/sign.rs
```diff
@@ -261,79 +277,90 @@ impl Sign for GPGSign {
...
-					))
-				})
-		} else {
-			Err(SignBuilderError::SSHSigningKey(
-				String::from("Currently, we only support a pair of ssh key in disk."),
-			))
-		}
-	}
+	signing_key_path: String,
+	_pub_key_temp_file: Option<NamedTempFile>,
```
cruessler, line 333: Why is this prefixed with `_`? Does this warrant a comment?

cruessler (changes requested):
First of all, sorry for the long delay! (I wanted to make sure I had the time and focus needed for an actual review, and that took some time, unfortunately.)

Overall, this looks good to me, as far as I can tell. I just have a few questions.

SergioRibera:
Any movement on this? 👀

asgeirn:
With some help from Claude I've rebased this branch on the latest master, it's at https://github.com/asgeirn/gitui/tree/feature/external-ssh-signer for those who want to try it.

extrawurst:
@asgeirn are you interested in driving this PR over the finish line? i adopted it. could use some folks actually using ssh-agent based signing to give it another go and test

extrawurst:
@cruessler I adopted this PR and cleaned up the remaining nits. can you give it a last test run 🙏

asgeirn:
Tested it with my setup (1password agent commit signing) and it works.
> warning: gitui@0.28.1: buildname '0.28.1-nightly 2026-07-01 (ca34d08)'

cruessler:
> @cruessler I adopted this PR and cleaned up the remaining nits. can you give it a last test run 🙏

Will do!

cruessler (changes requested):
I just tested this branch on my Linux machine (Ubuntu 26.04), configuring a new repo as follows:

```bash
git init
echo "test" > test.txt
git config set --local gpg.format ssh
git config set --local user.signingKey "..."
git config set --local commit.gpgsign true
```

Then, using the binary built via `cargo build`, I tried to commit, but the command seemed to hang indefinitely. This was the last line printed to the log:

```
18:36:13 [TRACE] (1) asyncgit::sync::sign: [asyncgit/src/sync/sign.rs:362] signing command: "ssh-keygen" "-Y" "sign" "-n" "git" "-f" "/tmp/.tmpskBtps" "-U"
```

Running the `ssh-keygen` command outside of `gitui` resulted in:

```
ssh-keygen -Y sign -n git -f /tmp/.tmpskBtps -U
Signing data on standard input
```

Also, I got two `ssh-add` processes running at about 100 % CPU in the background.

When I ran `ssh-add` before committing, though, `gitui` did not hang and `git cat-file commit HEAD` showed that the commit had been signed as it had `BEGIN SSH SIGNATURE` in the `gpgsig` header.

extrawurst:
@cruessler i created a reproducer test for your finding. please verify it works in the above case for you now. did you have any other findings or concerns that we should hold out a merge for?

cruessler:
I just checked again and it worked without me having to run `ssh-add` before. It’s possible that I used the wrong format for `user.signingKey` initially, though I don’t know for sure. No objections from my side!

cruessler (approved):

extrawurst:
@chirpcel congrats for landing this bad boy! Thanks everyone for your reviews and testing @naseschwarz, @cruessler, @asgeirn - lets test the shit out of the next nightly!

PR #2486, opened 2025-01-16
Conventional Commit Popup

tkr-sh:
It changes the following:
- Add a new "Conventional Commit" popup, with a feature "[gitmoji](https://gitmoji.dev/)" that allows quick templating messages with rich git commit messages following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

I followed the checklist:
- [ ] I added unittests (I don't think that there are unittests on popup usually)

Also, I know that the UI and the UX is different than the one of Fuzzy finder, but I prefer this one personally (I think that the UX is just better tbh), and it's still easy to follow, even for non vim users :) !

Video:
[asciicast](https://asciinema.org/a/kCIT0qMOBdwQkJfOWUJwzasUB)
_Note: the video is kinda glitched because of the emojis, but everything is aligned_


naseschwarz:
Hi @tkr-sh. Thank you for your effort and for sharing this.

If you're still interested, would you be so kind and add a few key screenshots (or even a video capture) that describe this change so we can discuss based on this? I guess this might help this patch along. It's unlikely many people will join the discussion if they have to assert whether a 1k patch does any shenanigans on their development machines or spool up a cloud instance for every iteration and I take it that there will be discussions.

tkr-sh:
Sure. Tho in theory, it could still work and have untrusted code. In any case, this should be reviewed, and it's straightforward enough so that people realize that there are no shady stuff. 
I'll do that in a few hours

naseschwarz:
Thank you!

> Tho in theory, it could still work and have untrusted code.

Sure, screenshots won't change the level of trust by much. They're not so much to inspire trust but to provide the input that's needed for a discussion that can move the PR along.

Cargo.toml
```diff
@@ -76,8 +77,9 @@ chrono = { version = "0.4", default-features = false, features = ["clock"] }
 maintenance = { status = "actively-developed" }
 
 [features]
-default = ["ghemoji", "regex-fancy", "trace-libgit", "vendor-openssl"]
+default = ["ghemoji", "regex-fancy", "trace-libgit", "vendor-openssl", "gitmoji"]
```
extrawurst, line 80: making this default on will be controversial. I remember we did that in the past and it bloats the binary for a lot of people that are not interested in this

tkr-sh: Well, it can be true sometimes, but in this specific case I don't think that it's a good argument because
```sh
# With the feature
$ cargo build --release && du -b target/release/gitui
15038504	target/release/gitui
# Without the feature
$ cargo build --release && du -b target/release/gitui
15025560	target/release/gitui
```
(_12944B diff ≈ 0.086% of the total binary size_)
But you could make the point that most people don't want gitmoji, and in this case, removing it as a default feature is understandable.

tkr-sh:
Added a video and fixed/upgraded some stuff @naseschwarz @extrawurst

PR #2489, opened 2025-01-18
fix: disable blame and history popup for untracked files

kpbaks:
An untracked file does not have any history data. Right now when you press `B` for the blame popup or the `H` for the history popup you get an empty popup where the title spins endlessly trying to find the file in the commit history, and show relevant information. This commit 

It changes the following:
- disables the two actions in the `StatusTreeComponent`, when the selected item is a file which is not tracked by git.

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/src/components/status_tree.rs
+++ b/src/components/status_tree.rs
@@ -394,11 +394,18 @@ impl Component for StatusTreeComponent {
 		out: &mut Vec<CommandInfo>,
 		force_all: bool,
 	) -> CommandBlocking {
+		let available = self.focused || force_all;
+		let selection = self.selection_file();
+		let selected_is_file = selection.is_some();
+		let tracked = selection.is_some_and(|s| {
+			!matches!(s.status, StatusItemType::New)
+		});
+
 		out.push(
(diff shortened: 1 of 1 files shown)


kpbaks:
I have run `make check`. The only error reported by `clippy` is this.
image

Hope this is okay. I think it could be addressed in another PR to refactor the cognitive complexity of the `event` function

Joshix-1:
I would add `#[expect(clippy::cognitive_complexity)]` to the function. `make check` should not fail. The refactor should be done in a future pr.

kpbaks:
I have added `#[expect(clippy::cognitive_complexity)]` to the function. Now `make check` passes.

extrawurst:
@kpbaks thanks for your contribution. please add a changelog entry and its good to go

extrawurst (changes requested):
Please add changelog entry matching other changelog entries style

kpbaks:
Added entry to CHANGELOG.md

extrawurst (changes requested):
LGTM

PR #2499, opened 2025-01-26
fix: File selector should jump back and highlight unstaged files

### LINKED ISSUE #2476: File selector should jump back and highlight unstaged files
**Describe the bug**
When you make a commit from staged changes and actually do the commit so that the commit message popup closes, the file highlighter does not jump back into the unstaged changes section. It only does this when you use the arrow up/down keys.

**To Reproduce**
1. Go to Status tab
2. Select unstaged changes and stage them
3. Create a commit and write a message and finish it. So the popup window with the commit message input closes
4. Now there is no file selector in the unstaged changes section only after you use the arrow key up/down

**Expected behavior**
I would expect that after a commit, at least the first item in the unstaged changes list is highlighted. So that you get a visual feedback where you are when you move the arrow keys up and down.

tommady:
This Pull Request fixes/closes #2476.

It changes the following:
-  src/popups/commit.rs

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+* After commit: jump back to unstaged area [[@tommady](https://github.com/tommady)] ([#2476](https://github.com/extrawurst/gitui/issues/2476))
+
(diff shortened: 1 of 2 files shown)


extrawurst:
@tommady please add a changelog entry next time like suggested by the checklist

PR #2515, opened 2025-02-11
feat(ui): Add use_selection_fg flag to control selection foreground color

Upsylonbare:
When set to `False` (default `True`), keep the foreground color of the selected object (commit hash, date, etc ...).

It changes the following:  
- Added a new `use_selection_fg` flag to the `Theme` struct, allowing users to toggle whether the selection foreground color is applied.  
- Set `use_selection_fg` to `true` by default in `Theme::Default`.  
- Extended unit tests to validate the behavior of `use_selection_fg`.  

I followed the checklist:  
- [x] I added unittests  

### DIFF
--- a/THEMES.md
+++ b/THEMES.md
@@ -49,3 +49,16 @@ Note that if you want to turn it off, you should use a blank string:
     line_break: Some(""),
 )
 ```
+## Customizing selection
+
+By default the `selection_fg` color is used to color the text of the selected line.
+Diff line, filename, commit hashes, time and author are re-colored with `selection_fg` color.
+This can be changed by specifying the `use_selection_fg` boolean in your `theme.ron`:
+
+```
+(
+    use_selection_fg: Some(false),
+)
+```
+
+By default, `use_selection_fg` is set to `true`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Select syntax highlighting theme out of the defaults from syntect [[@vasilismanol](https://github.com/vasilismanol)] ([#1931](https://github.com/extrawurst/gitui/issues/1931))
 * new command-line option to override the default log file path (`--logfile`) [[@acuteenvy](https://github.com/acuteenvy)] ([#2539](https://github.com/gitui-org/gitui/pull/2539))
 * dx: `make check` checks Cargo.toml dependency ordering using `cargo sort` [[@naseschwarz](https://github.com/naseschwarz)]
+* add `use_selection_fg` to theme file to allow customizing selection foreground color [[@Upsylonbare](https://github.com/Upsylonbare)] ([#2515](https://github.com/gitui-org/gitui/pull/2515))
 
 ### Changed
 * improve syntax highlighting file detection [[@acuteenvy](https://github.com/acuteenvy)] ([#2524](https://github.com/extrawurst/gitui/pull/2524))
--- a/src/ui/style.rs
+++ b/src/ui/style.rs
@@ -16,6 +16,7 @@ pub struct Theme {
 	command_fg: Color,
 	selection_bg: Color,
 	selection_fg: Color,
+	use_selection_fg: bool,
 	cmdbar_bg: Color,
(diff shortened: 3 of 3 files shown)


Upsylonbare:
Sure, with default selection_bg (Blue):

here is the before :
image

and after : (with use_selection_fg to false)
image

When using darker selection_bg, this is more useful
Without
image

With:
image

THEMES.md
```diff
@@ -49,3 +49,14 @@ Note that if you want to turn it off, you should use a blank string:
     line_break: Some(""),
 )
 ```
+## Customizing selection
+
+If you don't want to use the `selection_fg` color, you can also specify `use_selection_fg` in your `theme.ron`:
```
naseschwarz, line 54: It'd be nice to  establish a premise before explaining how to change the default behavior. Maybe preface this by something akin to "Selected items (such as files, commits or lines) are often colored using `selection_fg` in front of `selection_bg` instead of individual colors (such as `commit_hash`, `commit_time`, etc.)."?

naseschwarz (changes requested):
Thank you for your PR. I've tested this and it works nicely.

image

image

Please consider adding an entry to CHANGELOG.md.

THEMES.md
Upsylonbare: Hello, thank you for your review
Sure, I'll document the default behavior before explaining how to change it.

naseschwarz (approved):

naseschwarz (approved):

vlad-anger:
@Upsylonbare  I also would like to keep fg colors same on selection,
good PR!
Can we do better without introducing extra `use_selection_fg`?
If make `selection_fg` optional, no extra fields required

For `theme.ron` that will mean doing
`selection_fg: Some(None)` (to override default)

And default impl. in code:
`selection_fg: Some(Color::White)`
So we keep previous behavior

Here's patch implementing that based on your commits, 
let me know what you think:
https://gist.github.com/vlad-anger/a5fcd217fad7c68459980618e375d806

naseschwarz:
> Here's patch implementing that based on your commits, let me know what you think: 

Wouldn't this break user's theme configurations as it would change the selection_fg type?

vlad-anger:
> Wouldn't this break user's theme configurations as it would change the selection_fg type

Correct, it will.
Only for users, who define custom theme **and** they
specified custom `selection_fg` - will be required
to do _minor_ tweak.

Advantage is simplicity, logic incapsulated within one field

**Maybe** with my approach `selection_fg: Some(Some(your_color))`
will look a bit outstanding of other colors config 
Otherwise we need to take extra field for extra logic
branch of `selection_fg`, which also feels a bit much
So i'm not sure

extrawurst:
Breaking is fine as long as we document it properly in the changelog.

cruessler:
Another option: we already have more than one code path for loading themes, so adding another one could potentially automate the migration. I’m all for making users’ lives easier, but I also understand if we don’t want to increase code complexity in this instance.

https://github.com/gitui-org/gitui/blob/232ad89a4dd8a85eb0eb0bc77c5ea064a51d2dd6/src/ui/style.rs#L304-L324

vlad-anger:
Honestly i'm not sure what's better
`selection_fg: Some(Some(your_color))` will also look a bit weird in theme config,
because of how patchy works, while having extra field also not perfect  : )

Upsylonbare:
> Honestly i'm not sure what's better `selection_fg: Some(Some(your_color))` will also look a bit weird in theme config, because of how patchy works, while having extra field also not perfect : )

Yes you're right.
To me this can be merge as is at it is a more verbose way to write theme and don't introduce messy writings in theme's file. It seems more "simpler" to the user. The "Some()" uses in the theme file will surely evolve in the future though. I think this kind of writing doesn't please anyone but we can't change it for now.
I'm the kind of guy that personalize a lot its tools so will surely do other PR in the `style.rs` file. I can keep your remarks in a corner of my head.

extrawurst:
@Upsylonbare please resolve conflicts.
@naseschwarz can I get a final review and `merge_ready` if you think its a go?

Upsylonbare:
Sorry first time I deal with Github PR, I think I've clicked on the wrong button

naseschwarz (approved):

PR #2524, opened 2025-02-18
Improve syntax detection

acuteenvy:
Currently, gitui prioritizes file extensions for syntax detection. When a file lacks an extension (e.g. `Makefile`), or the extension isn't tied to a specific format (e.g. `.lock`), it disables syntax highlighting.

This PR changes the syntax detecton method to the entire filename and the first line of the file using [`find_syntax_for_file()`](https://docs.rs/syntect/5.2.0/syntect/parsing/struct.SyntaxSet.html#method.find_syntax_for_file).

I followed the checklist:
- [ ] I added unittests
- [x] I ran `make check` without errors
- [x] I tested the overall application
- [x] I added an appropriate item to the changelog

#### Screenshots of highlighting `Cargo.lock` and `Makefile` from this repository:
<details>
<summary>Before</summary>

old1
old2
</details>

<details>
<summary>After</summary>

new1
new2
<details>

PR #2527, opened 2025-02-19
Remove redundant `to_str()` conversion

acuteenvy:
The previously used function could only take `&str` as its argument, and I blindly copied it over here. This should've been part of #2524, sorry.

I followed the checklist:
- [ ] I added unittests


extrawurst:
@acuteenvy thanks for following this up! Are you interested in participating more in gituis development? I am in the process of migrating it into its own org - would you join our discord?

acuteenvy:
Sure, I'll join, but I probably won't be very active there, as I rarely use discord. As for participating more, well, I'll try to send a PR every once in a while, but I don't wanna make any promises. Thanks for the invitation though!

PR #2542, opened 2025-03-01
Push with refspec

### LINKED ISSUE #2537: Failure to follow config push settings
**Describe the bug**
I'm working on a repo with multiple remotes (origin and fork) and have a branch locally with different name which pushes to `main` branch on the fork

**To Reproduce**
1. set upstream of a branch to a differently named one:
`.git/config`
```
[branch "lints"]
	remote = workspace-lints
	merge = refs/heads/main
```
2. set default for push to upstream (without it the cli asks you to specify to which branch)
in `.git/config`
```
[push]
    default = upstream
```
3. prepare a commit and push with gitui
4. gitui creates a branch with the name of the local branch and pushes to it

**Expected behavior**
when executing `git push --verbose` on such branch
```
Pushing to gitlab.torproject.org:matt022/workspace-lints
To gitlab.torproject.org:matt022/workspace-lints
 = [up to date]          lints -> main
updating local tracking ref 'refs/remotes/workspace-lints/main'
Everything up-to-date
```

**Screenshots**
N/A

Rust: cargo 1.84.0 (66221abde 2024-11-19) (stable)
gitui: 0.27.0
OS: NixOS 25.05 (Warbler) x86_64
Host: Windows Subsystem for Linux - NixOS (2.4.11)
Kernel: Linux 5.15.167.4-microsoft-standard-WSL2
Shell: bash 5.2.37

**Additional context**
As noted in the logs this is a gitlab instance if that changes things

vlad-anger:
Fixes #2537

git2 merged [necessary PR](https://github.com/rust-lang/git2-rs/pull/1131)

This PR _is_blocked_ until new version git2 lib released
For now use git2 from master to be able to test PR.

---

Assume git config:
```git-config
[branch "oblava"]
	remote = origin
	merge = refs/heads/master
[push]
  default = upstream
```
Prev. behavior:
push -> push oblava -> remote oblava <- new remote created

New behavior:
push -> push oblava -> remote master

---

I followed the checklist:
- [x] I added unittests
- [x] I ran `make check` without errors (see commit with tests fixes)

### DIFF
--- a/asyncgit/src/error.rs
+++ b/asyncgit/src/error.rs
@@ -54,6 +54,10 @@ pub enum Error {
 	Git(#[from] git2::Error),
 
 	///
+	#[error("git config error: {0}")]
+	GitConfig(String),
+
+	///
 	#[error("strip prefix error: {0}")]
 	StripPrefix(#[from] StripPrefixError),
 
--- a/asyncgit/src/sync/branch/mod.rs
+++ b/asyncgit/src/sync/branch/mod.rs
@@ -245,7 +245,7 @@ pub fn get_branch_remote(
 
 /// Retrieve the upstream merge of a local `branch`,
 /// configured in "branch.*.merge"
-/// 
+///
 /// For details check git2 `branch_upstream_merge`
 pub fn get_branch_upstream_merge(
 	repo_path: &RepoPath,
--- a/asyncgit/src/sync/branch/mod.rs
+++ b/asyncgit/src/sync/branch/mod.rs
@@ -243,6 +243,25 @@ pub fn get_branch_remote(
 	}
 }
 
+/// Retrieve the upstream merge of a local `branch`,
+/// configured in "branch.*.merge"
+/// 
+/// For details check git2 `branch_upstream_merge`
+pub fn get_branch_upstream_merge(
+	repo_path: &RepoPath,
+	branch: &str,
+) -> Result<Option<String>> {
+	let repo = repo(repo_path)?;
+	let branch = repo.find_branch(branch, BranchType::Local)?;
+	let reference = bytes2string(branch.get().name_bytes())?;
+	let remote_name = repo.branch_upstream_merge(&reference).ok();
+	if let Some(remote_name) = remote_name {
+		Ok(Some(bytes2string(remote_name.as_ref())?))
+	} else {
+		Ok(None)
+	}
+}
+
 /// returns whether the pull merge strategy is set to rebase
 pub fn config_is_pull_rebase(repo_path: &RepoPath) -> Result<bool> {
 	let repo = repo(repo_path)?;
@@ -672,6 +691,49 @@ mod tests_branches {
 			&root.as_os_str().to_str().unwrap().into();
 
 		assert!(get_branch_remote(repo_path, "foo").is_err());
+	}
+
+	#[test]
+	fn test_branch_no_upstream_merge_config() {
+		let (_r, repo) = repo_init().unwrap();
+		let root = repo.path().parent().unwrap();
+		let repo_path: &RepoPath =
+			&root.as_os_str().to_str().unwrap().into();
+
+		let upstream_merge_res =
+			get_branch_upstream_merge(&repo_path, "master");
+		assert!(
+			upstream_merge_res.is_ok_and(|v| v.as_ref().is_none())
+		);
+	}
+
+	#[test]
+	fn test_branch_with_upstream_merge_config() {
+		let (_r, repo) = repo_init().unwrap();
+		let root = repo.path().parent().unwrap();
+		let repo_path: &RepoPath =
+			&root.as_os_str().to_str().unwrap().into();
+
+		let branch_name = "master";
+		let upstrem_merge = "refs/heads/master";
+
+		let mut config = repo.config().unwrap();
+		config
+			.set_str(
+				&format!("branch.{branch_name}.merge"),
+				&upstrem_merge,
+			)
+			.expect("fail set branch merge config");
+
+		let upstream_merge_res =
+			get_branch_upstream_merge(&repo_path, &branch_name);
+		assert!(upstream_merge_res
+			.as_ref()
+			.is_ok_and(|v| v.as_ref().is_some()));
+		assert_eq!(
+			&upstream_merge_res.unwrap().unwrap(),
+			upstrem_merge
+		);
 	}
 }
 
--- a/asyncgit/src/sync/config.rs
+++ b/asyncgit/src/sync/config.rs
@@ -62,6 +62,52 @@ pub fn untracked_files_config_repo(
 	Ok(ShowUntrackedFilesConfig::All)
 }
 
+// see https://git-scm.com/docs/git-config#Documentation/git-config.txt-pushdefault
+/// represents `push.default` git config
+#[derive(PartialEq, Eq)]
+pub enum PushDefaultStrategyConfig {
+	Nothing,
+	Current,
+	Upstream,
+	Simple,
+	Matching,
+}
+
+impl Default for PushDefaultStrategyConfig {
+	fn default() -> Self {
+		Self::Simple
+	}
+}
+
+impl<'a> TryFrom<&'a str> for PushDefaultStrategyConfig {
+	type Error = crate::Error;
+	fn try_from(
+		value: &'a str,
+	) -> std::result::Result<Self, Self::Error> {
+		match value {
+			"nothing" => Ok(Self::Nothing),
+			"current" => Ok(Self::Current),
+			"upstream" | "tracking" => Ok(Self::Upstream),
+			"simple" => Ok(Self::Simple),
+			"matching" => Ok(Self::Matching),
+			_ => Err(crate::Error::GitConfig(format!(
+				"malformed value for push.default: {value}, must be one of nothing, matching, simple, upstream or current"
+			))),
+		}
+	}
+}
+
+pub fn push_default_strategy_config_repo(
+	repo: &Repository,
+) -> Result<PushDefaultStrategyConfig> {
+	(get_config_string_repo(repo, "push.default")?).map_or_else(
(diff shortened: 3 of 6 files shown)


naseschwarz (changes requested):
Just for completeness, I'll add a reference to our discussion @vlad-anger. Thank you for the productive discussion!

In #2537 we discussed that the new behavior should be conditional on the value of `push.default`: Only if `push.default == upstream`, `merge.*.branch` is used as the remote branch.

cruessler:
Just to make sure I get the context right: does any of this require changes to https://github.com/gitui-org/gitui/blob/35b2529c52d601d9640530c55c3805e3a9882fea/asyncgit/src/sync/remotes/mod.rs#L133-L233? Or is that not the case as this PR deals with branches while the code I linked to deals with remotes?

vlad-anger:
> Or is that not the case as this PR deals with branches while the code I linked to deals with remotes?

Yeah those helpers out of scope, they tell which remote can be used,
while this PR is fixing different local/remote _branches_ issue,

Ex. locally i have branch t1, which should be pushed to remote t2.
To achieve that, you could have `push.default=upstream` &
`branch.t1.merge=refs/heads/t2`
For now gitui will try to push t1 -> t1 always

vlad-anger:
Tweaked logic, updated PR desc.
Waiting for upstream version release, so CI build won't fail and we
can do final checks

asyncgit/src/sync/remotes/push.rs
```diff
@@ -163,9 +171,23 @@ pub fn push_raw(
 		PushType::Tag => "tags",
 	};
 
-	let branch_name =
+	let mut push_ref =
 		format!("{branch_modifier}refs/{ref_type}/{branch}");
-	remote.push(&[branch_name.as_str()], Some(&mut options))?;
+
+	if !delete
+		&& matches!(
```
naseschwarz, line 178: Is there a reason to use `matches!` unless you want to do pattern matching? Shouldn't this be a simple equality?

asyncgit/src/sync/config.rs
```diff
@@ -62,6 +62,54 @@ pub fn untracked_files_config_repo(
...
+pub fn push_default_strategy_config_repo(
+	repo: &Repository,
+) -> Result<PushDefaultStrategyConfig> {
+	(get_config_string_repo(repo, "push.default")?).map_or_else(
+		|| Ok(PushDefaultStrategyConfig::default()),
+		|entry_str| {
+			Ok(PushDefaultStrategyConfig::try_from(
+				entry_str.as_str(),
+			)
+			.unwrap_or_default())
```
naseschwarz, line 108: Just a note: The behavior of the git command line client is to reject a push if the value of push.default is defined but not valid.

```
% git -c push.default=snerflop push origin HEAD
error: malformed value for push.default: snerflop
error: must be one of nothing, matching, simple, upstream or current
fatal: unable to parse 'push.default' from command-line config
128 %
```

Doing this might come in handy in case `git` adds new options in the future, which gitui does not support, so we don't accidentally implement incorrect push behavior.

asyncgit/src/sync/remotes/push.rs
```diff
@@ -163,9 +171,23 @@ pub fn push_raw(
...
+			push_default_strategy,
+			PushDefaultStrategyConfig::Upstream
+		) {
```
naseschwarz, line 181: The way `push_ref` is constructed leads me to believe that this code is also used to push tags, but `branch.*.merge` should only be applied to branches, right? Thus, this condition should probably be extended by `&& ref_type == PushType::Branch`.

asyncgit/src/sync/config.rs
```diff
@@ -62,6 +62,54 @@ pub fn untracked_files_config_repo(
...
+	fn try_from(
+		value: &'a str,
+	) -> std::result::Result<Self, Self::Error> {
+		match value {
+			"nothing" => Ok(Self::Nothing),
+			"current" => Ok(Self::Current),
+			"upstream" => Ok(Self::Upstream),
+			"simple" => Ok(Self::Simple),
+			"matching" => Ok(Self::Matching),
+			_ => Err(crate::Error::Generic(format!(
```
naseschwarz, line 92: May I suggest to introduce a new error type, like `Error::GitConfig`? We could add the offending configuration key and maybe provide slightly nicer error messages in gitui.

asyncgit/src/sync/branch/mod.rs
```diff
@@ -243,6 +243,25 @@ pub fn get_branch_remote(
 	}
 }
 
+/// Retrieve the upstream merge of a local `branch`,
+/// configured in "branch.*.merge"
+/// 
+/// For details check git2 `branch_upstream_merge`
+pub fn get_branch_upstream_merge(
```
naseschwarz, line 250: `asyncgit/src/sync/branch/mod.rs` seems to have rather good test coverage. Would it be reasonable to add one for `get_branch_upstream_merge`?

asyncgit/src/sync/config.rs
```diff
@@ -62,6 +62,54 @@ pub fn untracked_files_config_repo(
 	Ok(ShowUntrackedFilesConfig::All)
 }
 
+// see https://git-scm.com/docs/git-config#Documentation/git-config.txt-pushdefault
+/// represent `push.default` git config
```
naseschwarz, line 66:
```suggestion
/// represents `push.default` git config
```

asyncgit/src/sync/config.rs
```diff
@@ -62,6 +62,54 @@ pub fn untracked_files_config_repo(
...
+		value: &'a str,
+	) -> std::result::Result<Self, Self::Error> {
+		match value {
+			"nothing" => Ok(Self::Nothing),
+			"current" => Ok(Self::Current),
+			"upstream" => Ok(Self::Upstream),
+			"simple" => Ok(Self::Simple),
+			"matching" => Ok(Self::Matching),
+			_ => Err(crate::Error::Generic(format!(
+				"{value} is incorrect push.default"
```
naseschwarz, line 93: Let's steal the error message from git. This makes it easier for users to find a solution.

```suggestion
				"malformed value for push.default: {value}, must be one of nothing, matching, simple, upstream or current"
```

naseschwarz (commented):
Thank you very much for addressing my concerns. :+1: 

Here are some thoughts on the current version.

naseschwarz:
Apart from the review item regarding pushing tags, I believe the push logic in this version is correctly aligned with git for `push.default` in {`upstream`, `simple`}.

`matching` and `upstream` are options that, as far as I can tell, fall back to `simple` in case a push source is given, which I assume is always the case in `push_raw` (unless delete is true, in which case only a destination is used). Thus, I'd agree that the implementation covers these correctly, too. It might be worth making this explicit so that devs keep this in mind when working on `push_raw`, e.g. by actually explicitly writing `match` on `push_default_strategy` and adding a short remark in the branch that does nothing.

asyncgit/src/sync/remotes/push.rs
extrawurst: Yeah lets make it derive `PartialEq` and then its a simple `==`

vlad-anger: ok

asyncgit/src/sync/config.rs
vlad-anger: Good point. Now we'll fall with normal err
image

asyncgit/src/sync/config.rs
vlad-anger: It's here. I've made it's string, cause not sure that kind of error will _always_ be with config key entry

asyncgit/src/sync/branch/mod.rs
vlad-anger: Added

asyncgit/src/sync/config.rs
vlad-anger: haha

asyncgit/src/sync/config.rs
vlad-anger: Good point, like it!

asyncgit/src/sync/config.rs
```diff
@@ -62,6 +62,54 @@ pub fn untracked_files_config_repo(
...
+
+impl<'a> TryFrom<&'a str> for PushDefaultStrategyConfig {
+	type Error = crate::Error;
+	fn try_from(
+		value: &'a str,
+	) -> std::result::Result<Self, Self::Error> {
+		match value {
+			"nothing" => Ok(Self::Nothing),
+			"current" => Ok(Self::Current),
+			"upstream" => Ok(Self::Upstream),
```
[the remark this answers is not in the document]
vlad-anger: Ja it's legacy but let's support

vlad-anger:
@naseschwarz I liked first review round, let's go for another one!

naseschwarz (approved):
I've tested this against the git2@d1ae3b6c2d1200e7d82468af447fa66259225ecf and it works as I would have expected. A push in `gitui` now behaves like a `git push <push default> $(git rev-parse --abbrev-ref HEAD)` in the terminal.

There's only one small formatting remark.

naseschwarz:
There has been a git2 release, upgrade and patch in #2567! :)

asyncgit/src/error.rs
```diff
@@ -53,6 +53,10 @@ pub enum Error {
 	#[error("git error:{0}")]
 	Git(#[from] git2::Error),
 
+	///
+	#[error("git config error:{0}")]
```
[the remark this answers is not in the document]
vlad-anger: Yeap

naseschwarz (changes requested):
Hey @vlad-anger. Please excuse me missing this in the previous two rounds, but I have one more request: Would you please add an entry to CHANGELOG.md?

vlad-anger:
Yeap added changelog. Rebased on top of master

vlad-anger:
@naseschwarz Thanks for reviewing! :)

naseschwarz (approved):

PR #2547, opened 2025-03-05
Add timeouts for commit hooks

### LINKED ISSUE #2546: Program hangs if repo uses prepare-commit-hook that requires user input
**Describe the bug**
If a repo implements a prepare-commit-msg which requires user input gitui will hang on various operations that trigger the hook (commit, rebase, merge, cherry pick). 

This is a pretty common pattern in repos using commitizen as one of the options that the project recommends is using [prepare-commit-msg](https://github.com/commitizen/cz-cli#traditional-git-hooks) 

**To Reproduce**

minimal example here: https://github.com/jaredmcateer/cz-gitui

1. Open repo with prepare-commit-msg that requires user input
2. Add/Modify a file
3. Stage file
4. Go to 'staged change' in gitui
5. Press 'c'
7. Observe gitui is unresponsive

**Screenshots**

In your processes you will see something like this:
Image

 - OS/Distro + Version: tested on Mac OS Sonoma 14.7.1 & Pop_OS! 22.0.4 LTS
 - GitUI Version: gitui nightly 2025-03-04 ()
 - Rust version: rustc 1.82.0 (f6e511eec 2024-10-15)

**Additional context**
Lazygit also suffers from this, and the current workaround is to override all the commit-like commands with custom keybindings to disable Husky, but that will also kill other useful things like pre-commit hooks that run tests/linters.

DaRacci:
This Pull Request closes #2546.

It changes the following:
- Adds timeouts to git hooks

I followed the checklist:
- [x] I added unittests

### DIFF
--- a/asyncgit/src/sync/hooks.rs
+++ b/asyncgit/src/sync/hooks.rs
@@ -2,6 +2,7 @@ use super::{repository::repo, RepoPath};
 use crate::error::Result;
 pub use git2_hooks::PrepareCommitMsgSource;
 use scopetime::scope_time;
+use std::time::Duration;
 
 ///
 #[derive(Debug, PartialEq, Eq)]
@@ -10,6 +11,13 @@ pub enum HookResult {
 	Ok,
 	/// Hook returned error
 	NotOk(String),
+	/// Hook timed out
+	TimedOut {
+		/// Stdout
+		stdout: String,
+		/// Stderr
+		stderr: String,
+	},
 }
 
 impl From<git2_hooks::HookResult> for HookResult {
@@ -22,6 +30,11 @@ impl From<git2_hooks::HookResult> for HookResult {
 				stderr,
 				..
 			} => Self::NotOk(format!("{stdout}{stderr}")),
+			git2_hooks::HookResult::TimedOut {
+				stdout,
+				stderr,
+				..
+			} => Self::TimedOut { stdout, stderr },
 		}
 	}
 }
@@ -31,29 +44,65 @@ pub fn hooks_commit_msg(
 	repo_path: &RepoPath,
 	msg: &mut String,
 ) -> Result<HookResult> {
+	hooks_commit_msg_with_timeout(repo_path, msg, None)
+}
+
+/// see `git2_hooks::hooks_commit_msg`
+#[allow(unused)]
+pub fn hooks_commit_msg_with_timeout(
+	repo_path: &RepoPath,
+	msg: &mut String,
+	timeout: Option<Duration>,
+) -> Result<HookResult> {
 	scope_time!("hooks_commit_msg");
 
 	let repo = repo(repo_path)?;
-
-	Ok(git2_hooks::hooks_commit_msg(&repo, None, msg)?.into())
+	Ok(git2_hooks::hooks_commit_msg_with_timeout(
+		&repo, None, msg, timeout,
+	)?
+	.into())
 }
 
 /// see `git2_hooks::hooks_pre_commit`
 pub fn hooks_pre_commit(repo_path: &RepoPath) -> Result<HookResult> {
+	hooks_pre_commit_with_timeout(repo_path, None)
+}
+
+/// see `git2_hooks::hooks_pre_commit`
+#[allow(unused)]
+pub fn hooks_pre_commit_with_timeout(
+	repo_path: &RepoPath,
+	timeout: Option<Duration>,
+) -> Result<HookResult> {
 	scope_time!("hooks_pre_commit");
 
 	let repo = repo(repo_path)?;
 
-	Ok(git2_hooks::hooks_pre_commit(&repo, None)?.into())
+	Ok(git2_hooks::hooks_pre_commit_with_timeout(
+		&repo, None, timeout,
+	)?
+	.into())
 }
 
 /// see `git2_hooks::hooks_post_commit`
 pub fn hooks_post_commit(repo_path: &RepoPath) -> Result<HookResult> {
+	hooks_post_commit_with_timeout(repo_path, None)
+}
+
+/// see `git2_hooks::hooks_post_commit`
+#[allow(unused)]
+pub fn hooks_post_commit_with_timeout(
+	repo_path: &RepoPath,
+	timeout: Option<Duration>,
+) -> Result<HookResult> {
 	scope_time!("hooks_post_commit");
 
 	let repo = repo(repo_path)?;
 
-	Ok(git2_hooks::hooks_post_commit(&repo, None)?.into())
+	Ok(git2_hooks::hooks_post_commit_with_timeout(
+		&repo, None, timeout,
+	)?
+	.into())
 }
 
 /// see `git2_hooks::hooks_prepare_commit_msg`
@@ -66,8 +115,26 @@ pub fn hooks_prepare_commit_msg(
 
 	let repo = repo(repo_path)?;
 
-	Ok(git2_hooks::hooks_prepare_commit_msg(
-		&repo, None, source, msg,
+	Ok(git2_hooks::hooks_prepare_commit_msg_with_timeout(
+		&repo, None, source, msg, None,
+	)?
+	.into())
+}
+
+/// see `git2_hooks::hooks_prepare_commit_msg`
+#[allow(unused)]
+pub fn hooks_prepare_commit_msg_with_timeout(
+	repo_path: &RepoPath,
+	source: PrepareCommitMsgSource,
+	msg: &mut String,
+	timeout: Option<Duration>,
+) -> Result<HookResult> {
+	scope_time!("hooks_prepare_commit_msg");
+
+	let repo = repo(repo_path)?;
+
+	Ok(git2_hooks::hooks_prepare_commit_msg_with_timeout(
+		&repo, None, source, msg, timeout,
 	)?
 	.into())
 }
@@ -77,7 +144,7 @@ mod tests {
 	use std::{ffi::OsString, io::Write as _, path::Path};
 
 	use git2::Repository;
-	use tempfile::TempDir;
+	use tempfile::{tempdir, TempDir};
 
 	use super::*;
 	use crate::sync::tests::repo_init_with_prefix;
@@ -125,7 +192,7 @@ mod tests {
 		let (_td, repo) = repo_init().unwrap();
 		let root = repo.workdir().unwrap();
 
-		let hook = b"#!/bin/sh
+		let hook = b"#!/usr/bin/env sh
 	echo 'rejected'
 	exit 1
 			";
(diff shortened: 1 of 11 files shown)


DaRacci:
Opening early as a draft.

Initial changes are working and have tests added for them just need to figure out how we are going to implement a timeout value, if thats user configuration or a default impl.

DaRacci:
Was wanting to get some feedback on this before progressing further, is this the correct place to implement the timeout logic?

extrawurst:
This would impact other users like Gitbutler. @Byron what do you think about this change?

asyncgit/src/sync/hooks.rs
```diff
@@ -174,4 +280,53 @@ mod tests {
...
+
+		git2_hooks::create_hook(
+			&repo,
+			git2_hooks::HOOK_COMMIT_MSG,
+			hook,
+		);
+
+		let res = hooks_pre_commit(
+			&root.to_str().unwrap().into(),
+			Duration::ZERO,
```
cruessler, line 301: Should this be a non-zero value? At least, I was expecting a non-zero value based on the test’s name.

cruessler (commented):
From an implementation standpoint, this looks good. The only thing I would add is tests for non-zero timeouts (maybe that’s something you already planned or maybe I missed something).

asyncgit/src/sync/hooks.rs
DaRacci: you're right this should be a non zero, will get that fixed up

DaRacci:
Also worth mentioning as its done in this PR, I've changed the test shell shebang to use `/usr/bin/env sh` as this is a more widely accepted and allows distributions like NixOS to run the tests.

DaRacci:
I've also noticed that while running tests that since i use the non POSIX Shell (nushell) that i need to prefix the command with SHELL=bash, i was wondering if it may be worth testing if the $SHELL is a known non compliant shell and trying to look for a default POSIX shell in the path?

asyncgit/src/sync/hooks.rs
```diff
@@ -26,50 +30,148 @@ impl From<git2_hooks::HookResult> for HookResult {
...
+			None,
+		)),
+		Err(RecvTimeoutError::Disconnected) => {
+			unreachable!()
+		}
+	}
+}
+
 /// see `git2_hooks::hooks_commit_msg`
 pub fn hooks_commit_msg(
```
Byron, line 66: Even though it's certainly fine to make this a breaking change, I think making new variants of these functions with `_with_timeout()` suffix could be considered.

asyncgit/src/sync/hooks.rs
```diff
@@ -26,50 +30,148 @@ impl From<git2_hooks::HookResult> for HookResult {
...
+	});
+
+	match rx.recv_timeout(timeout) {
+		Ok(result) => result,
+		Err(RecvTimeoutError::Timeout) => Ok((
+			HookResult::NotOk("hook timed out".to_string()),
+			None,
+		)),
+		Err(RecvTimeoutError::Disconnected) => {
+			unreachable!()
```
Byron, line 60: If `f()` panics, `tx` would drop which I *think* may actually trigger this.

asyncgit/src/sync/hooks.rs
```diff
@@ -26,50 +30,148 @@ impl From<git2_hooks::HookResult> for HookResult {
...
+	let (tx, rx) = channel();
+	let _ = std::thread::spawn(move || {
+		let result = f();
+		tx.send(result)
+	});
+
+	match rx.recv_timeout(timeout) {
+		Ok(result) => result,
+		Err(RecvTimeoutError::Timeout) => Ok((
+			HookResult::NotOk("hook timed out".to_string()),
```
Byron, line 56: It would be nice if timeout-errors would be discoverable. Then it would need to be strongly typed.

DaRacci:
Now that the implementation is solid how do i go about with setting the timeout, how do we expose configurable value for the user and should there be a default timeout?

asyncgit/src/sync/hooks.rs
```diff
@@ -26,50 +30,148 @@ impl From<git2_hooks::HookResult> for HookResult {
...
+
+	let (tx, rx) = channel();
+	let _ = std::thread::spawn(move || {
+		let result = f();
+		tx.send(result)
+	});
+
+	match rx.recv_timeout(timeout) {
+		Ok(result) => result,
+		Err(RecvTimeoutError::Timeout) => Ok((
```
Byron, line 55: I don't think this is a viable way of handling hook timeouts. After all, the thread that calls `f()` will keep running, and so will the hook program.
Depending on what it does, there can be strange interactions if the calling application keeps running hooks which pile up that way.

Instead, I think the timeout functionality would have to go into `git2-hooks`, and the one controlling the `std::process::Command` would have to use polling to eventually [`kill`](https://doc.rust-lang.org/std/process/struct.Child.html#method.kill) the hook process. This is similar to what a user would do in the terminal if something takes too long.

Byron (commented):
Thanks for letting me know, @extrawurst.

The changes seem to not apply to `git2-hooks`, so wouldn't affect GitButler in its current form. However, that would change if timeouts would move into `git2-hooks` as I seem to be proposing :).

asyncgit/src/sync/hooks.rs
DaRacci: I've tested this and if a panic occurs in the thread it propagates down and the unreachable isn't reached.

DaRacci:
@Byron I've updated the implementation and it is now inside git2-hooks, I've added new functions suffixed with_timeout for each hook but the run_hooks function has a breaking change having the timeout parameter added to it.

asyncgit/src/sync/hooks.rs
DaRacci: This is now strongly typed in `git2_hooks::HookResult` and `asyncgit::sync::hooks::HookResult`

asyncgit/src/sync/hooks.rs
DaRacci: I've implemented variants in both git2-hooks and asyncgit.

DaRacci:
The last remaining thing to do is move the tests i had written over to the git2_hooks crate, even so the current tests in asyncgit should be able to cover most cases i think.

DaRacci:
I was also thinking about the initial issue where the problem was an infinite wait because the hook was waiting for input, maybe a solution could be when a hook is running a new window is shown with the stdio from the hook and also allowing stdin, not sure if that would work how im thinking though.

git2-hooks/src/hookspath.rs
```diff
@@ -130,7 +136,28 @@ impl HookPaths {
 				"DUMMY_ENV_TO_FIX_WINDOWS_CMD_RUNS",
 				"FixPathHandlingOnWindows",
 			)
-			.output()?;
+			.stdout(Stdio::piped())
+			.stderr(Stdio::piped())
```
Byron, line 140: This changes the semantics we had before, where `stdin` was null (see the docs of [`output`](https://doc.rust-lang.org/std/process/struct.Command.html#method.output)).
Doing so should also make it impossible for a hook to hang waiting on `stdin`. Hooks which still hang are probably buggy in another way.

git2-hooks/src/hookspath.rs
```diff
@@ -130,7 +136,28 @@ impl HookPaths {
...
+			let timer = std::time::Instant::now();
+			while child.try_wait()?.is_none() {
+				if timer.elapsed() > timeout {
+					debug!("killing hook process");
+					child.kill()?;
+					return Ok(HookResult::TimedOut { hook });
+				}
+
+				std::thread::yield_now();
+				std::thread::sleep(Duration::from_millis(10));
```
Byron, line 156: Doesn't `sleep()` imply a yield, among other things?

git2-hooks/src/lib.rs
```diff
@@ -66,6 +67,10 @@ pub enum HookResult {
 		/// path of the hook that was run
 		hook: PathBuf,
 	},
+	TimedOut {
```
Byron, line 70: Docs are missing.
Probably the crate could use a `deny(missing_docs)` as well.

git2-hooks/src/hookspath.rs
```diff
@@ -106,7 +108,11 @@ impl HookPaths {
 
 	/// this function calls hook scripts based on conventions documented here
 	/// see <https://git-scm.com/docs/githooks>
-	pub fn run_hook(&self, args: &[&str]) -> Result<HookResult> {
+	pub fn run_hook(
```
Byron, line 111: I still think this doesn't have to be a breaking change unless one really wants to tell folks to deal with the timeout. That timeouts are implemented now probably means most people didn't have issues with this (or solved them by fixing the hooks), so I'd argue that `run_hook_with_timeout()` should be favored.

git2-hooks/src/hookspath.rs
Byron, line 156: One could think about a quadratic backoff (or a custom sequence) here where the first few milliseconds are checked more often, and as it takes longer, one checks less often.
This might also be overthinking it as 10ms seems fine.

Depending on how this is done, I think one *should* mention in the docs that the `timeout` may prolong the actual runtimes, right now by up to 10ms.

git2-hooks/src/hookspath.rs
```diff
@@ -106,7 +108,11 @@ impl HookPaths {
 
 	/// this function calls hook scripts based on conventions documented here
 	/// see <https://git-scm.com/docs/githooks>
```
Byron, line 110: Needs a documentation update.

Byron (commented):
Thanks a lot for the quick turnaround!

The new implementation definitely addresses the previous issue, and now the timeout has an actual effect on the hook.

When looking at `git2-hooks` exclusively, I am not sure I saw a test for this though.

DaRacci:
I'll get some tests put into the git2-hooks crate and resolve these issue when I get some times on the weekend to do so, please let me know if there is anything else that should be looked into for when i get to it.

git2-hooks/src/lib.rs
DaRacci: I looked at adding this to the crate but there are a number of errors that are outside the scope of this PR.

Byron:
> @Byron I've updated the implementation and it is now inside git2-hooks, I've added new functions with suffixed `_with_timeout` for each hook but the `run_hooks` function has a breaking change having the timeout parameter added to it.

The idea would be to leave the `run_hooks()` function alone, and have it call the `_with_timeout()` function internally with a zero-timeout. That way one trivially avoids the breaking change while providing the feature to those who are interested in it.

> I was also thinking about the initial issue where the problem was an infinite wait because the hook was waiting for input, maybe a solution could be when a hook is running a new window is shown with the stdio from the hook and also allowing stdin, not sure if that would work how im thinking though.

I don't think this can be a solution, as hooking up `stdin` must be controlled by the caller. For now, I'd think that hooks inherit stdin from the calling process, and that should work fine even if these have no stdin, as hooks can detect this and even if not they would never be able to hang - reading from a closed file will immediately fail.

Nonetheless I agree that timeouts for hooks may be solving for 'buggy' hooks, and it might be better to fix those (or understand precisely why they maybe hang legitimately so that can be fixed on the side of `git2-hooks` instead).

DaRacci:
> The idea would be to leave the `run_hooks()` function alone, and have it call the `_with_timeout()` function internally with a zero-timeout. That way one trivially avoids the breaking change while providing the feature to those who are interested in it.

I've made this change so run_hooks is left alone functionally and there is now the with_timeout version.

> I don't think this can be a solution, as hooking up `stdin` must be controlled by the caller. For now, I'd think that hooks inherit stdin from the calling process, and that should work fine even if these have no stdin, as hooks can detect this and even if not they would never be able to hang - reading from a closed file will immediately fail.
> 
> Nonetheless I agree that timeouts for hooks may be solving for 'buggy' hooks, and it might be better to fix those (or understand precisely why they maybe hang legitimately so that can be fixed on the side of `git2-hooks` instead).

I think for now its best to track this in a new issue / PR, this one is pretty complete and ready to go as it is.

extrawurst:
@DaRacci please update with master

naseschwarz (changes requested):
Hi @DaRacci. Thank you for your work on this, I appreciate it. This all seems well though-out and usable.

I do have some points regarding some details of your implementation I'd like to discuss before I give the go-ahead.

DaRacci:
Apologies, I'm in the middle of building & swapping to a new computer I'll get back to this when I'm able to.

git2-hooks/src/lib.rs
```diff
@@ -247,7 +346,7 @@ mod tests {
 
 		assert_eq!(res, HookResult::NoHookFound);
 
-		let hook = b"#!/bin/sh
+		let hook = b"#!/usr/bin/env sh
```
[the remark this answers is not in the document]
DaRacci: You are right NixOS does provide a link to /bin/sh by default but I've gotten into the habit of always rewriting it to `/usr/bin/env` as it just works for every interpreter, I can change this back if it causes issues.

src/popups/options.rs
```diff
@@ -99,6 +102,19 @@ impl OptionsPopup {
...
+		Self::add_header(txt, "");
+
+		Self::add_header(txt, "Hooks");
+		self.add_entry(
+			txt,
+			width,
+			"Timeout",
+			&self.options.borrow().hook_timeout().map_or_else(
+				|| "None".to_string(),
+				|d| format!("{d:?}"),
```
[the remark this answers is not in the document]
DaRacci: I was thinking the same thing originally but writing some redundant code or adding a new library to handle it seems like overkill for a single use.

git2-hooks/src/lib.rs
```diff
@@ -144,34 +161,75 @@ pub fn hooks_commit_msg(
 	Ok(res)
 }
 
+/// Git hook: `commit_msg`
+///
+/// See [`hooks_commit_msg`] for more details.
+pub fn hooks_commit_msg_with_timeout(
```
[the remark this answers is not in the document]
DaRacci: I agree unifying making these call the timeout variants does make sense to ensure any changes aren't missed, However you mention a unified interface for git2-hooks, are you implying i should remove the non timeout variants to only leave the timeout versions with an Option Duration in git2-hooks?

git2-hooks/src/hookspath.rs
```diff
@@ -135,48 +138,186 @@ impl HookPaths {
...
+///
+/// Attempt 7:
+///     Sleep Time=(7^2)×1=49
+///         Actual Sleep: 49 milliseconds
+///         Total Sleep: 140 milliseconds
+///
+/// Attempt 8:
+//     Sleep Time=(8^2)×1=64, capped by `MAX_SLEEP_MILLIS` of 50
+//          Actual Sleep: 50 milliseconds
+//          Total Sleep: 190 milliseconds
```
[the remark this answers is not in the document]
DaRacci: I think leaving it were it is makes sense as IDE's will pick up on it when working with the function.

asyncgit/src/sync/hooks.rs
```diff
@@ -224,4 +296,106 @@ mod tests {
...
+	#[test]
+	fn test_run_with_timeout_kills() {
+		let (_td, repo) = repo_init().unwrap();
+		let root = repo.path().parent().unwrap();
+
+		let temp_dir = tempdir().expect("temp dir");
+		let file = temp_dir.path().join("test");
+		let hook = format!(
+			"#!/usr/bin/env sh
+sleep 1
```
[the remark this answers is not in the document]
naseschwarz: Ideally, you'll never hit those 60 s, right, as the timeout would cancel the test before that.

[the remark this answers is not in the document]
DaRacci: You're correct, I had thought this was one of the tests without a timeout.

git2-hooks/src/hookspath.rs
```diff
@@ -290,4 +432,53 @@ mod test {
...
+	#[test]
+	fn test_timeout_with_quadratic_backoff_cost() {
+		let timeout = Duration::from_millis(100);
+		let start = std::time::Instant::now();
+		let result =
+			timeout_with_quadratic_backoff(timeout, || Ok(false));
+		let elapsed = start.elapsed();
+
+		assert_eq!(result.unwrap(), false);
+		assert!(elapsed < timeout + Duration::from_millis(10));
```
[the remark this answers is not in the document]
DaRacci: I've honestly got no idea how to make my own Instant mock or sans-i/o implementation, I've relaxed some of the test timings some more however.

git2-hooks/src/hookspath.rs
```diff
@@ -135,48 +138,186 @@ impl HookPaths {
...
-				stdout,
-				stderr,
-				hook,
-			})
+			if !timeout_with_quadratic_backoff(timeout, || {
+				Ok(child.try_wait()?.is_some())
+			})? {
+				debug!("killing hook process");
+				child.kill()?;
+				return Ok(HookResult::TimedOut { hook });
```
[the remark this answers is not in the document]
DaRacci: I've implemented this with a process group so it actually kills and doesn't wait on under unix but I've got no idea where to start with killing a process tree under windows so it will still wait for children to finish under windows

DaRacci:
I've rebased and addressed most of the suggestions but there's some tests that seems to be failing at random that are unrelated to my changes as they still randomly fail or pass even on the master branch for me.

PR #2548, opened 2025-03-05
Copy text using OSC52

### LINKED ISSUE #2366: Add OSC52 Support

I do development on a remote server using ssh, I love gitui and use it daily on my remote server. it seems that gitui uses xsel on Linux to copy text, but in my case it will give error likes:
```
copy to clipboard error: 
`"xsel"`: /usr/bin/xsel: Can't open display: (null)  
 Inappropriate ioctl for device
```

This error is caused that I do not forward X11 when login into the server using ssh. We know that gitui  does not need to run any graphics since it works great on terminal without X on Linux, and the error will mislead the user.  

As a result, I think gitui should have a better solution to support text copy.

I hope that gitui will support OSC52 encoding. This encoding is supported by many terminals, for example wezterm, iterm.  OSC52 supports you copy text from remote server to your clipboard of local machine.

**Additional context**

vim has a plugin [vim-oscyank](https://github.com/ojroques/vim-oscyank/tree/main) to support OSC52, it should be a good reference implmentation.

naseschwarz:
This Pull Request closes #2366.

It changes the following:
- If copying using wl-clip or xclip and xsel fails, copy using [OSC52](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands) (Search for "Ps = 5 2").

This does not change the default behavior. On both wayland and X, wl-copy and xclip/xsel will be preferred. Only if one path fails, an OSC52 escape sequence will be issued.

I have tested this by running tmux, which has the option to turn support for this off and on (using `:set -g set-clipboard off` or `:set -g set-clipboard on`) and SSH. When tmux's clipboard support was off, no text was copied. Once I enabled it, copying worked as expected. From this I conclude that the whole setup works as expected.

N.B.: There is no easy path for gitui itself to check whether copying worked. Also, this essentially **disables error reporting on wl-clip, xclip and xsel**, as it acts as a fallback for these options.

I followed the checklist:
- [x] I added unittests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 * improve syntax highlighting file detection [[@acuteenvy](https://github.com/acuteenvy)] ([#2524](https://github.com/extrawurst/gitui/pull/2524))
 * After commit: jump back to unstaged area [[@tommady](https://github.com/tommady)] ([#2476](https://github.com/extrawurst/gitui/issues/2476))
+* use OSC52 copying in case other methods fail [[@naseschwarz](https://github.com/naseschwarz)] ([#2366](https://github.com/gitui-org/gitui/issues/2366))
 
 ## [0.27.0] - 2024-01-14
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1162,6 +1162,7 @@ dependencies = [
  "anyhow",
  "asyncgit",
  "backtrace",
+ "base64",
  "bitflags 2.8.0",
  "bugreport",
  "bwrap",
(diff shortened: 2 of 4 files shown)


cruessler:
I’m going to add myself as a reviewer!

cruessler (approved):
I tested this PR on my Linux machine, using Alacritty, after having commented out the `wl-copy` block, forcing the usage of escape codes.

From my side, everything looks good!

extrawurst:
Awesome once the CI is green I will merge. Thank you both!

naseschwarz:
CI failed as my split up copy function was only called on Linux and thus were marked as dead code on Mac OS and Windows. Should be fixed now.

naseschwarz:
I'm really sorry for bothering you because of build failures in CI, @extrawurst. My bad. [I have now set up a CI run in my own namespace and, apart from `cargo deny` because of paste, it succeeds.](https://github.com/naseschwarz/gitui/actions/runs/13854164472/job/38767174816)

extrawurst:
@naseschwarz Thank you!

PR #2563, opened 2025-03-14
Dismiss commit msg key

### LINKED ISSUE #2552: Also support escape for dismissing commit error message
I love using pre-commit hooks and gitui nicely displays the output of the pre-commit hook when it fails:

Image

(the stacked dialogs... so beautiful 🤩)

To close this error, I now have to press _enter_. _Escape_ does not work. The first couple of times, I actually quit gitui because I could not figure out how to close the dialog. Enter just does not compute for me, especially since gitui is pretty consistent with _escape_ closing the current dialog.

I eventually figured out _enter_, but that breaks my brain every time. After pressing _enter_, I am now back in the commit dialog:

Image

I need to edit or stage stuff now, so I want to get rid of the commit dialog. What do I do? My monkey brain tells me: Close dialog again = repeat action = repeat key press. But I am in a text field. Pressing _enter_ will just insert a line break. So for a couple of seconds, I just drool on my keyboard until eventually "close dialog = escape" echoes back to me from the future. Every time I feel like I was briefly unconscious.

I want to close the commit error message dialog with _escape_. I think that would be consistent.

wessamfathi:
This Pull Request fixes/closes #2552.

It changes the following:
- The default key to close the commit error message popup is now the Escape key

I followed the checklist:
- [ ] I added unittests


naseschwarz:
I tested this and the escape key works.

naseschwarz:
The command legend is still set to Enter, though:

image

You probably also want to update `MsgPopup::commands`, replace `strings::commands::close_msg` by `strings::commands::close_popup`.

naseschwarz (approved):
Thank you for the change. I tested it again. It still works fine and displays the correct symbol in the legend.

PR #2569, opened 2025-03-18
upgrade dirs

extrawurst:
seems we have to wait for `shellexpand` to upgrade dirs if we do not want dupes: https://gitlab.com/ijackson/rust-shellexpand


naseschwarz:
shellexpand 3.1.1 has been released with dirs < 7.

image

PR #2571, opened 2025-03-18
Resolve core.hooksPath relative to GIT_WORK_TREE

### LINKED ISSUE #2281: Pre-commit hooks are executed only when launched from the git root
**Describe the bug**
Pre-commit hook is ignored.

**To Reproduce**
1. `mkdir .githooks && git config --local core.hooksPath .githooks`
2. Create `.githooks/pre-commit` with `echo "Hook triggered" && exit 1` and `chmod +x` it.
3. `git commit` now fails, as expected.
4. `gitui` ignores the hook.

**Expected behavior**
`pre-commit` behaviour is the same when using `git` CLI.

**Screenshots**
N/A.

 - OS/Distro + Version: `Debian 12`
 - GitUI Version: `gitui 0.25.2`
 - Rust version: `rustc 1.76.0`

**Additional context**
Relates to:
- https://github.com/extrawurst/gitui/pull/1054
- (maybe) https://github.com/extrawurst/gitui/issues/1711

naseschwarz:
git supports a relative path in core.hooksPath.

`man git-config`:

> A relative path is taken as relative to the directory where the hooks are
> run (see the "DESCRIPTION" section of githooks[5]).

`man githooks`:

> Before Git invokes a hook, it changes its working directory to either
> $GIT_DIR in a bare repository or the root of the working tree in a >
> non-bare repository.

I.e. relative paths in core.hooksPath in non-bare repositories are always relative to GIT_WORK_TREE.

There is a further exception; I believe this is not considered for path resolution:

> An exception are hooks triggered during a push (pre-receive, update,
> post-receive, post-update, push-to-checkout) which are always executed
> in $GIT_DIR.

This Pull Request fixes/closes #2281.

It changes the following:
- Synchronizes core.hooksPath resolution with the git command line client

I followed the checklist:
- [x] I added unittests

### DIFF
--- a/asyncgit/src/sync/hooks.rs
+++ b/asyncgit/src/sync/hooks.rs
@@ -76,6 +76,23 @@ pub fn hooks_prepare_commit_msg(
 mod tests {
 	use super::*;
 	use crate::sync::tests::repo_init;
+	use std::fs::File;
+	use std::io::Write;
+	use std::path::Path;
+
+	fn create_hook_in_path(path: &Path, hook_script: &[u8]) {
+		File::create(path).unwrap().write_all(hook_script).unwrap();
+
+		#[cfg(unix)]
+		{
+			std::process::Command::new("chmod")
+				.arg("+x")
+				.arg(path)
+				// .current_dir(path)
+				.output()
+				.unwrap();
+		}
+	}
 
 	#[test]
 	fn test_post_commit_hook_reject_in_subfolder() {
@@ -167,6 +184,39 @@ mod tests {
 		)
 		.unwrap();
 
+		assert_eq!(
+			res,
+			HookResult::NotOk(String::from("rejected\n"))
+		);
+
+		assert_eq!(msg, String::from("msg\n"));
+	}
+
+	#[test]
+	fn test_hooks_commit_msg_reject_in_hooks_folder_githooks_moved_absolute(
+	) {
+		let (_td, repo) = repo_init().unwrap();
+		let root = repo.path().parent().unwrap();
+		let mut config = repo.config().unwrap();
+
+		const HOOKS_DIR: &'static str = "my_hooks";
+		config.set_str("core.hooksPath", HOOKS_DIR).unwrap();
+
+		let hook = b"#!/bin/sh
+	echo 'msg' > \"$1\"
+	echo 'rejected'
+	exit 1
+	        ";
+		let hooks_folder = root.join(HOOKS_DIR);
+		std::fs::create_dir_all(&hooks_folder).unwrap();
+		create_hook_in_path(&hooks_folder.join("commit-msg"), hook);
+
+		let mut msg = String::from("test");
+		let res = hooks_commit_msg(
+			&hooks_folder.to_str().unwrap().into(),
+			&mut msg,
+		)
+		.unwrap();
 		assert_eq!(
 			res,
 			HookResult::NotOk(String::from("rejected\n"))
--- a/asyncgit/src/sync/hooks.rs
+++ b/asyncgit/src/sync/hooks.rs
@@ -97,7 +97,7 @@ mod tests {
 	#[test]
 	fn test_post_commit_hook_reject_in_subfolder() {
 		let (_td, repo) = repo_init().unwrap();
-		let root = repo.path().parent().unwrap();
+		let root = repo.workdir().unwrap();
 
 		let hook = b"#!/bin/sh
 	echo 'rejected'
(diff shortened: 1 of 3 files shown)


git2-hooks/src/lib.rs
```diff
@@ -98,7 +98,11 @@ pub fn create_hook(
 	path
 }
 
-fn create_hook_in_path(path: &Path, hook_script: &[u8]) {
+/// helper method to create a git hook in a custom path (used in unittests)
+///
+/// # Panics
+/// Panics if hook could not be created
+pub fn create_hook_in_path(path: &Path, hook_script: &[u8]) {
```
extrawurst, line 105: If we just want this to be for testing, we might want to consider feature gating it so it does not become public api which would make it semver breaking to change

naseschwarz: I agree, `git2_hooks::create_hook` already has the same problem.

I've moved `git2_hooks::create_hook_in_path` into `git2-testing`. I've also moved `git2_hooks::create_hook` into a private mod and publish it via `git2_hooks::test_utils::crate_hook` only if the `test-utils` feature flag is set. (I can't move `create_hook` to `git2-testing`, as it has a dependency on "`HookPaths`".)

I'm open to as simpler and cleaner solution. Having this feature flag and the function in test_utils makes this a bit more explicit, but this is of course never tested or even built if all tests on this are run in this workspace with the flag enabled.

naseschwarz: If you want to deal with this discussion separately, I can easily move this into a separate PR by the way.

asyncgit/Cargo.toml
```diff
@@ -18,7 +18,7 @@ dirs = "5.0"
 easy-cast = "0.5"
 fuzzy-matcher = "0.3"
 git2 = "0.20"
-git2-hooks = { path = "../git2-hooks", version = ">=0.4" }
+git2-hooks = { path = "../git2-hooks", version = ">=0.4", features = ["test-utils"] }
```
extrawurst, line 21: wait this means we will always use its test-utils feature, not just for test-runs? is that intended?

extrawurst:
@Byron this might affect Gitbutler. Would love your pair of eyes

asyncgit/Cargo.toml
naseschwarz: Yes. Do you prefer to set this feature flags only for tests?

We then have to always run tests with that feature flag and propagate it from the toplevel crate to git2-hooks. For gitui this is fine, as there's a request for running `make check` in the PR template. It may impede developers who just mechanically run `cargo test --workspace`, unless there is a way to enable a feature flags just for tests, which I don't believe there is.

I assumed enabling it for the workspace was okay, as your concern was semver breaking changes and dependencies are fixed within the gitui workspace. The way this is organized right now is not worse than before, at least, with semver concerns alleviated.

Thank you for looking at this.

naseschwarz: That being said, maybe we should just copy/paste the function to asyncgit for tests and don't waste time on discussing or publishing a random feature flag that sounds fishy.

extrawurst: yeah let’s do that 👍 simple and pragmatic

asyncgit/src/sync/hooks.rs
```diff
@@ -174,4 +191,37 @@ mod tests {
 
 		assert_eq!(msg, String::from("msg\n"));
 	}
+
+	#[test]
+	fn test_hooks_commit_msg_reject_in_hooks_folder_githooks_moved_absolute(
+	) {
+		let (_td, repo) = repo_init().unwrap();
+		let root = repo.path().parent().unwrap();
```
Byron, line 199: `repo.workdir()` should rather be used here.

Byron (commented):
Thanks for calling me in @extrawurst , and thanks @naseschwarz for tackling this!

I'd arrive at the same conclusion on how the implementation should look like, and it's great to see all the right tests.

Of course, I'd have a few nits but they are not material.

It's worth noting that `core.hooksPath` is very relevant to security on shared drives in particular, and since `git2` follows Git and will refuse to open such repositories it should be safe to use `core.hooksPath` here. In `gitoxide` one would use [`trusted_path()`](https://docs.rs/gix/latest/gix/config/struct.Snapshot.html#method.trusted_path) instead to obtain the fully-interpolated path if it's in a trusted configuration file.

naseschwarz:
Thank you for your time and input, @Byron.

extrawurst:
Thank you!

Byron:
Would it be possible to have a release for this?

If there is a different schedule, I am OK to use `git2-hooks` as Git dependency as well, but thought I could ask first 😅.

extrawurst:
@Byron we should probably move this into its own repo to simplify visibility and separate release schedule. will cut a release now

extrawurst:
@Byron released as git2-hooks `0.5.0`

PR #2591, opened 2025-04-01
Draft PoC of blame using gitoxide

### LINKED ISSUE #2587: Interactive Blame doesn't appear to be as fast as tig or non-interactive git blame
Hi, I came to check out this project because I'm looking for a tool for command line interactive blame. I do a lot of source code archaeology , and I would like a single key-press way to go back to the commit that introduced the line in my cursor. As an LLVM developer, I was super excited to try out a project advertised as "blazing fast Rust". :)

However, I ran gitui blame on [llvm-project/llvm/lib/Target/X86/X86ISelLowering.cpp,](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/X86/X86ISelLowering.cpp) (helpfully github says "(Sorry about that, but we can’t show files that are this big right now.)") and it's still going as I type out this issue / feature request. This isn't a bug so much as, if you are interested in doing further performance engineering and looking for a workload to benchmark, profile, and optimize, I think you should consider this one.

This source file may be a crime against software engineering, but I will say I've done my part to move code out of it, although it was almost entirely done in self-interest.

cruessler:
This is an early-stage PR to explore how difficult it will be to integrate `gix-blame` into `gitui` and how it will compare to the existing implementation, in particular with respect to performance.

A detailed description and measurements will follow.


extrawurst:
@cruessler do you think it would be possible to aim to make that the headline feature for `0.29`?

cruessler:
@extrawurst I’m pretty sure that’s a realistic goal!

We would only have to decide how much tolerance we have for `gix-blame` differing from `git blame` due to differences in the underlying diff algorithm. (If you haven’t seen yet, there’s this tracking issue for differences in the diff algorithm: https://github.com/GitoxideLabs/gitoxide/issues/2308.)

As far as I can tell, things have improved significantly in that area recently, though. Behind a feature flag, I’ve updated the diff library used by `gix-diff`, `imara-diff`, to a new version, 0.2. Using 0.2, `gix-blame` matches `git blame` on around 95 % of all lines in the `gitoxide` repository I tested this with (off the top of my head, I still want to document that somewhere). I’m planning on pushing a commit that uses that feature flag in this PR this weekend.

I hope that makes sense, let me know if you want more context!

cruessler:
Before declaring this ready for review, I want to run a couple of benchmarks.

Issue #2611, opened 2025-04-15
Use gitoxide crates to simplify code


Joshix-1:
1. in `HooksPaths::run_hook_os_str` in `git2-hooks/hookspath.rs`

              This is also a capability of `gix-command`, but it's also just done if the command can be converted to bytes at least.

_Originally posted by @Byron in https://github.com/gitui-org/gitui/pull/2483#discussion_r2044024971_

2. in `sh_path` in `git2-hooks/hookspath.rs`

              One could consider to use `gix_path::env::shell()` here.

_Originally posted by @Byron in https://github.com/gitui-org/gitui/pull/2483#discussion_r2044028443_

Byron:
Thanks for keeping track of this.

I think it's fair to close this issue once #2612 is merged as well, there is no need to force `gix` in. The plan is to eventually offer `gix-hooks` anyway.

extrawurst:
Closed as of #2612

PR #2612, opened 2025-04-15
use gix_path::env::shell() to get the shell in git2-hooks

Joshix-1:
This deals with one half of #2611

It changes the following:
-
-

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/git2-hooks/Cargo.toml
+++ b/git2-hooks/Cargo.toml
@@ -14,7 +14,7 @@ keywords = ["git"]
 
 [dependencies]
 git2 = ">=0.17"
-gix-path = "0.10.15"
+gix-path = "0.10"
 log = "0.4"
 shellexpand = "3.1"
 thiserror = "2.0"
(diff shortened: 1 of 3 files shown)


git2-hooks/Cargo.toml
```diff
@@ -14,6 +14,7 @@ keywords = ["git"]
 
 [dependencies]
 git2 = ">=0.17"
+gix-path = "0.10.15"
```
naseschwarz, line 17: Let's not depend on the exact patch version here, same as for the other dependencies.

```suggestion
gix-path = "0.10"
```

naseschwarz (changes requested):
Thanks @Joshix-1. I have just one remark.

naseschwarz (approved):

Byron (approved):
LGTM! Thanks again!

*Please note that `gix-path` will probably increment its major version every couple of months, but I don't expect any actual breakage here so that should be fine.*

extrawurst:
Thanks everyone!

Issue #2488, opened 2025-01-16
GitUI was close due to an unexpected panic (I think it is an async issue)


maxtremblay:
**Describe the bug**
I can't open gitui.

**To Reproduce**

Go in a git project and launch gitui. Then, navigate to the second tab (log). Then, it closes with the error message and I can't open it anymore.

**Expected behavior**
Launch gitui.

**Screenshots**

![Image](https://github.com/user-attachments/assets/ed36e0b5-ba42-42e5-811d-e42583b7b3dd)

**Context (please complete the following information):**
 - OS/Distro + Version: Arch Linux (everything is up-to-date as Jan 16th, 2025).
 - GitUI Version: Unsure, when I install with pacman I have 
```
❯ pacman -Qs gitui
local/gitui 0.26.3-2
    Blazing fast terminal-ui for git written in Rust
```

At the same time 
```
❯ gitui --version
gitui nightly 2024-09-12 (0f5cf89)
```

 - Rust version: 1.84

**Additional context**

This is the content of the log file when running `gitui -l`

```
14:51:31 [TRACE] (1) gitui::app: [src/app.rs:154] open repo at: RefCell { value: Path(".") }
14:51:31 [TRACE] (6) mio::poll: [/build/.cargo/registry/src/index.crates.io-6f17d22bba15001f/mio-0.8.11/src/poll.rs:551] registering event source with poller: token=Token(0), interests=READABLE
14:51:31 [TRACE] (6) mio::poll: [/build/.cargo/registry/src/index.crates.io-6f17d22bba15001f/mio-0.8.11/src/poll.rs:551] registering event source with poller: token=Token(1), interests=READABLE
14:51:31 [TRACE] (1) asyncgit::tags: [asyncgit/src/tags.rs:67] request
14:51:31 [TRACE] (1) gitui: [src/main.rs:215] app start: 14 ms
14:51:31 [TRACE] (1) gitui::app: [src/app.rs:375] update
14:51:31 [TRACE] (1) asyncgit::tags: [asyncgit/src/tags.rs:67] request
14:51:31 [ERROR] 
GitUI was close due to an unexpected panic.
Please file an issue on https://github.com/extrawurst/gitui/issues with the following info:

PanicHookInfo { payload: Any { .. }, location: Location { file: "asyncgit/src/revlog.rs", line: 186, col: 14 }, can_unwind: true, force_no_backtrace: false }
trace:
   0: <unknown>
   1: <unknown>
   2: <unknown>
[...]

PR #2617, opened 2025-04-16
Improve error messages

### LINKED ISSUE #2599: path ownership rights issue leads to invalid bare repo detection
When I run gitui, I get the following error:

```
invalid path
please run gitui inside of a non-bare git repository
```

The repository is not bare! git status reports no problems.

Following the docs, I run gitui with logging enabled

`gitui -l`

The log file contains the following line:

```
[ERROR] repo open error: repository path [redacted] is not owned by current user; class=Config (7); code=Owner (-36)
```

This appears to be an error directly from git, which is related to this CVE 2022-24765 (https://nvd.nist.gov/vuln/detail/cve-2022-24765)

The workaround I found was to configure git to treat the directory as safe.

The global approach is:

```
git config --global --add safe.directory *
```

This defeats the extra safety checks added by git to address cve-2022-24765.

In my single-user-machine context, I am happy with that risk.

In a true multi-user environment, and more fine-grained approach would be recommended.

acuteenvy:
This Pull Request fixes/closes #2599.

It changes the following:
- Improve error messages when the repository cannot be opened.
  - for example, when the repo doesn't exist:
  ```sh
  # previously
  $ gitui
  invalid path
  please run gitui inside of a non-bare git repository
  
  # now
  $ gitui
  Error: could not find repository at '.'; class=Repository (6); code=NotFound (-3)
  ```
- Remove the panic handler from rayon, since `panic::set_hook` already handles all panics. Setting both panic handlers causes the panic message to be displayed twice.
- Do both `log::error!` and `eprintln!` wherever possible

I followed the checklist:
- [ ] I added unittests
  - ~~I get some errors in `Cargo.toml`, but it hasn't been changed in this PR:~~ (see https://github.com/gitui-org/gitui/pull/2617#issuecomment-2810329134)
  ```
  error: Dependencies for . are not sorted
  error: Cargo.toml for . is not formatted
  Checking asyncgit...
  Checking filetreelist...
  Checking git2-hooks...
  Checking git2-testing...
  Checking scopetime...
  error: Dependencies for scopetime are not sorted
  error: Cargo.toml for scopetime is not formatted
  make: *** [Makefile:108: sort] Error 1
  ```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -73,7 +73,6 @@ use std::{
 	io::{self, Stdout},
 	panic,
 	path::Path,
-	process,
 	time::{Duration, Instant},
 };
 use ui::style::Theme;
@@ -120,20 +119,24 @@ enum Updater {
 	NotifyWatcher,
 }
 
+/// Do `log::error!` and `eprintln!` in one line.
+macro_rules! log_eprintln {
+	( $($arg:tt)* ) => {{
+		log::error!($($arg)*);
+		eprintln!($($arg)*);
+	}};
+}
+
 fn main() -> Result<()> {
 	let app_start = Instant::now();
 
 	let cliargs = process_cmdline()?;
 
 	asyncgit::register_tracing_logging();
-
-	if !valid_path(&cliargs.repo_path) {
-		eprintln!("invalid path\nplease run gitui inside of a non-bare git repository");
-		return Ok(());
-	}
+	ensure_valid_path(&cliargs.repo_path)?;
 
 	let key_config = KeyConfig::init()
-		.map_err(|e| eprintln!("KeyConfig loading error: {e}"))
+		.map_err(|e| log_eprintln!("KeyConfig loading error: {e}"))
 		.unwrap_or_default();
 	let theme = Theme::init(&cliargs.theme);
 
@@ -142,7 +145,7 @@ fn main() -> Result<()> {
 		shutdown_terminal();
 	}
 
-	set_panic_handlers()?;
+	set_panic_handler()?;
 
 	let mut repo_path = cliargs.repo_path;
 	let mut terminal = start_terminal(io::stdout(), &repo_path)?;
@@ -292,13 +295,13 @@ fn shutdown_terminal() {
 		io::stdout().execute(LeaveAlternateScreen).map(|_f| ());
 
 	if let Err(e) = leave_screen {
-		eprintln!("leave_screen failed:\n{e}");
+		log::error!("leave_screen failed:\n{e}");
 	}
 
 	let leave_raw_mode = disable_raw_mode();
 
 	if let Err(e) = leave_raw_mode {
-		eprintln!("leave_raw_mode failed:\n{e}");
+		log::error!("leave_raw_mode failed:\n{e}");
 	}
 }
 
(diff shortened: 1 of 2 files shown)


naseschwarz:
> error: Dependencies for scopetime are not sorted

That's a good one. There is only one dependency in `scopetime/Cargo.toml`. If I run that on you branch, by the way, I don't get any errors.

acuteenvy:
I just found out why - `cargo-sort` v1.1.0 [is broken and has been yanked from crates.io](https://github.com/DevinR528/cargo-sort/releases/tag/v1.1.0). I installed the broken version from the Arch repositories.

Installing v1.0.9 fixes this issue.

src/main.rs
```diff
@@ -389,30 +394,15 @@ fn start_terminal(
...
+		log_eprintln!("\nGitUI was closed due to an unexpected panic.\nPlease file an issue on https://github.com/gitui-org/gitui/issues with the following info:\n\n{e}\n\ntrace:\n{backtrace:?}");
 	}));
 
 	// global threadpool
 	rayon_core::ThreadPoolBuilder::new()
-		.panic_handler(|e| {
-			let backtrace = Backtrace::new();
-			shutdown_terminal();
-			log_eprintln!("\nGitUI was close due to an unexpected panic.\nPlease file an issue on https://github.com/gitui-org/gitui/issues with the following info:\n\n{:?}\ntrace:\n{:?}", e, backtrace);
-			process::abort();
```
naseschwarz, line 414: You got a panic twice if you kept this panic handler here? But this aborts the process? Would you please be so kind and provide some steps to reproduce, please?

Also, I tested this with and without this handler and I cannot find a situation where the panic is printed twice. Also, if this handler isn't there, the terminal is not reset into normal mode if a panic occurs in a rayon thread. (This is why this handler was introduced here originally.)

naseschwarz (changes requested):
Thanks for your work on this. This clearly improves error handling. For a non-repo folder, we get / got:

image

For dubious ownership, the original cause here, we get / got /git says:

image

I'd almost argue that the new message is better than what git says. "Let's just ignore this security feature with this command" maybe is not soooo cool. ;)

I have to pick on one thing, though. The panic handler was there for a reason, as far as I can tell. See my remark below.

naseschwarz:
Also, would you please be so kind and add a changelog entry?

src/main.rs
acuteenvy: > Would you please be so kind and provide some steps to reproduce?

Just put `panic!()` somewhere inside `rayon_core::spawn` (e.g. in `asyncgit/src/revlog.rs`). I first saw the double panics in some issues, #2488 for instance.

> if this handler isn't there, the terminal is not reset into normal mode if a panic occurs in a rayon thread

The terminal will be reset by the normal panic handler, which runs even when the panic occurs in a rayon thread.

naseschwarz: > > if this handler isn't there, the terminal is not reset into normal mode if a panic occurs in a rayon thread
> 
> The terminal will be reset by the normal panic handler, which runs even when the panic occurs in a rayon thread.

You're right. I must have made a mistake when testing that behavior yesterday. Sorry about that.

I'll leave the duplicate panic business to someone else, though. It feels like I'm missing something here.

naseschwarz (approved):
I'm not sure about the duplicate panic issue. @extrawurst, you added this initially, you probably have an opinion yourself.

src/main.rs
extrawurst: wow this is ancient history, got introduced in 74a86ba - did we check that this behaves well in case of a panic in and outside of rayon now with this change? if so i am cool with simplyfing it @naseschwarz

extrawurst:
Ok I tested the extra error handler for rayon is not needed as far as i can see. removed in on master. @acuteenvy please rebase and then we can merge this

extrawurst (approved):

PR #2620, opened 2025-04-17
Show cursor on panic

naseschwarz:
ratatui::Terminal starts by hiding the cursor. If we panic and abort, that Terminal instance is not dropped, which leaves restoring the cursor state to us.

It changes the following:
- Shows the terminal cursor if gitui panics

I followed the checklist:
- ~~I added unittests~~
- ~~I added an appropriate item to the changelog~~
  - left out intentionally for such a minor item

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -288,8 +288,10 @@ fn setup_terminal() -> Result<()> {
 }
 
 fn shutdown_terminal() {
+	let mut stdout = io::stdout();
+
 	let leave_screen =
-		io::stdout().execute(LeaveAlternateScreen).map(|_f| ());
+		stdout.execute(LeaveAlternateScreen).map(|_f| ());
 
 	if let Err(e) = leave_screen {
 		eprintln!("leave_screen failed:\n{e}");
@@ -299,6 +301,10 @@ fn shutdown_terminal() {
 
 	if let Err(e) = leave_raw_mode {
 		eprintln!("leave_raw_mode failed:\n{e}");
+	}
+
+	if let Err(e) = stdout.execute(crossterm::cursor::Show) {
+		eprintln!("Showing cursor failed:\n{e}");
 	}
 }
 


src/main.rs
```diff
@@ -300,6 +302,10 @@ fn shutdown_terminal() {
 	if let Err(e) = leave_raw_mode {
 		eprintln!("leave_raw_mode failed:\n{e}");
 	}
+
+	if let Err(e) = stdout.execute(crossterm::cursor::Show) {
```
extrawurst, line 306: i am confused. since we never do `cursor::Hide` why do we have to explicitly do `Show`? so far this happened implicitly if doing `LeaveAlternateScreen` right?

extrawurst: Ok i now see what you mean. forget what i said :)

extrawurst:
I think we should revisit the whole panic handlings. this is a bandaid over the fact that we are not having propper dropping happen which is only the case if we abort instead of panic.

ideally we find out how to handle the rayon panic in a way that we still end the program but run all the drops properly

extrawurst:
@naseschwarz reverted this now again after pushing 9271b41 - i could not produce a case where the cursor was not shown again since we now should be running drops for everything (including terminal) but please have another look

naseschwarz:
Thanks. I actually tested this PR without the rayon handler too, as I was expecting this to be merged sooner or later. On my terminal, alacritty on Linux, the cursor is still hidden, when I do the following:

1. checkout master
2. apply patch from below
3. cargo run
4. go to tab 2, revlog
5. look at the diff for the just applied pach

-> panic

-> cursor still hidden

```
From d8deeb2125b51efa02cde590bfcc9605a5d684c8 Mon Sep 17 00:00:00 2001
From: Naseschwarz <naseschwarz@0x53a.de>
Date: Wed, 16 Apr 2025 20:22:33 +0200
Subject: [PATCH] Panic on diff view

---
 asyncgit/src/diff.rs | 1 +
 1 file changed, 1 insertion(+)

diff --git a/asyncgit/src/diff.rs b/asyncgit/src/diff.rs
index 5316e67b..1d98ea00 100644
--- a/asyncgit/src/diff.rs
+++ b/asyncgit/src/diff.rs
@@ -122,6 +122,7 @@ impl AsyncDiff {
 		self.pending.fetch_add(1, Ordering::Relaxed);

 		rayon_core::spawn(move || {
+			panic!();
 			let notify = Self::get_diff_helper(
 				&repo,
 				params,
--
2.49.0
```

naseschwarz:
Also, grepping through a capture of everything that's written to the output, at least CSI 25h, which would show the cursor, is never printed in `master` on panic right now:

```
% grep 25h terminal-log | xxd | wc -l
0
```

PR #1652, opened 2023-04-13
Simplify theme overrides

cruessler:
This Pull Request closes #1367.

It changes the following:

Theme overrides are now loaded the same way key overrides are loaded. The config file, `theme.ron`, does not have to contain a complete theme anymore. Instead, it is possible to specify only the values that are supposed to override their corresponding default values.

This PR is marked as draft because this change is breaking, and their is no automatic migration for existing configs. If that’s ok, I would extend the `Breaking Changes` entry, otherwise I would add code to migrate existing themes.

Update 2023-06-25: I added code for migrating existing themes to the new format.

I followed the checklist:

- [x] I added unittests
- [x] I ran `make check` without errors
- [x] I tested the overall application
- [x] I added an appropriate item to the changelog

PR #2627, opened 2025-04-29
Feat/respect xdg spec on all os

### LINKED ISSUE #1498: Follow XDG spec on windows

KlassyKat:
This Pull Request fixes/closes #1498.

It changes the following:
- Unifies OS logic for loading config.
- Adds a ordered list of locations to search for config.
- List includes `XDG_CONFIG_HOME` as the first priority.
- Implements the same logic for getting the cache_path with `XDG_CACHE_HOME` as the top priority
- No longer creates an empty `gitui` config folder by default.

I followed the checklist:
- [x] I added unittests

### DIFF
--- a/KEY_CONFIG.md
+++ b/KEY_CONFIG.md
@@ -19,12 +19,17 @@ Create a `key_bindings.ron` file like this:
 )
 ```
 
-The config file format based on the [Ron file format](https://github.com/ron-rs/ron).
+The keybinding config file uses the [Ron file format](https://github.com/ron-rs/ron).
 The location of the file depends on your OS:
-* `$HOME/.config/gitui/key_bindings.ron` (mac)
-* `$XDG_CONFIG_HOME/gitui/key_bindings.ron` (linux using XDG)
-* `$HOME/.config/gitui/key_bindings.ron` (linux)
-* `%APPDATA%/gitui/key_bindings.ron` (Windows)
+`gitui` will look for an existing `/gitui` in the following order:
+* `$XDG_CONFIG_HOME/gitui/` (with `XDG_CONFIG_HOME` set)
+* `$HOME/.config/gitui/`
+* Default OS Location:
+    * `$HOME/Library/Application Support/` (mac)
+    * `$HOME/.config/gitui/` (linux)
+    * `%APPDATA%/gitui/` (Windows)
+
+Key bindings are configured in `key_bindings.ron` within your first found `gitui` config folder.
 
 See all possible keys to overwrite in gitui: [here](https://github.com/gitui-org/gitui/blob/master/src/keys/key_list.rs#L83)
 
--- a/src/args.rs
+++ b/src/args.rs
@@ -159,15 +159,42 @@ fn get_app_cache_path() -> Result<PathBuf> {
 }
 
 pub fn get_app_config_path() -> Result<PathBuf> {
-	let mut path = if cfg!(target_os = "macos") {
-		dirs::home_dir().map(|h| h.join(".config"))
-	} else {
-		dirs::config_dir()
-	}
-	.ok_or_else(|| anyhow!("failed to find os config dir."))?;
+	// List of potential config directories in order of priority
+	let potential_config_dirs = [
+		env::var_os("XDG_CONFIG_HOME").map(PathBuf::from),
+		// This is in the list since it was the hardcoded behavior on macos before
+		// I expect this to be what most people have XDG_CONFIG_HOME set to already
+		// But explicitly including this will avoid breaking anyone's existing config
+		dirs::home_dir().map(|p| p.join(".config")),
+		dirs::config_dir(),
+	]
+	.into_iter()
+	// Remove any that resulted in None from unset env vars
+	.flatten()
+	// Remove any paths that aren't found on the system
+	.filter(|path| path.is_dir());
 
-	path.push("gitui");
-	Ok(path)
+	let mut target_config_dir = None;
+
+	for potential_dir in potential_config_dirs {
+		let search_path = potential_dir.join("gitui");
+
+		// Prefer any preexisting gitui config dir
+		if search_path.is_dir() {
+			target_config_dir = Some(search_path);
+			break;
+		}
+
+		// Set fallback to first existing directory
+		if target_config_dir.is_none() {
+			target_config_dir = Some(search_path);
+		}
+	}
+
+	let config_dir = target_config_dir
+		.ok_or_else(|| anyhow!("failed to find os cache dir."))?;
+
+	Ok(config_dir)
 }
 
 #[test]
--- a/src/args.rs
+++ b/src/args.rs
@@ -150,16 +150,14 @@ fn setup_logging(path_override: Option<PathBuf>) -> Result<()> {
 }
 
 fn get_path_from_candidates(
-	candidates: Vec<Option<PathBuf>>,
+	candidates: impl IntoIterator<Item = Option<PathBuf>>,
 ) -> Result<PathBuf> {
-	let valid_candidates = candidates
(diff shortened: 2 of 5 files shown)


KlassyKat:
> No longer creates an empty `gitui` config folder by default.
I'm a little trepidatious about this particular change since I am not certain there is no logic relying on a config folder existing; though as far as I can tell there isn't.

If it doesn't present any issues (or even if it does present minor issues) I am a strong advocate for not forcing an empty `gitui` config folder to exist for any users that don't feel any need to configure `gitui`. I see this as an extremely annoying UX for anyone that doesn't want/feel the need for any configuration of `gitui` and likes to keep their dotfiles clean.

For an app as functional as `gitui` out of the box I don't see the portion of users who don't feel the need for any additional config as a particularly small group.

src/args.rs
```diff
@@ -149,28 +148,104 @@ fn setup_logging(path_override: Option<PathBuf>) -> Result<()> {
...
-		dirs::config_dir()
+	// List of potential config directories in order of priority
+	let config_dir_candidates = [
+		env::var_os("XDG_CONFIG_HOME").map(PathBuf::from),
+		// This is in the list since it was the hardcoded behavior on macos before
+		// I expect this to be what most people have XDG_CONFIG_HOME set to already
+		// But explicitly including this will avoid breaking anyone's existing config
+		dirs::home_dir().map(|p| p.join(".config")),
+		dirs::config_dir(),
+	];
```
KlassyKat, line 199: I know that I put this right there in the comments but I just wanted to make sure this was mentioned at the top of the thread.

`dirs::home_dir().map(|p| p.join(".config"))`

Is only included in the list since this was the previous hard coded behavior on macOS. This is exactly the location I expect most people to have `XDG_CONFIG_HOME` set to on Windows, macOS, and Linux. But for any macOS users who don't have `XDG_CONFIG_HOME` set, their existing `gitui` config would break if this was not included.

I don't see this as particularly problematic, as if the `$HOME/.config` directory exists on a users system it is a very reasonable place to search for config files. So the only real cost of this I see is a couple of line of extra verbosity in the code and the docs, in exchange for not breaking macOS users existing config.

KlassyKat:
Could also add a `$GITUI_CONFIG_DIR` env var look for at the top of the priority list, but I see that as maybe a bit overkill for what is a relatively simple config structure.

src/args.rs
```diff
@@ -149,28 +148,104 @@ fn setup_logging(path_override: Option<PathBuf>) -> Result<()> {
...
 }
 
 pub fn get_app_config_path() -> Result<PathBuf> {
-	let mut path = if cfg!(target_os = "macos") {
-		dirs::home_dir().map(|h| h.join(".config"))
-	} else {
-		dirs::config_dir()
+	// List of potential config directories in order of priority
+	let config_dir_candidates = [
+		env::var_os("XDG_CONFIG_HOME").map(PathBuf::from),
```
naseschwarz, line 193: At least according to spec, `XDG_CONFIG_HOME` should only be considered if it's absolute. (Same for `_CACHE`...)

src/args.rs
naseschwarz: I don't particularly fancy this approach, to be honest. I get the original request to have gitui honor `%XDG_CONFIG_HOME%` on Windows if it is set, as Windows is a lost cause anyhow. However, this also introduces the special handling of this variable into the MacOS and Linux path, which just don't need it at all?

src/args.rs
```diff
@@ -149,28 +148,104 @@ fn setup_logging(path_override: Option<PathBuf>) -> Result<()> {
...
+		// This is in the list since it was the hardcoded behavior on macos before
+		// I expect this to be what most people have XDG_CONFIG_HOME set to already
+		// But explicitly including this will avoid breaking anyone's existing config
+		dirs::home_dir().map(|p| p.join(".config")),
+		dirs::config_dir(),
+	];
+
+	get_path_from_candidates(config_dir_candidates)
+		.map_err(|_| anyhow!("failed to find os config dir."))
```
naseschwarz, line 202: It's not really the "**os** config dir" we were looking for, right?

KEY_CONFIG.md
```diff
@@ -19,12 +19,17 @@ Create a `key_bindings.ron` file like this:
 )
 ```
 
-The config file format based on the [Ron file format](https://github.com/ron-rs/ron).
+The keybinding config file uses the [Ron file format](https://github.com/ron-rs/ron).
```
naseschwarz, line 22:
```suggestion
The key binding config file uses the [Ron file format](https://github.com/ron-rs/ron).
```

naseschwarz (changes requested):
Thank you for your PR!

I'm not a regular Windows user, so I didn't test the new behavior. The following is just based on code review.

> I'm a little trepidatious about this particular change since I am not certain there is no logic relying on a config folder existing; though as far as I can tell there isn't.

I agree with that. There was a feature that created a null theme file a while back, but that has been removed in #1652. No need to create an empty directory.

I have a few remarks regarding the implementation.

src/args.rs
naseschwarz: Ah, forget it. If you fix the absolute path remark, it should work. No need to be too pedantic here.

naseschwarz: So, one remark is left: I would *not* be a fan of this, if we were to keep creating config directories, as this would put the `.gitui` option first on windows if `XDG_CONFIG_HOME` weren't set.

src/args.rs
```diff
@@ -149,28 +148,104 @@ fn setup_logging(path_override: Option<PathBuf>) -> Result<()> {
...
 }
 
+fn get_path_from_candidates(
+	candidates: impl IntoIterator<Item = Option<PathBuf>>,
+) -> Result<PathBuf> {
+	let mut target_dir = None;
+
+	// Filter into existing directories
+	for potential_dir in
+		candidates.into_iter().flatten().filter(|p| p.is_dir())
```
naseschwarz, line 158: Sorry, one more remark: At least for XDG, `Path::is_dir` isn't a precondition for considering that path. If it doesn't exist, create it.

naseschwarz (changes requested):

src/args.rs
KlassyKat: > If, when attempting to write a file, the destination directory is non-existent an attempt should be made to create it with permission 0700.

This only applies when attempting to write a file.

> When attempting to read a file, if for any reason a file in a certain directory is unaccessible, e.g. because the directory is non-existent, the file is non-existent or the user is not authorized to open the file, then the processing of the file in that directory should be skipped.

So believe this point is fairly moot when it comes to `XDG_CONFIG_HOME`.

That being said it does have some consideration for `XDG_CACHE_HOME`, but I think there is still some room for interpretation. I think you could consider that line to be referring to the creation of an app subdirectory and not to create the root directory itself. In which case that is already being done. Furthermore when it comes to creating good UX a user may have an `XDG_CACHE_HOME` defined which would result in that being the location to write to by default. But say the user wanted to keep `XDG_CACHE_HOME` only having the logs of particular apps creating a `gitui` folder in the fallback cache directory would allow them to filter it out.

I don't really imagine anyone wanting to do that, but it seems superior to have the option than not to.

naseschwarz: Yes, for `XDG_CONFIG_HOME` it does not make a difference as of now, as we don't create things there. For `XDG_CACHE_HOME` it does, though.

PR #2628, opened 2025-04-29
Wip 81 log graph

peso:
This Pull Request is wip towards fixing #81 

It changes the following:
- work-in-progrss for adding git-graph DO NOT MERGE

Lots of bugs and strange behaviour


peso:
Right now I'm working on refactoring git-graph so the git backend can be swapped. It is based on git2 and gitui is moving away from git2 towards gix.

yk-kd:
Fonts can be made configurable to display graphs more neatly.
So it would be great if you could make the letter of the join configurable.

https://github.com/kovidgoyal/kitty/pull/7681
https://github.com/rbong/vim-flog
https://www.reddit.com/r/neovim/comments/1ec118n/git_graph_teaser/

peso:
The symbols are already configurable. [git-graph](https://github.com/mlange-42/git-graph) (not the vs code plugin "git graph") has a configurable option called "style" to change the symbols used. 

Regarding the specific symbols you link to, [git-graph issue 90](https://github.com/mlange-42/git-graph/issues/90) requests the enhancement you propose. Note however, that the proposed symbol-set is larger than the current symbol-set and thus require some changes at the algorithm level. It is a bit more work than simply changing the symbols.

yk-kd:
Sorry if I'm wrong.
>     -s, --style <style>      Output style. One of [normal/thin|round|bold|double|ascii].
I thought style could only be set up from these five, since it was written above.

peso:
You are right. As a user you can only change symbols by picking one of the predefined styles. There is no support for changing a single symbol (eg. "join") in a style.

peso:
The current code is a little more stable than that from a year ago. It still has a lot of rough edges and performance is much better but still not good enough if too many branches are present.

PR #2638, opened 2025-05-12
Avoid exposing internal tuple in CommitList.marked

peso:
This Pull Request is a minor cleanup

It changes the following:
- Use a different function to avoid exposing the tuple used to sort commit id in CommitList.marked

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/src/components/commitlist.rs
+++ b/src/components/commitlist.rs
@@ -45,6 +45,8 @@ pub struct CommitList {
 	items: ItemBatch,
 	highlights: Option<Rc<IndexSet<CommitId>>>,
 	commits: IndexSet<CommitId>,
+	/// Commits that are marked. The .0 is used to provide a sort order.
+	/// It contains an index into self.items.items
 	marked: Vec<(usize, CommitId)>,
 	scroll_state: (Instant, f32),
 	tags: Option<Tags>,
--- a/src/components/commitlist.rs
+++ b/src/components/commitlist.rs
@@ -117,15 +117,6 @@ impl CommitList {
 	}
 
 	///
-	#[expect(
-		clippy::missing_const_for_fn,
-		reason = "as of 1.86.0 clippy wants this to be const even though that breaks"
-	)]
-	pub fn marked(&self) -> &[(usize, CommitId)] {
-		&self.marked
-	}
-
-	///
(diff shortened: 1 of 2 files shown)


src/components/commitlist.rs
```diff
@@ -45,6 +45,8 @@ pub struct CommitList {
 	items: ItemBatch,
 	highlights: Option<Rc<IndexSet<CommitId>>>,
 	commits: IndexSet<CommitId>,
+	/// Commits that are marked. The .0 is used to provide a sort order.
```
naseschwarz, line 48: First half is obvious, second half is its usage, not a description of .0

naseschwarz (changes requested):
Thank you for your PR. I have one item to pick on - if we add comments, they should be accurate.

src/components/commitlist.rs
peso: I think usage is important in this case. 

The .0 was really strange to me when I discovered it in code. After seeing that it was the index in the commit list, it made even less sense to me. It was not until I saw it being sorted in `fn mark()` that I figured why it was added - to make the list order stable to insertions and deletions, so I think the motivation for having it there is important. 

I considered changing it to `m̀arked: Vec<CommitId>` (no pair) which would be easier to understand.  The sorting would then have to look up the index using `self.commits.get_index_of()` but I expected somebody to argue that it would be some clock cycles slower.

I also wanted to keep the comment condensed - as few non-obvious parts as possible. The usage of .0 was the only thing not obvious to me.

If you prefer a more verbose comment that puts the "what" (description of content) before the "why" (description of function), how about this:
```
/// The marked commits.
/// .0 holds the commit index into self.items.items - used for ordering the list
/// .1 is the commit id of the marked commit
```

peso: @naseschwarz I'm uncertain how you would like me to proceed. Can you suggest a change that would be needed for you to approve the PR? Do you want the comment deleted, expanded, reworded?

extrawurst: @peso lets to the more verbose comment

extrawurst (changes requested):
lets fix the comment and get it merged

PR #2639, opened 2025-05-12
Minimal docs at module level

peso:
This Pull Request provides minimal top-level documentation.
This makes the initial navigation of cargo docs easier.

It changes the following:
- add module comment to main
- add module comment to tabs
- add module comment to cmdbar
- add module comment to components
- add module comment to components::commitlist
- add crate comments to asyncgit

I followed the checklist:
- [ ] I added unittests


peso:
Next time, I should remember running rustfmt and cargo clippy before pushing ...

asyncgit/src/lib.rs
```diff
@@ -1,4 +1,10 @@
-//! asyncgit
+/*!
+`AsyncGit` is a library that provides non-blocking access to Git
```
extrawurst, line 2: technically not accurate because we also offer a bunch of `sync` apis

src/components/mod.rs
```diff
@@ -1,3 +1,7 @@
+/*!
+Components are the visible building blocks of tabs.
```
extrawurst, line 2: not adding much value

peso: I disagree, mostly because of the word "component".  For someone who is not familiar with the code base components could be anything. 

One thing that got me sidetracked at the beginning was that "gitui::components" sits next to "gitui::ui". My first thought was that components therefore must be something different from ui. 

This comment limits components to be visible things used by tabs. As a developer I can now look at the screen and have a good idea of what a gitui::component is.

asyncgit/src/lib.rs
peso: Ok. Add comments on sync part.

src/main.rs
```diff
@@ -1,3 +1,36 @@
...
+//!
+//! - User Interface
+//!   - [tabs] for main navigation
+//!   - [components] for visual elements used on tabs
+//!   - [popups] for temporary dialogs
+//!   - [ui] for tooling like scrollbars
+//! - Git Interface
+//!   - [asyncgit] (crate) for async operations on repository
+//! - Distribution and Documentation
+//!   - Project files: README.md, LICENSE.md, CHANGELOG.md, CONTRIBUTING.md, etc
```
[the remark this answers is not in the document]
peso: I have also unlisted the usage documents. However, I feel that these file ought to be mentioned somewhere in the documentation as they are important for users. Maybe some later PR.

src/components/mod.rs
cruessler: I understand both sides of the argument. It certainly makes sense to add context on what a “component” is in the context of `gitui`, in particular if the location in the source tree doesn’t already give you that information. On the other hand, I also find the comment in its current form to be too terse and potentially explaining too little. Is there maybe a way to add more context?

extrawurst: yeah lets add at least one example of how we are composing these components to make up a tab (just examlatory, not exhaustive)
something like "TAB `X` consists of `Y` and `Z` and `Z` consists of for example `Foo`"

extrawurst: and that composition works manually via forwarding the right functions like draw, commands and event

peso: Good suggestions, I think this will improve the value of the description. I will make a more elaborate version with a small example.

peso:
github says there is one more requested change, but I think I did address them all.
Did I miss one?

cruessler (approved):
I think the docs are really helpful! I think this can be merged. In case anyone wants further changes I suggest iterating after merging.

PR #2643, opened 2025-05-17
Use `gitoxide` in `get_commits_info` and `get_commit_info`

cruessler:
- Implement `From<gix::ObjectId>` for `CommitId`
- Use `gitoxide` in `get_commits_info` and `get_commit_info`

This PR changes `get_commits_info` and `get_commit_info` to use `gitoxide` under the hood. It does not explicitly change any behaviour, although there possibly are subtle differences that I’m not aware of.

This implementation doesn’t log an error when either `author` or `committer` cannot be resolved using `mailmap` as the underlying implementation returns `Option<Signature>` instead of `Result<Signature>`. If you want, I can change that, though.

I decided to duplicate the logic of `get_message` into `gix_get_message`, using a prefix for the time both implementations are required.

I did not attempt to re-organize any of the code. My plan is to convert more functions to `gitoxide` before making any changes in that direction.

I followed the checklist:

### DIFF
--- a/asyncgit/src/sync/commits_info.rs
+++ b/asyncgit/src/sync/commits_info.rs
@@ -3,12 +3,9 @@ use std::fmt::Display;
 use super::RepoPath;
 use crate::{
 	error::Result,
-	sync::{
-		commit_details::get_author_of_commit,
-		repository::{gix_repo, repo},
-	},
+	sync::repository::{gix_repo, repo},
 };
-use git2::{Commit, Error, Oid};
+use git2::Oid;
 use scopetime::scope_time;
 use unicode_truncate::UnicodeTruncateStr;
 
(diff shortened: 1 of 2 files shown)


cruessler:
So far, I have not been able to reproduce the test failures reliably on my machine. My current best guess is that they are somehow timing-related, but I don’t yet know why that might be the case as the code in questions seems to be sync. I think it is this call that triggers the error in rare circumstances, but that’s all I got:

https://github.com/gitui-org/gitui/blob/d8eb1f353f03c0cb4b1eccd536e930832de37729/asyncgit/src/sync/blame.rs#L97

cruessler:
I’m currently working on adding rename tracking to `gitoxide` blame: GitoxideLabs/gitoxide#2022. Instead of trying to debug the CI failures we hit in this PR, another option would be to wait for the PR in `gitoxide` to land, then port `gitui`’s blame view to `gitoxide` and then retry porting `get_commits_info` to `gitoxide`.

extrawurst:
@cruessler whatever works best for you

cruessler:
@extrawurst Since rename tracking for blame has landed in `gitoxide` a few days ago, I think I’m leaning towards making the switch for blame first.

extrawurst:
Sounds good to me

extrawurst:
@cruessler should we mark this `draft` while we wait for blame to land?

cruessler:
> @cruessler should we mark this `draft` while we wait for blame to land?

Makes total sense, I’ve converted it to draft! (I hope I’m not spamming too much with all these experimental PRs that explore using `gitoxide` in various parts of `gitui`.)

extrawurst:
waiting for https://github.com/gitui-org/gitui/pull/2591

PR #2646, opened 2025-05-19
Update scc and sdd to non-yanked versions

cruessler:
This makes `cargo deny check` pass on my machine.


extrawurst:
@cruessler awesome! thanks!

PR #2647, opened 2025-05-19
Group Dependabot updates

cruessler:
This PR configures Dependabot to use one grouped PR instead of individual PRs for its daily updates.


extrawurst:
lets give this a try

PR #2664, opened 2025-06-09
Use `gitoxide` for `get_tags`

cruessler:
This PR is a draft, currently intended for a discussion with @Byron about the part of the API to be used for getting a reference’s name. I intend to mark it as ready once I’ve figured out how to best use the existing APIs.

Update 2025-06-09: this PR is now ready for review.

This PR changes the following:

- It converts `get_tags` to use `gitoxide` under the hood.
- It seems the current implementation might have a bug. At least, the `gitoxide` version properly show an `@` symbol in the tags popup for annotated tags, something the `libgit` version doesn’t do. (I didn’t investigate this in depth, though.)

I followed the checklist:

### DIFF
--- a/asyncgit/src/sync/commits_info.rs
+++ b/asyncgit/src/sync/commits_info.rs
@@ -96,6 +96,15 @@ impl From<gix::ObjectId> for CommitId {
 	}
 }
 
+impl From<gix::Commit<'_>> for CommitId {
+	fn from(commit: gix::Commit<'_>) -> Self {
+		#[allow(clippy::expect_used)]
+		let oid = Oid::from_bytes(commit.id().as_bytes()).expect("`Oid::from_bytes(commit.id().as_bytes())` is expected to never fail");
+
+		Self::new(oid)
+	}
+}
+
 impl From<CommitId> for gix::ObjectId {
 	fn from(id: CommitId) -> Self {
 		Self::from_bytes_or_panic(id.0.as_bytes())
--- a/asyncgit/src/sync/tags.rs
+++ b/asyncgit/src/sync/tags.rs
@@ -63,23 +63,20 @@ pub fn get_tags(repo_path: &RepoPath) -> Result<Tags> {
 						.map(Into::into)?;
 	let platform = gix_repo.references()?;
 	for mut reference in (platform.tags()?).flatten() {
-		let commit = reference.peel_to_commit().ok();
-		let tag = reference.peel_to_tag().ok();
+		let commit = reference.peel_to_commit();
+		let tag = reference.peel_to_tag();
 
-		if let Some(commit) = commit {
-			let name = tag
-				.as_ref()
-				.and_then(|tag| {
-					let tag_ref = tag.decode().ok();
-					tag_ref.map(|tag_ref| tag_ref.name.to_string())
-				})
-				.unwrap_or_else(|| {
-					reference.name().shorten().to_string()
-				});
-			let annotation = tag.and_then(|tag| {
-				let tag_ref = tag.decode().ok();
-				tag_ref.map(|tag_ref| tag_ref.message.to_string())
-			});
+		if let Ok(commit) = commit {
+			let tag_ref = tag.as_ref().map(gix::Tag::decode);
+
+			let name = match tag_ref {
+				Ok(Ok(tag)) => tag.name.to_string(),
+				_ => reference.name().shorten().to_string(),
+			};
+			let annotation = match tag_ref {
+				Ok(Ok(tag)) => Some(tag.message.to_string()),
+				_ => None,
+			};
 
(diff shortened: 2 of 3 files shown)


asyncgit/src/sync/tags.rs
```diff
@@ -64,52 +58,31 @@ pub fn get_tags(repo_path: &RepoPath) -> Result<Tags> {
...
-			if let Some(commit) = commit {
-				adder(commit, Tag { name, annotation });
-			}
-
-			return true;
+				.and_then(|tag| {
+					let tag_ref = tag.decode().ok();
+					tag_ref.map(|tag_ref| tag_ref.name.to_string())
+				})
+				.unwrap_or_else(|| reference_name[10..].to_string());
```
Byron: Oh right, this should be `reference.name().shorten()`.

cruessler: That was exactly what I was looking for, thanks!

extrawurst:
@cruessler lgtm. do we need a @Byron review? do you know of anything not working as before?

cruessler:
Based on my observations, the `gitoxide`-based version is more correct in that it properly marks annotated tags. On the left, you can see the `git2`-based version while on the right there is the `gitoxide`-based one. Only the version on the right shows the `@` symbol, indicating annotated tags.

image

I haven’t found any issues yet, but I also haven’t been using the tag popup on a regular basis. If @Byron wants to have a quick last look, that would be welcome, but not strictly necessary. :-)

asyncgit/src/sync/commits_info.rs
```diff
@@ -96,6 +96,15 @@ impl From<gix::ObjectId> for CommitId {
 	}
 }
 
+impl From<gix::Commit<'_>> for CommitId {
+	fn from(commit: gix::Commit<'_>) -> Self {
+		#[allow(clippy::expect_used)]
+		let oid = Oid::from_bytes(commit.id().as_bytes()).expect("`Oid::from_bytes(commit.id().as_bytes())` is expected to never fail");
```
Byron, line 102: (Nit) I often struggle with good `.expect("messages")` as they shouldn't be redundant, or else I could use the `.unwrap()` I want to avoid. Here I'd probably write that `git2 commits are always 20 bytes long` as the reason I expect this to always work.

Funnily enough, while writing it I was wondering… are they? Is there SHA256 and would that work then? Answer: certainly once `gix` has SHA256 compiled in, but it felt interesting how a good `.expect()` makes you think.

asyncgit/src/sync/tags.rs
```diff
@@ -64,52 +58,29 @@ pub fn get_tags(repo_path: &RepoPath) -> Result<Tags> {
...
-						.flatten()
-				});
-
-			if let Some(commit) = commit {
-				adder(commit, Tag { name, annotation });
-			}
-
-			return true;
+	let gix_repo: gix::Repository =
+				gix::ThreadSafeRepository::discover_with_environment_overrides(repo_path.gitpath())
```
Byron, line 62: When seeing this I think `gix` should have an `open_with_environment_overrides`. Even though discover won't actually get to go upwards, that would much more clearly indicate the intent.

A PR is very welcome, it should not be more than a refactor.. The key here is really just

```rust
        if std::env::var_os("GIT_DIR").is_some() {
            return Self::open_with_environment_overrides(directory.as_ref(), trust_map).map_err(Error::Open);
        }
```

which could also be part of the `open_with_environment_overrides()` that we really want here.
Pretty please :D?

asyncgit/src/sync/tags.rs
```diff
@@ -64,52 +58,29 @@ pub fn get_tags(repo_path: &RepoPath) -> Result<Tags> {
...
+						.map(Into::into)?;
+	let platform = gix_repo.references()?;
+	for mut reference in (platform.tags()?).flatten() {
+		let commit = reference.peel_to_commit();
+		let tag = reference.peel_to_tag();
```
Byron, line 67: To be extra-awesome, I'd invert the logic here and `peel_to_tag()` only, and then get the object and peel that again, and only if it's not a tag, peel to a commit.
To streamline this, there is `reference.peel_to_kind(gix::object::Kind::Tag)`, and if that fails, one can peel again to a Commit maybe. This method is mutating, so one won't lose progress - if it's not a tag it's a commit (usually), and it will do no work to realise that.

But if you want nice code, this is probably it, and I don't think this is a performance issue either.
Gosh, I shouldn't spend my time with premature performance optimisation!

Byron (commented):
I love the fly-by correctness fix!

Of course I couldn't resist to take a look, and it looks good to me, too!

extrawurst:
added to changelog in 1e804801224c1306702e9b5223bd681f8e042111

PR #2671, opened 2025-06-25
Update which requirement from 7.0 to 8.0

musicinmybrain:
This Pull Request fixes/closes #{issue_num}.v **N/A**

It changes the following:
- Updates the `which` dependency in `Cargo.toml` from 7.0 to 8.0
- Updates `Cargo.lock` with `cargo update which`.

I followed the checklist:
- [ ] I added unittests **N/A, nothing new to test**
- [x] I ran `make check` without errors. (On Fedora 42, I needed to use `CC=clang CXX=clang++` because `onig_sys` does not build with GCC 15 due to incompatible pointer types in the oniguruma C code. I also needed `python-unversioned-command` installed, and I needed to `cargo install cargo sort`. None of this has anything to do with this PR.)
- [ ] I added an appropriate item to the changelog **N/A, I think. It looks like you do not include dependency updates in the changelog.**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -567,7 +567,7 @@ dependencies = [
  "crossterm_winapi",
  "mio",
  "parking_lot",
- "rustix",
+ "rustix 0.38.43",
  "serde",
  "signal-hook",
  "signal-hook-mio",
@@ -1471,7 +1471,7 @@ dependencies = [
  "itoa",
  "libc",
  "memmap2",
- "rustix",
+ "rustix 0.38.43",
  "smallvec",
  "thiserror 2.0.12",
 ]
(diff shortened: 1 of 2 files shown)


cruessler:
Are there any reasons to update this dependency specifically? Just curious, I’m not questioning that these kinds of updates are generally useful.

musicinmybrain:
> Are there any reasons to update this dependency specifically? Just curious, I’m not questioning that these kinds of updates are generally useful.

Based on https://github.com/harryfei/which-rs/blob/8.0.0/CHANGELOG.md, the immediate benefits are rather minor: better handling of I/O errors, and one fewer dependency.

(I filed this PR while updating the [`rust-which` package](https://src.fedoraproject.org/rpms/rust-which) in Fedora. I patched packages as needed to avoid a `rust-which7` compat package, and it’s our policy to offer patches upstream.)

extrawurst:
> one fewer dependency.

Thats actually the opposite if you check the diff on the lock file. I would defer this one until we have a chance to not duplicate the `rustix` dependency

alerque:
Agreed, this should include (or hold off and be included in) an update that keeps multiple major versions of rustix from hitting the dependency tree.

musicinmybrain:
> Either update rustix (consulting MSRV or other considerations?) or hold off until rustix can be bumped.

This PR is nine months old. The change proposed here was eventually committed, four months after this PR, in 3e98a402065c3135fbc36892131d103d640c432d, after dependabot proposed it in https://github.com/gitui-org/gitui/pull/2723.

PR #2673, opened 2025-06-29
Use `gitoxide` in `get_status`

cruessler:
### Update 2025-07-06

I’ve been using this branch daily for the last week. I found one issue that I’ve addressed in a separate commit. I think it’s now ready for review!

### Initial description

This is a draft PR. I think that, at this point, most of the initial work of porting is done. I’m going to be using this branch as my daily driver throughout the week to find any issues there might still be left. I’m planning on marking it as ready for review at the end of the week.

This PR changes `asyncgit::sync::get_status` to use `gitoxide`. As a side-effect, it lets `gitoxide` handle reading `status.showUntrackedFiles`. The resulting behaviour, with respect to `status.showUntrackedFiles`, is closer to what git does than it was before this PR. I’ve updated 2 tests according to the new behaviour.

### DIFF
--- a/asyncgit/src/error.rs
+++ b/asyncgit/src/error.rs
@@ -124,6 +124,10 @@ pub enum Error {
 	),
 
 	///
+	#[error("gix::pathspec::init::Error error: {0}")]
+	GixPathspecInit(#[from] Box<gix::pathspec::init::Error>),
+
+	///
 	#[error("gix::reference::head_tree_id::Error error: {0}")]
 	GixReferenceHeadTreeId(
 		#[from] gix::reference::head_tree_id::Error,
@@ -188,6 +192,12 @@ impl<T> From<crossbeam_channel::SendError<T>> for Error {
 impl From<gix::discover::Error> for Error {
 	fn from(error: gix::discover::Error) -> Self {
 		Self::GixDiscover(Box::new(error))
+	}
+}
+
+impl From<gix::pathspec::init::Error> for Error {
+	fn from(error: gix::pathspec::init::Error) -> Self {
+		Self::GixPathspecInit(Box::new(error))
 	}
 }
 
--- a/asyncgit/src/error.rs
+++ b/asyncgit/src/error.rs
@@ -124,6 +124,24 @@ pub enum Error {
 	),
 
 	///
+	#[error("gix::status::Error error: {0}")]
+	GixStatus(#[from] Box<gix::status::Error>),
+
+	///
+	#[error("gix::status::iter::Error error: {0}")]
+	GixStatusIter(#[from] Box<gix::status::iter::Error>),
+
+	///
+	#[error("gix::status::into_iter::Error error: {0}")]
+	GixStatusIntoIter(#[from] Box<gix::status::into_iter::Error>),
+
+	///
+	#[error("gix::status::index_worktree::Error error: {0}")]
+	GixStatusIndexWorktree(
+		#[from] Box<gix::status::index_worktree::Error>,
+	),
+
+	///
 	#[error("amend error: config commit.gpgsign=true detected.\ngpg signing is not supported for amending non-last commits")]
 	SignAmendNonLastCommit,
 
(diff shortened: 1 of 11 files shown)


asyncgit/src/sync/status.rs
```diff
@@ -134,59 +177,75 @@ pub fn get_status(
...
+		StatusType::Stage => {
+			let iter = status.into_iter(Vec::new())?;
+
+			for item in iter {
+				let item = item?;
+
+				let path = item.location().to_string();
+
+				match item {
+					gix::status::Item::IndexWorktree(_) => {}
```
cruessler: @Byron Context: this is moving the status view in `gitui` to `gitoxide`, in particular the staging area.

There is [`tree_index_status`](https://docs.rs/gix/latest/gix/struct.Repository.html#method.tree_index_status) which seems like something I’d want to use for this purpose, but the docs explicitly say: “This is a low-level method - prefer the [Repository::status()](https://docs.rs/gix/latest/gix/struct.Repository.html#method.status) platform instead”. Reading the docs for `status`, I didn’t find a way to configure it to give the same information as `tree_index_status`. Am I missing anything?

Byron: I think I see what the problem is.
There is iterators for everything, and for index-worktree changes, and both are used.

But there is no iterator version for tree-index changes, which is what you could be looking for, only this relatively low-level callback version. The reason for that is that this is so fast there isn't a need be smart about it. Ideally, there would be a version which just returns a Vector of these changes, but in absence of it you could just call that and build a Vec yourself in the callback, or extract the needed information right there.

There is no alternative, this one would have to be used.

cruessler: Thanks a lot, really helpful! I changed the implementation to use `tree_index_status`.

I noticed that there is no `ChangeRef::location`, so I’m now getting it via `change_ref.fields().0`. Would it make sense to add an explicit `ChangeRef::location` to `gitoxide`?

Byron: Yes, definitely, particularly for consistency. Ideally it's also advertised in the `fields()` docs.

cruessler: I’ve implemented the suggestion in GitoxideLabs/gitoxide#2071!

asyncgit/src/error.rs
```diff
@@ -123,6 +123,44 @@ pub enum Error {
 		#[from] gix::object::find::existing::with_conversion::Error,
 	),
 
+	///
+	#[error("gix::pathspec::init::Error error: {0}")]
+	GixPathspecInit(#[from] Box<gix::pathspec::init::Error>),
```
extrawurst, line 128: I am wondering if this should be moved into a separate `GixError` Type to keep the Cambrian explosion under control with all these gix errors?

asyncgit/src/sync/config.rs
```diff
@@ -56,6 +56,13 @@ pub fn untracked_files_config_repo(
 		}
 	}
 
+	// This does not reflect how git works according to its docs that say: "If this variable is not
```
extrawurst, line 59: good catch, we should put a reminder into your tracker issue to clean that up eventually

asyncgit/src/sync/status.rs
```diff
@@ -134,59 +177,93 @@ pub fn get_status(
 ) -> Result<Vec<StatusItem>> {
 	scope_time!("get_status");
 
-	let repo = repo(repo_path)?;
+	let repo: gix::Repository =
```
extrawurst: maybe we should make this a helper method as we want to make sure we always use a consistent same way to create this repo using `discover_with_environment_overrides`, right? we have introduced this mouthful already in a few places

cruessler:
> just a few questions and nits, other looks good to go to me

Great suggestions! I’ve added them as items to the tracking issue to unblock this PR. Also, I’ve already implemented one of them. :-)

cruessler:
@extrawurst This PR is ready from my side! Do you want further changes before it is merged?

extrawurst:
@cruessler no sorry for the delay! looks good to me! 🥳

PR #2684, opened 2025-07-31
Improve error message on config dir error

### LINKED ISSUE #2683: Better error message if config (and/or HOME) directory is not writeable
When `gitui` starts and there is no config and the home directory is write protected, it fails with an unspecific message: `Error: Permission denied (os error 13)` while trying to create folders and/or the config itself.

(Original Discussion for more context: #2682)

**To Reproduce**
(Simplest on Linux)
1. Spawn a new shell with a different home directory (just change `HOME` env variable)  
    which is not writable (for example owned by root)
3. run `gitui`
4. Result: `Error: Permission denied (os error 13)`

A clear and concise description of what `gitui` failed to do. So in this case probably something like:
```
Failed to write config to /some/path/.config
Error: Permission denied (os error 13)
```

Not implementing it and let the user figure it out :-D
I personally used `strace` to find the issue because I first thought it had something to do with my terminal setup (see #2682)

sobczal2:
This Pull Request closes #2683.

It changes the error message when the config and/or HOME directory is not writable from:
```
Error: Permission denied (os error 13)
```
to
```
Error: failed to write config to /home/sobczal/Dev/playground/test/.config/gitui

Caused by:
    Permission denied (os error 13)
```

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -50,7 +50,10 @@ pub fn process_cmdline() -> Result<CliArgs> {
 
 	let confpath = get_app_config_path()?;
 	fs::create_dir_all(&confpath).with_context(|| {
-		format!("failed to write config to {}", confpath.display())
+		format!(
+			"failed to create config directory: {}",
+			confpath.display()
+		)
 	})?;
 	let theme = confpath.join(arg_theme);
 
--- a/src/args.rs
+++ b/src/args.rs
@@ -1,5 +1,5 @@
 use crate::bug_report;
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Context, Result};
 use asyncgit::sync::RepoPath;
 use clap::{
 	builder::ArgPredicate, crate_authors, crate_description,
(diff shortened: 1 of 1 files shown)


src/args.rs
```diff
@@ -49,7 +49,9 @@ pub fn process_cmdline() -> Result<CliArgs> {
 		.map_or_else(|| PathBuf::from("theme.ron"), PathBuf::from);
 
 	let confpath = get_app_config_path()?;
-	fs::create_dir_all(&confpath)?;
+	fs::create_dir_all(&confpath).with_context(|| {
+		format!("failed to write config to {}", confpath.display())
```
cruessler: Nit: what do you think about `failed to create config directory: {}`?

cruessler (approved):
If you want, you can address my comment, but you can also keep the existing version, up to you!

src/args.rs
cruessler: (Also, you need to rebase as there were some minor changes that make CI green again.)

sobczal2:
Thanks for the review, I also feel like your suggested message is a little bit more accurate so I've changed it.

Issue #2665, opened 2025-06-10
Having one large, untracked file pegs single CPU core to ~100%


savchenko:
**Describe the bug**
Leaving stray ~700mb file in a repo results in `gitui` taking unreasonable amount of CPU time.

**To Reproduce**
1. Add random (very) large file to the git root.
2. Open `gitui`

**Expected behavior**
The file is ignored unless user explicitly adds it to the repository.

**Screenshots**
N/A

**Context (please complete the following information):**
 - OS/Distro + Version: Debian 13
 - GitUI Version: gitui 0.27.0 2025-01-14 (99f6967)
 - Rust version: rustc 1.86.0 (05f9846f8 2025-03-31)

**Additional context**
Removing the file resolves the issue in its entirety.

extrawurst:
Oh boy that sounds like an upstream issue with libgit2 and their slow status functions - @cruessler when can we migrate the status tab to gitoxide? 😉

cruessler:
Well, from what I can tell, `gitoxide`’s status API already has everything we need (I’ve already experimented with it a bit in a small side-project of mine). It’s even being integrated into `cargo package` as we speak (https://github.com/rust-lang/cargo/pull/15534). To summarize, I’ve just moved it to the front of my oxidize to-do list. 😉

cruessler:
@savchenko I merged #2673 earlier which switches the status tab to use `gitoxide` instead of `libgit2`. Would you be able to test the latest `master` to see whether your issue got fixed?

savchenko:
@cruessler still happens unfortunately.

#### Steps to reproduce

1. `cargo build --release` from 6685f9adb3c887ce9e9b230ff28e82c9ec326d54
2. `dd if=/dev/urandom of=random_big_file bs=1M count=1000`
3. Launch freshly compiled `gitui`

The CPU usage is following more or less the sinusoidal curve:

https://github.com/user-attachments/assets/46dff383-c88f-4757-a349-7b6fe54cf471

The screen recording above is made from this repository with the 1Gb untracked file. I also see the spinner appearing in the top-left corner of the `gitui` UI.

cruessler:
Thanks for the detailed steps!

I was able to reproduce following your instructions (also built from 6685f9a) and then selecting the file in the status view so that `gitui` started generating the diff.

[...]

PR #2685, opened 2025-07-31
Use gitoxide in get_diff

cruessler:
This is a draft PR that is still somewhat rough around the edges. It replaces the `libgit2` logic in `get_diff` with `gitoxide`. A few things don’t work yet, but the code is already capable of generating correct diffs in some cases. I hope to get this PR into a reviewable/mergable state by the end of next week, but I won’t make any promises. :-)

This will hopefully also address #2665.

### Update 2026-01-16

I’ve updated to the latest changes in `gix_diff`’s API. Some keyboard shortcuts don’t yet work as expected in all scenarios which I will have a look at next.

### Update 2025-08-08

I’m now down to just 4 tests not passing! Also, as I’d hoped, this PR addesses #2665, completely eliminating any CPU load because binary files are ignored by the diff algorithm.

There’s a few remaining issues/open questions:

- `gitoxide` does not provide structured diffs yet which is why this PR, as a workaround, resorts to string matching on a given hunk’s lines. This will hopefully be addressed upstream through GitoxideLabs/gitoxide#2102. Once `gitoxide` has dedicated data structures, I should be able to simplify this PR considerably.
- `gitoxide` does not provide extended headers such as `@@ -315,6 +315,12 @@ impl ConsumeHunk for FileDiff {` the same way `git2` does (unless I’m missing something). The `impl ConsumeHunk for FileDiff {` is not part of what we get from `gitoxide`. As far as I can see, this only affects the UI.
- `gitoxide` produces different diffs than `git2` in certain cases which is known as the “slider problem”. I hope to be able to address this upstream as well, but it will take time and it will block this PR if we’re aiming for `git2` compatibility (and I think there’s strong reasons to do that).

### DIFF
--- a/asyncgit/src/sync/diff.rs
+++ b/asyncgit/src/sync/diff.rs
@@ -19,7 +19,7 @@ use git2::{
 use gix::{
 	bstr::ByteSlice,
 	diff::blob::{
-		unified_diff::{ConsumeHunk, ContextSize, NewlineSeparator},
+		unified_diff::{ConsumeHunk, ContextSize, DiffLineKind},
 		UnifiedDiff,
 	},
 	ObjectId,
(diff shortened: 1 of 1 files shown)


extrawurst:
it looks like https://github.com/GitoxideLabs/gitoxide/pull/2102 is merged, so we can move forward on this, right?

cruessler:
> it looks like [GitoxideLabs/gitoxide#2102](https://github.com/GitoxideLabs/gitoxide/pull/2102) is merged, so we can move forward on this, right?

Yes, we can, thanks for the reminder! I’ll have a look once I get to it.

cruessler:
I’ve merged the latest `master` and updated to the new API. This still needs a bit of work as keyboard shortcuts don’t work in all cases yet. Also, I need to address the test failures. :-)

PR #2687, opened 2025-08-05
Extract GixError

cruessler:
This PR groups `gitoxide`-related errors under a new `GixError` variant. I had to add boilerplate to handle the conversion from source errors to `Error`, but the error structure is more clear now and the added code should hopefully not be difficult to understand. What do you think?


extrawurst:
LGTM

PR #2696, opened 2025-08-21
feat: add ai-powered commit message helpers

yasunogithub:
This Pull Request fixes/closes #{issue_num}.

It changes the following:
-
-

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/commit_helpers.ron.example
+++ b/commit_helpers.ron.example
@@ -1,0 +1,68 @@
+// Example configuration for GitUI commit helpers
+// Copy this file to your GitUI config directory as "commit_helpers.ron"
+// 
+// Config directory locations:
+// - Linux: ~/.config/gitui/
+// - macOS: ~/Library/Application Support/gitui/
+// - Windows: %APPDATA%/gitui/
+//
+// Template variables available in commands:
+// - {staged_diff}   - Output of 'git diff --staged --no-color'
+// - {staged_files}  - List of staged files from 'git diff --staged --name-only'
+// - {branch_name}   - Current branch name
+//
+// Helper navigation:
+// - Ctrl+G: Open helper selection (if multiple helpers configured)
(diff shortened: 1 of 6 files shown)


cruessler:
@yasunogithub Do you know how other CLI tools interface with LLMs? Is your approach in this PR inspired by similar approaches in other tools?

yasunogithub:
> @yasunogithub Do you know how other CLI tools interface with LLMs? Is your approach in this PR inspired by similar approaches in other tools?

@cruessler 
Honestly, I don't have deep knowledge of how other CLI tools specifically interface with LLMs. But based on what I researched, there seem
  to be a few common patterns:

  1. Direct API calls - Tools like aicommits make HTTP requests directly to OpenAI/Anthropic
  2. Local model integration - Tools use ollama or similar for local LLMs
  3. Shell command execution - Some tools spawn subprocesses to call LLM CLIs

  Our approach is simple but flexible - we execute any shell command you configure. So you could:
  - Call curl to hit OpenAI API directly
  - Use ollama run llama3.2 for local models
  - Run custom Python/Node scripts
  - Chain multiple commands with pipes

  The template variables like {staged_diff} give the commands git context, and the RON config lets you set up multiple helpers with hotkeys.

  I chose this approach because any CLI command works - users aren't locked into specific LLM providers or APIs. Want to try a new AI
  service? Just update your config file.

cruessler:
Thanks for the context! I was mainly asking in order to avoid reinventing the wheel in case there are already existing patterns out there.

yasunogithub:
> Thanks for the context! I was mainly asking in order to avoid reinventing the wheel in case there are already existing patterns out there.


Thanks, I agree—this implementation is simple but still consistent with the common patterns, so I don't feel like reinventing the wheel.

PR #2709, opened 2025-08-31
Support opening branches popup from all tabs

kpbaks:
This Pull Request makes it so pressing <kbd>b</kbd> opens the Branches popup in all 5 tabs, not only Status and Log tab. This keeps the behavior consistent, and makes the program simpler to use, since you do not have to associate being in a subset of the available tabs, in order to open a, what I would consider, global popup.

The <kbd>b</kbd> is not being used in any of the 3 missing tabs, so the change does not conflict with any existing keybinds.

If accepted, I think it would benefit the codebase to have a concept of global keybinds, that are accessible from all tabs, to deduplicate the repeated code to e.g. add a branches keybind.

I followed the checklist:
- [ ] I added unittests


extrawurst:
@kpbaks great idea! Why is it `draft`?

kpbaks:
Miss-click I think 😅

PR #2727, opened 2025-10-01
Fix all mismatched_lifetime_syntaxes warnings

linkmauve:
Since Rust 1.89.0, a warning is emitted by default when the lifetime of the return type is implicitly elided but matches an input parameter of the function.

rustc tells us how to fix it (but weirdly enough there is no `cargo fix` for it), so let’s do that, it makes the elided lifetime much more evident anyway.

I followed the checklist:
- [ ] I added unittests (not relevant)
- [ ] I added an appropriate item to the changelog (not relevant)

### DIFF
--- a/asyncgit/src/sync/remotes/mod.rs
+++ b/asyncgit/src/sync/remotes/mod.rs
@@ -120,7 +120,7 @@ pub fn get_default_remote(repo_path: &RepoPath) -> Result<String> {
 /// and Err if there was a problem finding the branch
 fn get_current_branch(
 	repo: &Repository,
-) -> Result<Option<git2::Branch>> {
+) -> Result<Option<git2::Branch<'_>>> {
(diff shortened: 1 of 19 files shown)


extrawurst:
Nice, can you enable that warning to be `deny` in the entire workspace then?

linkmauve:
I usually dislike projects using `-Werror` or similar (`-D warnings` in rustc I think), it would make more sense to put that only in your CI rather than forcing every developer to fix the warnings immediately?  Plus you already build without warnings otherwise, it would be an obvious regression if a PR started doing so.

I couldn’t figure how to deny a warning in the entire workspace though, instead of in every crate like you’re currently doing.

linkmauve:
Here we go, I’ve rebased and added the `#![deny(mismatched_lifetime_syntaxes)]` to every crate in this workspace.

PR #2496, opened 2025-01-21
feat: file and status tab support pageup and pagedown

Fatpandac:
This Pull Request fixes #1951.

It changes the following:
- files tab support

I followed the checklist:
- [ ] I added unittests
- [X] I ran `make check` without errors
- [X] I tested the overall application
- [X] I added an appropriate item to the changelog

PR #2730, opened 2025-10-03
feat: message tab supports pageup and pagedown (#2623)

### LINKED ISSUE #2623: PgUp/PgDown in the "Files" view
Some of the commits have long lists of files. Scrolling through them in the `Files` with just `Up`/`Down` is somewhat daunting. Same goes for `Message` view.

One good example of long file lists and messages is when viewing commits left by `etckeeper`.

`PgUp`/`PgDown` keys seems to work well in some other views (e.g. `Commit` view). It would be very helpful to have the same in the `Files` & `Message` view. Honesty, having these keys work in any scrollable view would be great.

Version: 
gitui --version
gitui nightly 2025-02-28 (99f6967)

xlai89:
This Pull Request fixes/closes #2623 .

It changes the following:
- extend `DetailsComponent` with events of PageUp and PageDown

I followed the checklist:
- [ ] I added unittests

Much thanks to #2496 as a clear and good reference.

### DIFF
--- a/src/components/commit_details/details.rs
+++ b/src/components/commit_details/details.rs
@@ -33,7 +33,7 @@ pub struct DetailsComponent {
 	tags: Vec<Tag>,
 	theme: SharedTheme,
 	focused: bool,
-	current_width: Cell<u16>,
+	current_size: Cell<(u16, u16)>,
 	scroll: VerticalScroll,
 	scroll_to_bottom_next_draw: Cell<bool>,
 	key_config: SharedKeyConfig,
@@ -52,7 +52,7 @@ impl DetailsComponent {
 			theme: env.theme.clone(),
 			focused,
 			scroll_to_bottom_next_draw: Cell::new(false),
-			current_width: Cell::new(0),
+			current_size: Cell::new((0, 0)),
 			scroll: VerticalScroll::new(),
 			key_config: env.key_config.clone(),
 		}
@@ -246,7 +246,40 @@ impl DetailsComponent {
 
 	fn move_scroll_top(&self, move_type: ScrollType) -> bool {
 		if self.data.is_some() {
-			self.scroll.move_top(move_type)
+			let current_height = self.current_size.get().1 as usize;
+			match move_type {
+				ScrollType::PageDown => {
+					let new_top =
+						self.scroll.get_top().saturating_add(
+							current_height.saturating_sub(1),
+						);
+					if self.scroll.get_top() == new_top {
+						return false;
+					}
+					self.scroll.move_area_to_visible(
+						current_height,
+						new_top,
+						new_top.saturating_add(current_height),
+					);
+					true
+				}
+				ScrollType::PageUp => {
+					let new_top =
+						self.scroll.get_top().saturating_sub(
+							current_height.saturating_sub(1),
+						);
+					if self.scroll.get_top() == new_top {
+						return false;
+					}
+					self.scroll.move_area_to_visible(
+						current_height,
+						new_top,
+						new_top.saturating_add(current_height),
+					);
+					true
+				}
+				_ => self.scroll.move_top(move_type),
+			}
 		} else {
 			false
 		}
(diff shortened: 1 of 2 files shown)


xlai89:
The approach in this PR is different than in #2496 . Not sure if this is ok or should the method `move_top` of `VerticalScroll` be modified. I'm new in rust dev, new in this project, would appreciate any help and feedback :)

cruessler:
Sorry it took so long for this feedback! I just had a look at your PR, and one thing I noticed was that the logic related to changing `scroll_top` is now in two places. First, there’s `DetailsComponent::move_scroll_top` and then there’s `VerticalScroll::move_top`. Did you consider adding the new logic, the one related to `PageUp` and `PageDown`, to `VerticalScroll::move_top`? To me, that would seem like the most logical place for it to live. Or did you try and encounter any difficulties?

xlai89:
Hi @cruessler ! Thanks for the feedback. I modified `VerticalScroll::move_top` as you suggested. Because `PageUp` and `PageDown` need a page height, which is not necessary for other `ScrollType`s, I hope the approach with `Option` is okay.

cruessler:
@xlai89 Thanks for the update! There’s another suggestion I’d like to make. `VerticalScroll::update` already gets `visual_height` as a parameter. It also stores `top` and `max_top` on `self`. What do you think about following that pattern and storing `visual_height`/`page_height` as well? That way, you could remove the parameter `page_height` that you had to add to `move_top` and the component would stay consistent.

https://github.com/gitui-org/gitui/blob/60912c0b474f36781810ba9c2058ebbb3c8c401a/src/components/utils/scroll_vertical.rs#L82

xlai89:
@cruessler I stored `visual_height` into `VerticalScroll`, which made `VerticalScroll::move_top` much neater. But I did not modify other methods e.g. `VerticalScroll::update`, which still takes in a `visual_height` parameter, because that could cause many changes in other components and honestly, I am a bit confused with the many ways of scrolling (`move_top`, `update`,`update_no_selection`; `VerticalScroll` vs `Selection`). But please let me know if any further modification is needed.

cruessler:
We’re almost there, I think! There’s a few minor suggestions I’d still like to make, though. Can you move this line `self.visual_height.set(visual_height)` into `VerticalScroll::update` and remove `VerticalScroll::set_visual_height` and `VerticalScroll::get_visual_height`? Apart from that, you’re right that this section of the code is probably not the easiest to follow. :-)

xlai89:
Oh sure, makes total sense :)

cruessler:
Perfect! I’ve now started testing your changes on my machine, and I observed that I can’t seem to fully scroll to the top when using `PageUp`. Instead, there’s always one line missing that I can only scroll by using `Up`. Scrolling to the last line of the message using `PageDown` works, though.

xlai89:
You're right. I should have noticed that myself, sorry. After I understand `saturated_sub` correctly, it should be fixed now :)

src/components/utils/scroll_vertical.rs
```diff
@@ -38,9 +38,14 @@ impl VerticalScroll {
...
-			ScrollType::PageUp => old
-				.saturating_sub(self.visual_height.get())
-				.saturating_add(1),
+			ScrollType::PageUp => {
+				if old < self.visual_height.get() {
+					0
+				} else {
+					old.saturating_sub(self.visual_height.get())
+						.saturating_add(1)
+				}
```
cruessler: Would this also work?

```suggestion
				old.saturating_add(1)
					.saturating_sub(self.visual_height.get())
```

xlai89: Absolutely! Thank you very much for your patience and great hints!

src/components/commit_details/details.rs
```diff
@@ -33,7 +33,7 @@ pub struct DetailsComponent {
 	tags: Vec<Tag>,
 	theme: SharedTheme,
 	focused: bool,
-	current_width: Cell<u16>,
+	current_size: Cell<(u16, u16)>,
```
cruessler: Given the recent changes, it looks like we don’t need to change `current_width` to `current_size` anymore, but can keep it as `current_width`. What do you think?

xlai89: Good point! Changed that back and did a rebase as well.

cruessler (approved):
LGTM! Thanks for your patience!

PR #2734, opened 2025-10-06
fix the crashes when starting it from .git directory issue

### LINKED ISSUE #2732: Gitui crashes at start
**Describe the bug**
App crashes when starting it (from .git directory).

**To Reproduce**
1. Go to a repo and CD into the .git directory
2. Start the app

**Expected behavior**
Application to start

**Screenshots**

 - OS/Distro + Version: `macOS 14.6.1`
 - GitUI Version: `0.27`

**Additional context**
I get this stacktrace on the command prompt:

```
GitUI was close due to an unexpected panic.
   2: <rayon_core::job::HeapJob<BODY> as rayon_core::job::Job>::executethe following info:
   3: rayon_core::registry::WorkerThread::wait_until_cold
   4: std::sys::backtrace::__rust_begin_short_backtrace { file: "asyncgit/src/revlog.rs", line: 186, col: 14 }, can_unwind: true, force_no_backtrace: false }
   5: core::ops::function::FnOnce::call_once{{vtable.shim}}
   6: std::sys::pal::unix::thread::Thread::new::thread_start
   7: __pthread_startc_handlers::{{closure}}
│  2: std::panicking::rust_panic_with_hook
[1]    45588 abort      gituianic_handler::{{closure}}
```

0x61nas:
This Pull Request fixes/closes #2732.

It changes the following:

- Checks if the current working directory is `git`
- and if its we take its parent directory

I followed the checklist:

- [ ] ~I added unittests~ (no need)

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -34,9 +34,24 @@ pub fn process_cmdline() -> Result<CliArgs> {
 
 	let workdir =
 		arg_matches.get_one::<String>("workdir").map(PathBuf::from);
-	let gitdir = arg_matches
-		.get_one::<String>("directory")
-		.map_or_else(|| PathBuf::from("."), PathBuf::from);
+	let gitdir = 'blk: {
+		let mut cwd = env::current_dir()?;
+		let cwd_file_name = cwd
+			.file_name()
+			.ok_or(anyhow!("Coulden't get the CWD name"))?
+			.as_encoded_bytes();
+		if cwd_file_name == b".git" {
+			cwd = cwd
+				.parent()
+				.ok_or(anyhow!(
+					"Coulden't find the parent directory of .git"
+				))?
+				.to_path_buf();
+		}
+		break 'blk arg_matches
+			.get_one::<String>("directory")
+			.map_or_else(|| cwd, PathBuf::from);
+	};
 
 	let repo_path = if let Some(w) = workdir {
 		RepoPath::Workdir { gitdir, workdir: w }


cruessler:
@extrawurst What’s the expected outcome? Currently, I’m getting the following error message:

```
gitoxide/.git on  main [!?⇣]
❯ gitui

GitUI was closed due to an unexpected panic.
Please file an issue on https://github.com/gitui-org/gitui/issues with the following info:

panicked at asyncgit/src/revlog.rs:186:14:
failed to fetch: Gix(Discover(Open(Io(Custom { kind: Other, error: "'./.git/objects' wasn't a directory" }))))

trace:
   0: gitui::set_panic_handler::{{closure}}
   1: std::panicking::rust_panic_with_hook
   2: std::panicking::begin_panic_handler::{{closure}}
   3: std::sys::backtrace::__rust_end_short_backtrace
   4: __rustc::rust_begin_unwind
   5: core::panicking::panic_fmt
   6: core::result::unwrap_failed
   7: core::result::Result<T,E>::expect
   8: <rayon_core::job::HeapJob<BODY> as rayon_core::job::Job>::execute
   9: rayon_core::registry::WorkerThread::wait_until_cold
  10: std::sys::backtrace::__rust_begin_short_backtrace
  11: core::ops::function::FnOnce::call_once{{vtable.shim}}
  12: std::sys::pal::unix::thread::Thread::new::thread_start
  13: start_thread
             at ./nptl/pthread_create.c:448:8
  14: __GI___clone3
             at ./misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S:78:0

Rayon: detected unexpected panic; aborting
fish: Job 1, 'gitui' terminated by signal SIGABRT (Abort)
```

In general, I don’t think we should try to replicate the logic for detecting whether we’re in a valid git repository as that is already done by `libgit2` or `gitoxide`. Instead, we should focus on surfacing the error returned by `libgit2`/`gitoxide` in a more concise way. What do you think? If you agree, we should probably change this https://github.com/gitui-org/gitui/blob/1d2248571dbec63563ac360de46d6920209d96cd/asyncgit/src/revlog.rs#L186 in a way that the caller gets access to the error and can react accordingly.

extrawurst:
I agree

0x61nas:
hmm, that's weird it worked in my testing, i'll try to look again

0x61nas:
> In general, I don’t think we should try to replicate the logic for detecting whether we’re in a valid git repository as that is already done by libgit2 or gitoxide. Instead, we should focus on surfacing the error returned by libgit2/gitoxide in a more concise way. What do you think? If you agree, we should probably change this 

I thought of this first, but being in the `.git` directory is sort of a special case for us and am not sure if libgit2/gitoxide has a specific error for this we can catch it and try to recover

cruessler:
For the special case of opening `gitui` inside `.git`, your workaround works. What I was referring to was that `gitoxide` already gives you a good error message: `Gix(Discover(Open(Io(Custom { kind: Other, error: "'./.git/objects' wasn't a directory" }))))`. What I’m proposing is that we go for a more general solution that would allow us to surface all errors that would currently cause the `.expect()` I was linking to to panic. Instead of panicking in a worker, we would store the error inside `AsyncLog` and force consumers to write code to handle the error case. Does that explanation help?

0x61nas:
yeah, i do agree with you, but isn't that would be considered a major refactoring? and yes it would improve the error handing system in general but i still can't see a reasonable way to recover from this specific error, you might argue that the error message is kinda enough and we shouldn't add an additional complexity to handle this somewhat _rare_ case, but this PR is meant to address this particular issue (#2732) with no breaking changes

although that i do agree that if we just don't panic and display a nice little error message without scaring our users or do something sneaky like in the PR, i don't see it feasible  in this scope (Issue/PR)

so maybe we could consider this as a temporary solution and open another issue for the more correct solution that would require some more changes (but then if we do remove this _hack_ it might considered as a breaking change)

cruessler:
Yes, that would be a larger change, but I’d say go for it! 😄 In this particular instance, I think we shouldn’t feel constrained by the initial ticket’s scope, rather we should go for the solution that makes the most sense long-term. Or do you have other concerns that I haven’t addressed yet?

extrawurst:
I agree with @cruessler here. Will make this PR draft in the meantime

extrawurst:
closed in favor of #2768

PR #2745, opened 2025-10-23
Update gix from 0.71.0 to 0.74.1

cruessler:
This PR updates `gix` to its latest version, as of 2025-10-23. This update removes 2 or 3 transitive dependencies, but it duplicates a couple more. Since this is something I know we’re concerned about, what’s the best strategy to handle this? Another option I haven’t tried yet is upgrading to an older version, hoping it introduces less duplicate dependencies.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -51,6 +51,15 @@ skip-tree = [
     { name = "hashbrown" },
     # 2022-10-26 `getrandom` and `rustix` were added when `gitoxide` was
     # upgraded from 0.71.0 to 0.74.1.
+    # currently needed due to:
+    # * `tempfile v3.23.0`
+    # * `rand_core v0.6.4`
+    # * `redox_users v0.5.0`
     { name = "getrandom" },
+    # currently needed due to:
+    # * `crossterm v0.28.1`
+    # * `which v7.0.2`
+    # * `gix-index v0.42.1`
+    # * `tempfile v3.23.0`
     { name = "rustix" },
 ]


extrawurst:
moving forward gix is a higher priority. if the binary size does not explode with whatever dupes it introduces I am fine. but the CI needs to get green first :)

cruessler:
Perfect, then I’ll go ahead! :-)

We recently moved to MSRV 1.82 in `gitoxide` which is why all the tests that run on 1.81 fail. We could update to a version that supports 1.81 or update our own MSRV. What do you think?

Linting fails because it includes `cargo deny` which I would fix once we’ve decided what to do with respect to the MRSV.

extrawurst:
lets raise the MSRV to `1.82` then

cruessler:
> lets raise the MSRV to `1.82` then

I’ll update the PR tomorrow!

deny.toml
```diff
@@ -49,4 +49,8 @@ skip-tree = [
     # currently needed due to:
     # * `gix-hashtable v0.6.0`
     { name = "hashbrown" },
+    # 2022-10-26 `getrandom` and `rustix` were added when `gitoxide` was
```
extrawurst, line 52: would be good to point out the dependency making this needed, like I tried above

extrawurst (commented):
lgtm just one question

deny.toml
cruessler: Thanks for the hint, I’ve added the missing information.

PR #2746, opened 2025-10-23
add cli flag to open files tab with selected file

### LINKED ISSUE #2510: Flag to automatically open at file
I like to use gitui integrated on my neovim environment, so it makes easier to manage stuff. Even tough neovim has others plugin to manage file chnge history, gitui is simply better. However, navigating until the file to check for the history every time gitui is opened is kinda annoying, if there was an automatic way to invoke it everything would be a lot better.

Add a flag "--file [path]" and gitui will be automatically invoked on the File Tab focusing the filepath.

laktak:
This Pull Request fixes/closes #2510.

It changes the following:
- adds a `--file` cli flag to open the files tab with the given file already selected

I followed the checklist:
- [ ] I added unittests
  - I can't run it locally because of https://github.com/rust-onig/rust-onig/issues/203

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -165,7 +165,7 @@ macro_rules! log_eprintln {
 fn main() -> Result<()> {
 	let app_start = Instant::now();
 
-	let mut cliargs = process_cmdline()?;
+	let cliargs = process_cmdline()?;
 
 	asyncgit::register_tracing_logging();
 	ensure_valid_path(&cliargs.repo_path)?;
@@ -192,10 +192,12 @@ fn main() -> Result<()> {
 		Updater::Ticker
 	};
 
+	let mut args = cliargs;
+
 	loop {
 		let quit_state = run_app(
 			app_start,
-			cliargs.clone(),
+			args.clone(),
 			theme.clone(),
 			key_config.clone(),
 			&input,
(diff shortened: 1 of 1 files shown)


laktak:
I don't do Rust very often so I'm quite happy that it works ;)

However while I followed the guidelines as best as I could, I hope you can help with this error from clippy:

```
error: this function has too many arguments (8/7)
   --> src/main.rs:214:1
```

laktak:
Is that OK?

```
#[allow(clippy::too_many_arguments)]
fn run_app(
```

src/app.rs
```diff
@@ -155,6 +155,7 @@ impl App {
 		sender_app: Sender<AsyncAppNotification>,
 		input: Input,
 		theme: Theme,
+		select_file: Option<PathBuf>,
```
extrawurst: lets just pass the whole `CliArgs` in here as one param combining `selected_file` and `repo` - this should also solve the too many args clippy warning you got

laktak: I think I managed to implement it the way you wanted. I'm not a Rust expert so please check - thanks :)

src/app.rs
```diff
@@ -771,6 +787,9 @@ impl App {
 			InternalEvent::SelectBranch => {
 				self.select_branch_popup.open()?;
 			}
+			InternalEvent::SelectFile { path } => {
+				self.files_tab.find_file(&path);
```
extrawurst: this could fail in case of a race condition where the files for the repo are not loaded yet

laktak: OK, I've move the find_file call to be executed after everything is loaded to `components/revision_files.rs`

laktak:
@extrawurst guess I forgot to request a re-review - let me know if you need other changes

PR #2747, opened 2025-10-25
Preload commit message when undoing last commit

sashomasho:
## Summary
- Retrieves the commit message from HEAD before performing undo
- Preloads the message into the commit popup after undoing
- Improves UX by allowing easy re-committing with the same or modified message

## Test plan
- [ ] Perform a commit
- [ ] Undo the last commit
- [ ] Verify the commit message appears in the commit popup
- [ ] Test that the message can be edited and re-committed
- [ ] Test that closing the popup with Escape preserves the message

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -991,11 +991,30 @@ impl App {
 				self.status_tab.abort_rebase();
 			}
 			Action::UndoCommit => {
+				// Get commit message before undoing
+				let commit_msg = sync::get_head(&self.repo.borrow())
+					.ok()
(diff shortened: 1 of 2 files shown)


extrawurst:
not 100% convinced we need that. the last commit msg is still in the `Ctrl+N` commit message buffer.

sashomasho:
I was unaware of the Ctrl+N message buffer feature. Since that fulfills my needs, this is no longer as critical as I initially thought. We can go ahead and close this.

Issue #2748, opened 2025-10-26
After staging, next unstaged change is janky


Tillerino:
The logic that chooses the next unstaged change after staging a change is kind of janky and breaks my flow. I'll call this a bug, call it what you want. I _believe_ the logic is basically nothing. It just tries to keep the vertical _visual_ position in the tree of unstaged changes. Since that is a _tree_ though, it feels weird. Everything is fine if you have a flat list of files, but I usually don't.

Here's the set up:
```
mkdir -p /tmp/treebug
cd /tmp/treebug
git init

echo hunk > f
mkdir g
  echo hunk > g/g
  echo hunk > g/h
echo hunk > i
echo hunk > j
```
Then enter `gitui`...

<img height="300" alt="Image" src="https://github.com/user-attachments/assets/6ac3b5ad-211e-472d-a157-359a714eb81c" />

## Moving into directories

Ok, let's stage `f`.

<img height="300" alt="Image" src="https://github.com/user-attachments/assets/1ec92fa8-45b3-4e42-b8dd-f678d8da744e" />

Hm, I did not move to the next file. Instead, I moved to the directory g. So just pressing enter repeatedly will not stage file by file, but sometimes stage directories. I don't like this, but whatever. The big issue arises when I am staging from the diff view. This is my initial diff view:

<img height="300" alt="Image" src="https://github.com/user-attachments/assets/b4518d7b-4721-44d3-8650-62e5e3da5b77" />

Now, when I stage this hunk, my next screen is the following:

<img height="300" alt="Image" src="https://github.com/user-attachments/assets/3cef7add-901e-4e96-902b-0c83e5d03351" />

I am seeing the diff of nothing, and I can't do anything in this view. I am forced to return to the tree and navigate to the next file. So staging hunk-by-hunk by repeatedly pressing enter does not work either.

## Moving out of directories

Okay, `f` has been staged. We navigated to the next file manually:

<img height="300" alt="Image" src="https://github.com/user-attachments/assets/60b158f1-8800-45f0-8033-14becbd89729" />

Let's stage `g`:

<img height="300" alt="Image" src="https://github.com/user-attachments/assets/a19e876b-ca3f-4412-baaf-f47b05d0a996" />

Good, let's stage `h`:

<img height="300" alt="Image" src="https://github.com/user-attachments/assets/f9e86ac8-6e67-4d47-a032-1d0335fe4b46" />

[...]

PR #2767, opened 2025-10-29
Fix skipping a file when staging last file in a directory

Tillerino:
This Pull Request partly addresses #2748.

The "error" result from the binary search was not taken into account after an update to the status tree, which lead to files being skipped when the last file in a directory was staged.

**Before:**

[Screencast_20251029_204232.webm](https://github.com/user-attachments/assets/60755369-ef65-4592-ac0b-aedbcb971c52)

(skips `zzz`)

**After:**

[Screencast_20251029_204327.webm](https://github.com/user-attachments/assets/6140037b-7ddd-419b-b60c-dbd6f4865c00)

I followed the checklist:
- [X] I added unittests

### DIFF
--- a/src/components/utils/statustree.rs
+++ b/src/components/utils/statustree.rs
@@ -169,7 +169,7 @@ impl StatusTree {
 	}
 
 	///
-	pub fn is_empty(&self) -> bool {
+	pub const fn is_empty(&self) -> bool {
 		self.tree.items().is_empty()
 	}
 
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,2 +1,2 @@
-msrv = "1.82.0"
+msrv = "1.88.0"
 cognitive-complexity-threshold = 18
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -2,13 +2,13 @@ name: CD
 
 on:
   push:
-   tags:
-     - '*'
+    tags:
+      - "*"
   workflow_dispatch:
-    
+
 permissions:
   contents: write
-  
+
 jobs:
   release:
     strategy:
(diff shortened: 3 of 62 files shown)


src/components/utils/statustree.rs
```diff
@@ -520,7 +520,7 @@ mod tests {
 		res.update(&string_vec_to_status(&["a", "b"])).unwrap();
 		res.selection = Some(1);
 
-		res.update(&string_vec_to_status(&["d", "c", "a"])).unwrap();
+		res.update(&string_vec_to_status(&["a", "c", "d"])).unwrap();
```
Tillerino, line 518: I had to sort the list to keep this test working. The binary search assumes that the status list is sorted, but this is never validated in the status tree. I think this test used to pass accidentally, since the list is so short.

src/components/utils/statustree.rs
```diff
@@ -545,6 +545,21 @@ mod tests {
...
+		res.update(&string_vec_to_status(&["a/b", "c", "d"]))
+			.unwrap();
+		res.selection = Some(1);
+		assert_eq!(
+			res.selected_item().unwrap().info.full_path,
+			"a/b"
+		);
+
+		res.update(&string_vec_to_status(&["c", "d"])).unwrap();
+		assert_eq!(res.selected_item().unwrap().info.full_path, "c");
```
Tillerino: Without the fix, the selection skips c and jumps straight to d.

extrawurst:
@Tillerino could you fix the conflicts?

src/components/utils/statustree.rs
```diff
@@ -202,13 +202,13 @@ impl StatusTree {
...
 
-		if let Ok(i) = self.tree.items().binary_search_by(|e| {
+		let res = self.tree.items().binary_search_by(|e| {
 			e.info.full_path.as_str().cmp(last_selection)
-		}) {
-			return Some(i);
+		});
+		match res {
+			Ok(i) => Some(i),
+			Err(i) => Some(cmp::min(i, self.tree.len() - 1)),
```
cruessler, line 205: Will this ever pick `self.tree.len() - 1` or could we just return `i`? Can we add a test case specifically for the former scenario if this can happen? (I can’t tell whether this is covered by existing tests already.)

Also, `last_index` is now an unused parameter. Can it be removed?

src/components/utils/statustree.rs
```diff
@@ -545,6 +545,21 @@ mod tests {
 		assert_eq!(res.selection, Some(0));
 	}
 
+	#[test]
+	fn test_next_when_dir_disappears() {
+		let mut res = StatusTree::default();
```
cruessler: What do you think about renaming `res` to `tree` for clarity? (`res` probably comes from existing tests, but `tree` carries more context, so I think it is better here).

extrawurst:
@Tillerino would love to see this land, are you committed to push it over the finish line?

src/components/utils/statustree.rs
Tillerino: It will. If the last item from the tree is removed, the search will return the index at which that item would be inserted, which is exactly `self.tree.len()`. This was already covered in a test, but probably rather coincidentally. I added a dedicated test for removing the last item.

Removed `last_index`

src/components/utils/statustree.rs
Tillerino: agreed

Tillerino:
I rebased and incorporated the requested changes. Ready to go.

PR #2772, opened 2025-11-01
Give more space to right-side diff view

hongquan:
In the "inspect commit" view, the left pane (for commit message and list of changed files) and the right pane (for viewing diff) each takes a half of the window. It is not reasonable because the right pane has more content to show

image

This PR is to give more space to the right pane.

image

image

image

It changes the following:
- Change the layout constraints to reduce the width of the left pane.
- Provide different constraints for different window size, so that the left pane width is reasonable.

I followed the checklist:
- [ ] I added unittests


src/popups/inspect_commit.rs
```diff
@@ -67,29 +67,29 @@ pub struct InspectCommitPopup {
...
-
-			f.render_widget(Clear, rect);
-
-			self.details.draw(f, chunks[0])?;
-			self.diff.draw(f, chunks[1])?;
+		if !self.is_visible() {
+			return Ok(());
 		}
+		let constraints = match rect.width {
+			..80 => [
```
peso: I suspect this gives a jump when you go from 79 to 80 width.  At width=79 left panel gets 39 chars, but at width=80 left panel gets 32.
Maybe you want the first range to be ..64 instead of ..80?

hongquan: Thank you. Forgot to check.

PR #2789, opened 2025-11-25
Take workdir into account in `gix_repo`

### LINKED ISSUE #2769: bare repos + workdir with gix
Places that are already migrated to gix use `gix_repo` but in contrast to `repo` this one does not pass `workdir` which means the option of opening a repo as a tuple of `.git` path separate from `workdir` will not work for this.

see https://libgit2.org/docs/reference/main/repository/git_repository_set_workdir.html

we introduced support for this in #1026

cruessler:
In the recently released 0.75.0, `gitoxide` got a new method, `Repository::set_workdir`, that we can use to replicate the same functionality in `libgit2`.

@extrawurst Can yet test whether this fixes #2769 or let me know how to test that myself?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1221,9 +1221,9 @@ dependencies = [
 
 [[package]]
 name = "gix"
-version = "0.74.1"
+version = "0.75.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5fd3a6fea165debe0e80648495f894aa2371a771e3ceb7a7dcc304f1c4344c43"
+checksum = "60beff35667fb0ac935c4c45941868d9cf5025e4b85c58deb3c5a65113e22ce4"
 dependencies = [
  "gix-actor",
  "gix-attributes",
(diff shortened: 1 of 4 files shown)


extrawurst:
Could you build a test case on main that makes use of this where people for example manage their forgives that need to sit in a certain place on disk by a bare repo somewhere else but point it to the dot files parent folder (~/) which then allows you to git manage your dot files/folders - once that works we can use the same test to prove the gix impl works

cruessler:
I’ve created a test case that uses `get_status` to verify that passing a `workdir` works as expected. The same test also passes without changes on `gitui` 0.27.0 (where `get_status` still is backed by `libgit2`), so I believe that `gitoxide` matches `libgit2` as far as I understand the context. Let me know if you think the test should be extended or if there’s anything else you think should be tested!

extrawurst:
@cruessler nice, the only problem is this test also passes on `master` which I suppose it should not. Could you see if the test does something wrong or what the reason for this is. ideally you introduce the test proving we actually fix it first on `master` in a separate PR and annotated with `#[should_panic]` and remove the annotation in this PR

cruessler:
@extrawurst I updated this PR after #2790 got merged.

PR #2790, opened 2025-11-26
Add test that fails because gix_repo does not take workdir into account yet

cruessler:
This is a separate PR adding the test case that is also part of #2789, hopefully showing that it fails on `master`.

I tried marking it with `#[should_panic]`, but then I would have had to refactor the test as the following error message would have forced me to change the return type to `()` which would have required me to turn all `?` into `.unwrap()`, so I decided to not add the annotation.

```
error: functions using `#[should_panic]` must return `()`
```

### DIFF
--- a/asyncgit/src/sync/status.rs
+++ b/asyncgit/src/sync/status.rs
@@ -280,3 +280,50 @@ pub fn get_status(
 
 	Ok(res)
 }
+
+#[cfg(test)]
+mod tests {
+	use std::{fs::File, io::Write, path::Path};
+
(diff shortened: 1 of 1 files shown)


extrawurst:
I am not a huge believer in tests returning `Result` - whats the upside? Let's change the *3* `?` to `unwrap` and we can  `should_panic` :)

cruessler:
Sure! No objections, works for me. :-)

extrawurst:
Awesome!

Issue #2784, opened 2025-11-20
Amending a commit overwrites the author


ItsHarper:
### Describe the bug

Amending a commit overwrites the author

### To Reproduce

1. Checkout a branch where the latest commit was authored by someone else
2. Run `git log --pretty=full -n 1` and verify that the `Author` field shows someone other than you
3. Stage changes
4. Open the commit dialog
5. Press Ctrl-A to make the pending commit be an amend operation
6. Press Ctrl-D to finish the commit
7. Run `git log --pretty=full -n 1` again
8. Observe that you are now listed as the commit author

### Expected behavior

The original author should be preserved

### Context

 - OS: Chimera Linux (updated as of 2025-11-20)
 - GitUI Version: 0.27.0
 - Rust version: 1.91.0

cruessler:
I’ll have a look!

cruessler:
Currently, this seems to be somewhat intentional as the code for amending a commit passes `None` for `committer` which results in the committer being unchanged. I suggest we change that and override the committer with the current user, that is whatever is returned by `signature_allow_undefined_name` (which usually is `git config --get user.name/email`). I’ll prepare a PR soon!

https://github.com/gitui-org/gitui/blob/37d3dd291a6eb91f55d8e8091c376b10ee11d33f/asyncgit/src/sync/commit.rs#L43-L50

extrawurst:
> seems to be somewhat intentional [..] results in the committer being unchanged

isnt that exactly the opposite of what the bug report states?

cruessler:
> > seems to be somewhat intentional [..] results in the committer being unchanged
> 
> isnt that exactly the opposite of what the bug report states?

You’re absolutely right! Then I can’t seem to reproduce as the author is not overwritten if I do an amend using `gitui`. (This observation made me think we should probably change the committer if we amend, but now I’m not so sure anymore.) I can also try and check on a different machine to see whether I’m missing anything.

ItsHarper:
> This observation made me think we should probably change the committer if we amend, but now I’m not so sure anymore.

My understanding is that the committer should change, yeah

Let me double-check I can reproduce myself.

extrawurst:
@ItsHarper 

> My understanding is that the committer should change, yeah

But your bug report says:

> Observe that you are now listed as the commit author

So what is it ? 🙈

ItsHarper:
> [@ItsHarper](https://github.com/ItsHarper)
> 
[...]

PR #2792, opened 2025-11-28
Overwrite committer on amend when gpgsign = false

cruessler:
This fixes one of the issues reported and discussed in #2784.

When `gpgsign` was set to `false`, `amend` previously did not set the committer when creating an amended commit.

This PR does not change the behaviour when `gpgsign` is `true`.


extrawurst:
@cruessler I think thats worth a changelog entry!

cruessler:
Absolutely, I completely forgot, thanks for reminding me! :-)

kritsada1998mt-create:
Main

PR #2813, opened 2025-12-17
Add first snapshot tests using insta

cruessler:
This is a follow-up to #2411, with an important fix. The original PR was reverted in 6d62241 due to a regression in the stashes tab. When the app would open in the stashes tab, it would take a couple of seconds for the app to actually show content. The reason was that I had applied what looked like an optimization to me, but what actually apparently prevented an initial `QueueEvent::Notify` from being sent. This missing event caused the stashes view to only call `self.draw(terminal)?` on the next `QueueEvent::Tick` which by default is after 5 seconds (if you change `TICK_INTERVAL` to a different value, you’ll see that it determines how long the app stays empty). This fix restores the code that was initially taken from `run_app` in `main.rs`.

As far as I can tell, this could also be fixed by adding an initial `self.draw(terminal)?` call before the `loop` in `run_main_loop`, right after `self.app.update()?`.

```rust
// In `Gitui::run_main_loop`.
// This was the version that was part of #2411.
let event = select_event(
	&self.rx_input,
	&self.rx_git,
	&self.rx_app,
	&self.rx_ticker,
	&self.rx_watcher,
	&spinner_ticker,
)?;
// This was the version before #2411.
let event = if first_update {
	first_update = false;
	QueueEvent::Notify
} else {
	select_event(
		&self.rx_input,
		&self.rx_git,
		&self.rx_app,
		&self.rx_ticker,
		&self.rx_watcher,
		&spinner_ticker,
	)?
};
```


extrawurst:
@cruessler could you update this one to recent master. i would like to get this in now during the start of the next major version cycle

extrawurst:
Awesome!

Issue #2820, opened 2025-12-25
0.28.0 shows no files


savchenko:
**Describe the bug**
v0.28.0 fails to show any modified files.

**To Reproduce**
1. Update to 0.28.0 from 0.26.3
2. Go to a local repo which is symlinked from elsewhere
3. Observe empty status

**Expected behavior**
Modified/delete/new files are visible.

**Screenshots**

### v0.26.3
<img width="527" height="337" alt="Image" src="https://github.com/user-attachments/assets/6e98a902-8b45-4e1a-9e83-e896f66fa7a7" />

### v0.28.0
<img width="528" height="412" alt="Image" src="https://github.com/user-attachments/assets/c6afb9f5-7b79-4542-96d5-457ff66f49f0" />

**Context (please complete the following information):**
 - OS/Distro + Version: Debian 13.2
 - GitUI Version 0.28.0
 - Rust version: N/A, downloaded the release binary from GitHub

**Additional context**
I have double-checked that the content is actually not there and not just rendered in the BG colour:

<img width="605" height="690" alt="Image" src="https://github.com/user-attachments/assets/d4c93f6b-cd73-45db-ac86-4b6f87662a66" />

extrawurst:
@savchenko do you have a reproducer for us?

@cruessler I am afraid this might be due to our switch to gitoxide?

cruessler:
@extrawurst Yes, it sounds very likely that this is due to the switch to `gitoxide` in the status tab.

@savchenko Can you elaborate as to what you mean by “Go to a local repo which is symlinked from elsewhere”?

savchenko:
In this particular case:

```bash
$ pwd
/home/user/.config/nvim
$ pwd | xargs realpath
/home/user/.dotfiles/config/nvim
```

0.26.3 and 0.28.0 running from the same `$PWD` side-to-side, 0.28.0 shows current path in the title:

<img width="981" height="213" alt="Image" src="https://github.com/user-attachments/assets/c7424dc5-96f7-4d2b-a5f6-a521e2345601" />

extrawurst:
@savchenko if you want to help us, please share a shell script that sets-up a repo that reproduces the problem

savchenko:
Surprisingly, I can't. This was supposed to do it:

```bash
#!/bin/env bash

echo -e "\nhttps://github.com/gitui-org/gitui/issues/2820\n"

TIMESTAMP=$(date +%H-%M)

[...]

PR #2821, opened 2025-12-29
Don't stop status iter on error, log warning instead

cruessler:
This is a workaround for #2820. The reported issue is that, under very specific circumstances that involve a symlink and a submodule (see the linked issue for details), the status tab is empty. This is due to the fact that the status iter that collects individual status items exits early when it encounters an error. In that instance, all other status items are discarded and `get_status` returns just a single error.

This PR changes the iter to discard errors, but log a warning using `log::warn!` instead. `get_status` will then report all status items that didn’t error. That could potentially be confusing as the UI will not show any indicator that anything has gone wrong, but that seems an acceptable short-term trade-off, compared to showing nothing at all.

### DIFF
--- a/asyncgit/src/sync/status.rs
+++ b/asyncgit/src/sync/status.rs
@@ -202,7 +202,11 @@ pub fn get_status(
 			let iter = status.into_index_worktree_iter(Vec::new())?;
 
 			for item in iter {
-				let item = item?;
+				let Ok(item) = item else {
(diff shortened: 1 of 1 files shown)


extrawurst:
@cruessler if there is no more proper solution for the short-term, we should patch release this, no?

cruessler:
> @cruessler if there is no more proper solution for the short-term, we should patch release this, no?

Yes, I agree!

PR #2822, opened 2025-12-30
feat: added Justfile integration

PokumeKachi:
It changes the following:
- Added a Justfile that is functionally the same as the existing Makefile.

I followed the checklist:
- [x] I added unittests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+      - uses: extractions/setup-just@v4
 
       - name: Restore cargo cache
         uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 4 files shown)


extrawurst:
Thats great! Had that on my list for a while. I dont want to maintain both though, so lets remove the Makefile and make sure all reference of it in CI also switch to just

PokumeKachi:
Anything else to do before merging?

Also please think this through as there may be the possibility that some other collaborators may still want Makefile support.

extrawurst:
there is more GitHub action stuff than just `ci.yaml` did you check the others?

.github/workflows/ci.yml
```diff
@@ -52,14 +52,14 @@ jobs:
           cargo build
 
       - name: Run tests
-        run: make test
+        run: just test
```
extrawurst, line 56: pretty sure `just` is not installed by default in the default GitHub runner image. see https://github.com/casey/just?tab=readme-ov-file#github-actions

PokumeKachi:
Oh alright I will take a look into that tomorrow.

PokumeKachi:
@extrawurst Care to review?

PokumeKachi:
No response from maintainer.

PR #2824, opened 2025-12-31
perf: prevent repeated status fetches in large repos

### LINKED ISSUE #2823: gitui extremely slow on large repositories
**Describe the bug**

gitui can take 5+ minutes to load the status view on very large repositories, while `git status` and `gix status` complete in ~2 seconds.

**To Reproduce**
1. Open gitui in a very large repository
2. Navigate to the Status tab
3. Observe the endless loading spinner

**Expected behavior**
Status should load in a few seconds, similar to `git status` or `gix status`.

**Environment**
- gitui version: 0.27/0.28
- OS: Rocky Linux 8.6
- Rust version: 1.94.0-nightly (also tested on stable 1.92.0)
- Repository size: 1.7M files (uses `git-lfs`)

**Additional context**
Profiling revealed that the git index was being loaded 30+ times during startup.
I believe this is caused by asyncgit's `StatusParams` including a millisecond timestamp (`tick`) in its hash:

```
// asyncgit/src/status.rs
pub struct StatusParams {
    tick: u128, // milliseconds since epoch
    status_type: StatusType,
    config: Option<ShowUntrackedFilesConfig>,
}
```

The tick field was intended to force periodic re-fetches to detect external file changes.

However, because it changes every millisecond, the cache hash changes on every UI tick (every 5 seconds by default), causing a deadly cycle:
1. Cache miss => start new fetch
2. While fetch is pending, more ticks occur
3. Each tick after fetch completes triggers another cache miss 
4. Result: dozens of index loads, each taking ~15 seconds on my target repo

DannyStoll1:
This PR fixes/closes #2823

It changes the following:
- Remove tick from StatusParams hash
- Add generation counter that increments after each fetch completes
- Include generation in cache hash

This ensures:
- No repeated fetches while one is already pending, making gitui usable in very large repos
- New fetch starts immediately after completion, keeping gitui responsive in small repos
- External file changes will still be detected on the next fetch cycle

I followed the checklist:
- [ ] No added unittests: the change is a behavioral fix to existing code, and the existing tests pass
- [x] I tested the overall application: on my 1.7M file target repository, load time goes from 5+ minutes to ~2.5 seconds


extrawurst:
Oh wow interesting, I wonder when we broke that. Great find

cruessler:
Thanks for the PR (and the accompanying issue)! I’ll try to have a look over the weekend!

cruessler:
@DannyStoll1 From what I can tell, these lines seem to serve the same purpose: https://github.com/gitui-org/gitui/blob/7747d829cc21631e15063afb961700a0afa43fe8/asyncgit/src/status.rs#L95-L98

Am I missing something? Do you happen to know why they haven’t prevented `gitui` from becoming slow in your case?

DannyStoll1:
Apologies for the delay getting back to you on this.

`is_pending()` does block ticks that arrive during a fetch, but it doesn't prevent the problem because of the interaction between the millisecond-granularity tick and the 5-second UI tick interval.

For instance, this is the event sequence for a single AsyncStatus instance in a large repo where a fetch takes ~13 seconds:

```
T=0s:   Tick -> fetch(tick=T0)   -> pending=false -> hash=H0 -> MISS -> spawn -> pending=1
T=5s:   Tick -> fetch(tick=T5)   -> pending=true  -> BLOCKED
T=10s:  Tick -> fetch(tick=T10)  -> pending=true  -> BLOCKED
T=13s:  task completes           -> pending=0
T=15s:  Tick -> fetch(tick=T15)  -> pending=false -> hash=H15 (!= H0!) -> MISS -> spawn -> pending=1
T=20s:  Tick -> fetch(tick=T20)  -> pending=true  -> BLOCKED
T=25s:  Tick -> fetch(tick=T25)  -> pending=true  -> BLOCKED
T=28s:  task completes           -> pending=0
T=30s:  Tick -> fetch(tick=T30)  -> MISS -> spawn again...
```

Tick is in milliseconds, meaning the hash is always different on the first tick after completion, guaranteeing a cache miss and a new spawn. So `is_pending()` correctly blocks intermediate ticks, but the cycle never terminates -- every completion is followed by exactly one re-fetch.

This PR tries a different solution: the hash would only change when a fetch completes, so the very next tick after completion gets a miss and starts a new fetch, but ticks that arrive while pending are blocked to prevent queue buildup.

The behavior is the same for small repos -- the difference is that the cache invalidation is tied to completion rather than wall clock time, so there's no risk of the hash silently changing between "pending=0" and the next tick.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixes
+* fix extremely slow status loading in large repositories by replacing time-based cache invalidation with generation counter [[@DannyStoll1](https://github.com/DannyStoll1)]
```
extrawurst: please link to this issue like the other entries also do

cruessler:
> Apologies for the delay getting back to you on this.

No worries, and thanks for both the implementation as well as the detailed explanation!

Issue #2829, opened 2026-01-01
Shadow the window behind the modal when it's open


savchenko:
**Is your feature request related to a problem? Please describe.**
Opened modals blend with the rest of the window.

**Describe the solution you'd like**
Tint the background when a modal is open. Exactly what NeoVim does:

<img width="348" alt="regular window" src="https://github.com/user-attachments/assets/20bfcc96-dea7-4f38-8bc3-475cdf73a39a" />

<img width="348" alt="window with a modal opened" src="https://github.com/user-attachments/assets/0af4e0bc-10b6-4c73-9f3b-36fb2f62f3b5" />

**Describe alternatives you've considered**
Just deal with it.

**Additional context**
Screenshots above.

extrawurst:
That’s a fantastic idea. Not gonna be easy but I like it 👍

callpraths:
I'd like to take a crack at this. @extrawurst please drop any thoughts you have about approaches / gotchas and feel free to assign the issue to me if it helps keep track.

Thanks!

callpraths:
I have a proof of concept implementation up at #2843 

I considered two implementation approaches:

1. Each component owns the responsibility of rendering in "masked over" mode. This would require us to propagate an internal event from/when the popup being shown to the main content so that those components can update their internal state (conceptually add an `is_masked`, similar to `is_visibible`).
2. Manipulate the `Frame.Buffer` directly after the main content is rendered to gray out the already rendered main content.

I chose (2) here for several reasons:
1. Consistency: If each component implements its own "masked" state, it opens up the door for inconsistent color mapping. Components have more flexibility in how they indicate masked state, but this can lead to inconsistent UI treatment (one component may end up using `Color::Gray` vs. `Color::DarkGray` for the same thing in another. Keeping things consistent here would need disciplined use of theme colors).
2. Overhead: It adds extra complexity to each component's implementation, especially given the need for consistency. Most (all?) components should not be opinionated about the styling for masked state anyway, but need to be aware of this concept.
3. Ease of implementation: `gitui` does not have true z-index based layering in the UI, nor does it need it at the moment, in my opinion. The approach here aligns with the simplified approach to z-indices we have - there are only two layers really: main content vs. popups. In this approach, we render the main content, mask it if necessary, and then render the popups.


The open question I have are around what UI treatment we actually want. The OP suggested tinting the background, but based on some experiments, I think it might be better to dull the foreground color instead.

Let's first look at a few examples of setting the background color.

[...]

PR #2843, opened 2026-01-13
[poc] feat: dull tab behind popup

callpraths:
A proof-of-concept implementation for feature #2829 

After this change, the tabs behind popups use grayed out foreground to better focus user's attention onto the popup.

This PR needs at least the following before it can be merged:
- Agreement on the approach (discussion on the tracking issue).
- Extend theme so users can configure the muted foreground color chosen for masking.
- Other items in the checklist below.

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,6 +8,16 @@
             "program": "${workspaceRoot}/target/debug/gitui",
             "args": [],
             "cwd": "${workspaceRoot}",
-        }
+        },
+        {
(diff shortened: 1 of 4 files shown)


.vscode/launch.json
```diff
@@ -8,6 +8,16 @@
             "program": "${workspaceRoot}/target/debug/gitui",
             "args": [],
             "cwd": "${workspaceRoot}",
-        }
+        },
```
callpraths, line 11: split into own PR, if at all.

PR #2864, opened 2026-02-26
Fix Windows AltGr input in text fields

### LINKED ISSUE #2848: Cannot type AltGr-bound characters in the commit field
**Describe the bug**
When trying to commit staged changes, AltGr-bound characters do not work.
On German keyboards, this is used for all braces and brackets `AltGr + 7 = {, 8 = [, 9 = ], 0 = }, ß = \`.

**To Reproduce**
Use german keyboard layout.
Stage changes in the Status UI.
C -> Try to type the special characters using AltGr.

**Expected behavior**
The characters should land in the field.

 - Windows 11 10.0.26200
 - GitUI 0.28.0 2025-12-14 (5527160)

jacob-sheng:
Fixes #2848.

On Windows, AltGr is commonly reported as Ctrl+Alt. Text input previously only inserted chars when neither Ctrl nor Alt were set, so AltGr-generated characters were dropped/(ㄒoㄒ)/

This change treats Ctrl+Alt+Char as regular character input on Windows, while preserving existing Ctrl+Alt shortcuts.

Tests: cargo test

### DIFF
--- a/src/components/textinput.rs
+++ b/src/components/textinput.rs
@@ -616,7 +616,8 @@ impl TextInputComponent {
 				alt: true,
 				..
 			} => {
-				// On Windows, AltGr is commonly reported as Ctrl+Alt.
+				// On Windows, AltGr (Right Alt) can surface to applications as the Ctrl+Alt chord,
+				// so AltGr-generated characters may arrive with both `ctrl` and `alt` set.
 				if cfg!(windows) {
 					ta.insert_char(*c);
 					true


src/components/textinput.rs
```diff
@@ -610,6 +610,20 @@ impl TextInputComponent {
 				ta.scroll(Scrolling::PageUp);
 				true
 			}
+			Input {
+				key: Key::Char(c),
+				ctrl: true,
+				alt: true,
+				..
+			} => {
+				// On Windows, AltGr is commonly reported as Ctrl+Alt.
```
alerque: It's been a hot minute singe I was on a Windows machine, but I have written some keyboard drivers for it in my day and this doesn't ring true at all. Can you provide some citations or explanation here? AltGr is definitely it's own keycode ond most keyboards. Is the reporting a Windows machine *with* an international ISO physical keyboard but *without* a matching layout chosen in the UI?

jacob-sheng: I'm terribly sorry, this PR was written by AI and I hadn't carefully review it, and GPT clearly lacked sufficient world knowledge when handling this part. Thank you for pointing it out. I’ve now changed it to: on Windows, only after existing Ctrl+Alt shortcuts are matched, Ctrl+Alt+Char is treated as text input, so shortcuts like Ctrl+Alt+B will still retain their original functionality.

src/components/textinput.rs
```diff
@@ -610,6 +610,21 @@ impl TextInputComponent {
 				ta.scroll(Scrolling::PageUp);
 				true
 			}
+			Input {
+				key: Key::Char(c),
```
alerque, line 614: What does this key in particular have to do with anything?

src/components/textinput.rs
alerque: This makes even less sense. Please don't mark reviews as resolved until the person asking or the project maintainer marks them as resolved. See my later review comment, but this is sounding even less plausible — and the fact that even the problem explanation is just AI generated slop that even you as the contributor didn't check doesn't make upstream maintainers want to invest in checking it out.

alerque:
Please include information on a test case (Windows version, terminal, keyboard layout/language, and app usage pattern) that would actually be a problem so the problem and solution can be verified by somebody.

jacob-sheng:
Thank you for your review. After my reviewing and consideration I decided to close this PR.

The #2848  `where Windows with a German keyboard cannot input AltGr characters such as {[]}  ` does exist, but it is a known issue upstream in `Crossterm #820, #917, #937`

this PR merely added a workaround at the gitui level by forcibly treating characters with Ctrl|Alt as plain text input. While this may alleviate the issue in theory, it cannot accurately distinguish between AltGr and normal Ctrl+Alt shortcuts, and I didnt thoroughly consider these edge cases. That's on me.

Sorry for taking up your time😭

alerque:
Thanks for reviewing youself.

Next time please do that first. Not just with this project but *any time* you use LLM generated code or prose please review it yourself first, not just that it sounds plausible or looks okay at a glance, but verify that it is actually *correct*. It is very easy for LLMs to spew volumes of code and text that costs project maintainers lots of effort to sort through and determine what makes sense and what is slop or cruft. The onus is on *you* as the human contributor to bear this weight. If projects want just the raw stuff spewed from LLMs they can get it with less effort and a better idea how they got there by prompting an LLM themselves.

jacob-sheng:
I appreciate you taking the time to explain this, and I'll make sure to do a thorough review myself first going forward. Sorry for the extra work this caused, and thank you again for your patience.

PR #2890, opened 2026-03-24
Branching

### LINKED ISSUE #81: Show the branching structure in the log.

philocalyst:
This Pull Request fixes/closes #81.

It changes the following:
- Implements a complete view of commit lineage
- Borrows the logic/implementation from the well-tested [guitar](https://github.com/asinglebit/guitar)
- Conforms to the streaming model that was used in the previous version of the displays, meaning its performance impact is negligible.
- Has a new keybind to hide the graph, as a toggle.
- Expands and contracts dynamically based on what is currently on-screen
- Makes good use of space by compressing whitespace when possible (Through re-using lanes)
- Remains consistent against the rest of the UI.

Screenshot 2026-03-23 at 22 14 25
Screenshot 2026-03-23 at 22 15 36

I followed the checklist:
- [x] I added unittests


philocalyst:
@extrawurst Would love your feedback on this :) Quite large. 

OH! And I just noticed that I removed the missing docs lint, I'll add that back if we agree on this general direction.

gogongxt:
LGTM

CHANGELOG.md
```diff
@@ -7,16 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-### Changed
-* support proper pre-push hook ([#2809](https://github.com/gitui-org/gitui/issues/2809))
-* improve `gitui --version` message [[@hlsxx](https://github.com/hlsxx)] ([#2838](https://github.com/gitui-org/gitui/issues/2838))
-* rust msrv bumped to `1.88`
-
-### Fixed
-* fix extremely slow status loading in large repositories by replacing time-based cache invalidation with generation counter [[@DannyStoll1](https://github.com/DannyStoll1)] ([#2823](https://github.com/gitui-org/gitui/issues/2823))
-* fix panic when renaming or updating remote URL with no remotes configured [[@xvchris](https://github.com/xvchris)] ([#2868](https://github.com/gitui-org/gitui/issues/2868))
+### Added
+* Now with a graph view!
```
extrawurst: please match the ususal format

extrawurst:
I think we need to improve the visualization still, this is the same point in time of the history in three different tools: gitui, tig and fork:

Screenshot 2026-03-29 at 19 50 36
Screenshot 2026-03-29 at 19 50 27
Screenshot 2026-03-29 at 19 51 51

philocalyst:
Being precise here, the only difference I can see in the lack of a bottom connector? Or is there something else?

▰▰▰▰▰
Miles Wirth 🙃

VojtaStanek:
I am looking forward to this change! 

By a quick look the source it seems like it works only with merge commits with 2 parents, but commit can have an arbitrary number of parents (see e.g. [here](https://softwareengineering.stackexchange.com/questions/314215/can-a-git-commit-have-more-than-2-parents)).

philocalyst:
@VojtaStanek Yes! Octopus merges! That's not covered here.

I couldn't find an implementation that handles them in a way that maps well to streaming. They don't crash anything, and happen incredibly rarely. If someone is able to extend what I have here, they're welcome to; but it shouldn't block merging.

cruessler:
Thanks a lot for getting this started, that’s much appreciated! I just started reviewing the changes, but then I had a few more general questions.

In its current form, I find it hard to see merge commits and how they connect to their respective parents. The image below shows a comparison between this branch and `tig` in the `gitoxide` repository that has quite a few merge commits. Is there anything I need to set up to make merge commits more prominent? Does my terminal font need to have certain glyphs?

Also, in `tig`, vertical lines don’t get cut by horizontal ones, as opposed to `guitar`. I think I like `tig`’s approach better. What do you think?

Screenshot_20260407_212543

philocalyst:
No, that's just how they look :sad:.

Good points -- I also wonder if people like the pressed texts to the facets of the branch lines? I could implement that as well.

I think symbols... look cooler; but there are readability improvements we could aim for regardless, that these screenshots are making clear to me. The glyphs could expose be exposed as configuration.

▰▰▰▰▰
Miles Wirth 🙃

extrawurst:
> I also wonder if people like the pressed texts to the facets of the branch lines? I could implement that as well.

Not for now.

> The glyphs could expose be exposed as configuration.

Also not relevant for now. Lets find a visual style thats a good default and add customization in a later PR. THis one is big enough as is.

> In its current form, I find it hard to see merge commits and how they connect to their respective parents

I agree with this. we either find better symbols or go for the letter `M`.

Furthermore I think we should use different colored lines for parallel branches like tig does for more clear separation.

philocalyst:
Screenshot 2026-04-17 at 08 37 30
Screenshot 2026-04-17 at 08 39 29
Screenshot 2026-04-17 at 08 42 16

With the updates, the PR is now looking like the above. Is this a better direction? @extrawurst 

I plan to submit a follow-up PR with selection-based highlighting which I think will clear up the remaining vie
wing complexities (As ascii characters can only show so much, realistically). This would look like cutting over the "background" lines when a particular lineage is focused.

balintbarna:
Hi all, if you need a reference for what looks good and works well in TUI then lazygit (go) and git-igitt (rust) are both very good at visualizing the graph. I think the table layout of lazygit makes more sense:

short hash | timestamp | name | graph | (tags, branches) message

it can also switch between --all mode and current branch
and the colors and highlighting are great
I think it would be a good source of inspiration for the desired look
it also has multi-commit selection by holding down shift and a lot of useful shortcuts on this view, such as squash, fixup, drop, edit, amend, checkout, reset, cherry-pick, which is probably a separate issue from visualizing it, but mentioned it for completeness sake.

https://github.com/jesseduffield/lazygit#commit-graph
https://github.com/git-bahn/git-igitt

image

philocalyst:
When I looked at the algo's of git-gitt there was some issues around their design as a second-pass over a complete input, which is counter to the streaming design of gitui. Definitely still some issues I realize I miss when I review, I think there's still some issues with connecting pipes, but the tradeoffs of the algorithm feel solid to me.

balintbarna:
@philocalyst thanks for your reply. I don't have any insight on the code of the above mentioned projects, my comment is purely from a user perspective.

I'm a software professional and my main way of interacting with git is lazygit, before that I used the integration in vs code with the very popular git graph extension and before that I used source tree. I also briefly tried gitkraken and more recently gitui. I use the CLI when I need to or when it's simpler.

While lazygit has some serious bugs and I disagree with some of their decisions which make me miss git graph from vs code, it is my favourite tool so far, and a big part of that is their graph view, and even if the first version in gitui isn't equivalent, I would like to suggest it as inspiration for what it could become, in addition to the also fantastic git graph extension in vs code and the visually appealing git-igitt.

philocalyst:
I'm with you !!

Yeah I do plan to do follow-up PRs, just adding in more of these complex previews might have to be done on a second pass on just the active window.

▰▰▰▰▰
Miles Wirth 🙃

asyncgit/src/graph/oids.rs
```diff
@@ -0,0 +1,40 @@
+use crate::sync::CommitId;
+use std::collections::HashMap;
+
+pub struct Oids {
+	/// alias
+	pub ids: Vec<CommitId>,
```
cruessler: Just looking at `oids.rs`, I feel like `ids` could be removed as all the information it contains already lives in `aliases`. This would also entirely remove the risk that both could get out of sync.

asyncgit/src/graph/oids.rs
```diff
@@ -0,0 +1,40 @@
...
+			aliases: HashMap::new(),
+		}
+	}
+
+	pub fn get_or_insert(&mut self, id: &CommitId) -> u32 {
+		if let Some(&alias) = self.aliases.get(id) {
+			return alias;
+		}
+		let alias =
+			u32::try_from(self.ids.len()).expect("too many oids");
```
cruessler: Is there a way to remove the `expect`, e. g. by making `aliases` store `usize`s?

asyncgit/src/graph/oids.rs
```diff
@@ -0,0 +1,40 @@
...
+		Self {
+			ids: Vec::new(),
+			aliases: HashMap::new(),
+		}
+	}
+
+	pub fn get_or_insert(&mut self, id: &CommitId) -> u32 {
+		if let Some(&alias) = self.aliases.get(id) {
+			return alias;
+		}
```
cruessler: What do you think about changing this to `self.aliases.entry(*id).or_insert(...)`?

cruessler:
@philocalyst Thanks for the changes! I’ve now started the proper first review which should be ready by today or tomorrow if all goes according to plan! 😀

asyncgit/src/graph/mod.rs
```diff
@@ -0,0 +1,55 @@
+pub mod buffer;
+pub mod chunk;
+pub mod oids;
+pub mod walker;
+
+pub use walker::GraphWalker;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum ConnType {
```
cruessler: Is there any reason not to name this `ConnectionType`?

asyncgit/src/graph/oids.rs
```diff
@@ -0,0 +1,40 @@
+use crate::sync::CommitId;
+use std::collections::HashMap;
+
+pub struct Oids {
```
cruessler: Can you explain the purpose of this struct in a comment?

asyncgit/src/graph/oids.rs
cruessler: In case you decide to remove `ids`, could instances of `Oids` also be replaced by just `HashMap<CommitId, usize/u32>`?

philocalyst:
Thanks so much :)

▰▰▰▰▰
Miles Wirth 🙃

peso:
Hi there. I'm still working on getting git-graph, now gleisbau, into gitui. The speed requirement from gitui can be solved by splitting the global branch trace algorithm off to a background thread, and then start at head. This will feel as if it starts instantly, but it will traverse the entire repo in due time. The split also addresses the memory issue seen in the gitui+gitgraph prototype from about a year go.

This is a major refactoring, so it will take some time. git-igitt is part of the git-bahn family and uses git-graph/gleisbau under the hood.

I think it is a good idea to get a graph feature into gitui as soon as possible. Everybody wants it. Maybe gleisbau can produce a visual improvement, but the major step is simply getting a graph. Congratulations to @philocalyst for reaching that goal.

cruessler (changes requested):
I’m sorry it’s taking so long! In order to not make you wait longer, I’m now sending you my first batch of comments, hoping they are already useful. As soon as I’m done with the next batch, you’ll get the next round of comments. I hope that works for you!

asyncgit/src/graph/oids.rs
philocalyst: That's dandy!

asyncgit/src/graph/oids.rs
philocalyst: that's intelligent ! let me give it a shot

asyncgit/src/graph/mod.rs
philocalyst: no!

philocalyst:
Balls back in your court! @cruessler

cruessler:
> @philocalyst wrote:
> Balls back in your court! @cruessler

Really sorry it’s taking so long, I should be able to devote some time to this PR this week as I don’t have too many other obligations.

philocalyst:
@cruessler Appreciated! Have another PR I've been sitting on for LFS, but didn't want to overwhelm the team -- so I'll be monitoring my github notifications closely to make sure this week isn't wasted!!

asyncgit/src/lib.rs
```diff
@@ -8,7 +8,6 @@ It also provides synchronous Git operations.
 It wraps libraries like git2 and gix.
 */
 
-#![forbid(missing_docs)]
```
cruessler, line 11: I think this should be left in place. Or was there a particular reason to remove it?

asyncgit/src/revlog.rs
```diff
@@ -241,6 +274,18 @@ impl AsyncLog {
...
+				let infos = get_commits_info(
+					repo_path,
+					&entries[0..read],
+					1000,
+				)?;
+				let mut gw = arc_graph_walker.lock()?;
+				for info in &infos {
+					gw.process(info);
+				}
+			}
```
cruessler: This is something we unfortunately can’t do. We’re in the hot path of walking a repository’s complete history here, and calling `get_commits_info` for each and every commit would slow that walk down too much. (The same applies to the other place below where `get_commits_info` is called.)

I would suggest a different approach. Based on the assumption that people most of the time will only ever work with the most recent history, we should introduce a second log walker that collects child-parent relationships on demand. It would only collect these relationships as far as the viewport goes. The trade-off would be that when you jump to old parts of the history you would sometimes not immediately see the graph, but I think it would be worth it and it could even be mitigated by some sort of loading indicator.

Also, when the graph is not shown, I suggest to not load it at all.

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,327 @@
+use super::buffer::Buffer;
+use super::chunk::{Chunk, Markers};
+use super::oids::GraphOids;
+use super::{ConnectionType, GraphRow, MAX_LANE_COLORS};
+use crate::sync::{CommitId, CommitInfo};
+use im::Vector;
```
cruessler: I’m wondering why you chose `im::Vector` over `Vec`? Did you run any benchmarks that showed real-world benefits to using `im::Vector`?

asyncgit/src/graph/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+#[derive(Clone, Debug, Default)]
+pub struct GraphRow {
+	/// Number of active lanes at this commit row
+	pub lane_count: usize,
+
+	/// Which lane index this commit sits on
+	pub commit_lane: usize,
+
+	/// Whether this is a merge commit (two parents)
+	pub is_merge: bool,
```
cruessler, line 107: Does the algorithm also handle cases where a commit has more than two parents?

asyncgit/src/graph/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+
+	/// Whether this commit is a branch tip
+	pub is_branch_tip: bool,
+
+	/// Whether this commit has stash marker
+	pub is_stash: bool,
```
cruessler, line 113: If it reduces complexity, this could also be removed in the first implementation.

asyncgit/src/graph/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+
+	/// Connections emitted per lane:
+	/// None = empty space
+	/// Some((ConnectionType, `color_index`)) = draw this connector in this color
+	pub lanes: Vec<Option<(ConnectionType, usize)>>,
```
cruessler: Depending on the number of `GraphRow`s we need to hold in total, it might make sense to choose a representation that needs less memory.

asyncgit/src/graph/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+
+	/// Horizontal merge bridge: if this commit merges rightward,
+	/// (`from_lane`, `to_lane`) — the span to draw ─ ╭ ╮ across
+	pub merge_bridge: Option<(usize, usize)>,
```
cruessler: Same here and for `branches`: do we need `usize` or would smaller types be capable of holding the information we need as well?

asyncgit/src/graph/chunk.rs
```diff
@@ -0,0 +1,13 @@
...
+pub enum Markers {
+	Uncommitted,
+	Commit,
+}
+
+#[derive(Clone, Debug)]
+pub struct Chunk {
+	pub alias: Option<usize>,
+	pub parent_a: Option<usize>,
+	pub parent_b: Option<usize>,
```
cruessler: Commits can have more than two parents. Is the algorithm not supposed to handle that case?

cruessler (changes requested):
I’m slowly getting myself familiar with the PR, in order to make sure I actually understand myself what is going on. 😀 I’ve got a couple of questions and remarks, some more fundamental than others. I’m also going to push minor refactorings that don’t warrant a comment to your branch as well, in order to make some areas more idiomatic. I hope that’s ok, if not, please let me know!

asyncgit/src/lib.rs
philocalyst: That's explained earlier in this PR! Will add it back
>OH! And I just noticed that I removed the missing docs lint, I'll add that back if we agree on this general direction.

asyncgit/src/graph/walker.rs
philocalyst: This was just a convenience when porting from guitar, which [uses it](https://github.com/asinglebit/guitar/blob/676c3d9dded5f8aa41d5a1af64cff696d650eec1/Cargo.toml#L21). I'm sure there are performance benefits, however -- but I understand if you feel we don't need the dep.

asyncgit/src/graph/chunk.rs
philocalyst: This was covered elsewhere, but it's not designed to handle cases where there are more than two parents. Such cases are very rare and go much further than what a TUI could reasonably display (The l[inux kernel has merges with up to 66 parents](https://www.destroyallsoftware.com/blog/2017/the-biggest-and-weirdest-commits-in-linux-kernel-git-history)). I think drawing the line here, and representing this clearly in other ways (such as the status pane) is more reasonable.

asyncgit/src/revlog.rs
philocalyst: Got it, it already feels so much quicker than other clients so my privledge led to ignorance!

asyncgit/src/graph/mod.rs
philocalyst: [Covered here](https://github.com/gitui-org/gitui/pull/2890#discussion_r3375628994)

asyncgit/src/graph/mod.rs
philocalyst: It's not adding much imo! Let me know if that's a demand.

asyncgit/src/revlog.rs
cruessler: No worries, for repositories with less than 30k commits, there is not that much of a noticable difference on a modern machine. What I like to do for testing the app’s overall performance is start it in my copy of the Linux kernel which sits at about 1.4 M commits currently. There, every bit matters. 😀

asyncgit/src/graph/walker.rs
cruessler: I understand. I think it makes sense to use data structures from `std` in the first iteration unless there’s extremely compelling benchmarks that suggest otherwise. For this kind of operation, disk IO is probably the limiting factor anyway. (Your choice of `smallvec` makes more sense, based on my intuition, as it allows us to avoid a lot of heap allocations which can matter when the history is large. Though I haven’t run a benchmark yet, either.)

src/components/commitlist.rs
```diff
@@ -1069,4 +1355,24 @@ mod tests {
...
+			lanes: vec![Some((ConnectionType::CommitNormal, 0))],
+			merge_bridge: None,
+			branches: vec![],
+		};
+		let spans = cl.build_graph_spans(&row, 1);
+
+		assert_eq!(spans.len(), 2);
+		assert_eq!(spans[0].content, Cow::from(SYM_COMMIT));
+		assert_eq!(spans[1].content, Cow::from(SYM_SPACE));
+	}
```
cruessler, line 1378: It looks like this is the only test that’s added in this branch. Maybe that’s already on your radar, but just in case it’s not, I wanted to mention that `gitui` supports snapshot tests that should come in handy for the new features. The first example can be found here: https://github.com/gitui-org/gitui/blob/4f97fb5b1cd0ff2a0d2c7f1390236a1db02d8a13/src/gitui.rs#L235-L289

asyncgit/src/graph/walker.rs
balintbarna: If there are places where large vectors get copied, the immutable collections offer advantages, with almost free copies, and cheap changes to a copy without modifying the original, which is a unique combo.
Without looking at the code, my point is if that is the case it's worth doing some benchmarks, or just going with the better library. std vectors have the downside of either deep copy, or modifying the original as well when using references.

cruessler: I agree that it’s the benchmarks that matter! I maybe should have stressed the fact more that I’ve already run benchmarks for the log walker in the past where it became clear that decoding git objects from disk is by a large margin the area where the walk spends most of its time. And having looked at this PR, it looks like we’re talking about small vectors anyway, so I can’t imagine the theoretical gains of cheaper clones to even show up in this specific benchmark.

asyncgit/src/lib.rs
cruessler: Thanks! I had missed that bit!

asyncgit/src/graph/buffer.rs
```diff
@@ -0,0 +1,235 @@
...
+
+const CHECKPOINT_INTERVAL: usize = 100;
+
+pub struct Buffer {
+	pub current: Vector<Option<Chunk>>,
+	pub deltas: Vec<Delta>,
+	pub checkpoints: BTreeMap<usize, Vector<Option<Chunk>>>,
+	mergers: Vec<usize>,
+	pending_delta: Vec<DeltaOp>,
+}
```
cruessler, line 56: Can you add some docs on what the struct does as well as what its fields are for?

asyncgit/src/graph/buffer.rs
```diff
@@ -0,0 +1,235 @@
...
+#[derive(Clone, Debug)]
+pub struct Delta(pub Vec<DeltaOp>);
+
+const CHECKPOINT_INTERVAL: usize = 100;
+
+pub struct Buffer {
+	pub current: Vector<Option<Chunk>>,
+	pub deltas: Vec<Delta>,
+	pub checkpoints: BTreeMap<usize, Vector<Option<Chunk>>>,
+	mergers: Vec<usize>,
```
cruessler: Is this short for `merge_commits`? If yes, what do you think about renaming to `merge_commits`?

philocalyst: Totally!

src/components/commitlist.rs
philocalyst: I added quite a few new tests!

asyncgit/src/graph/walker.rs
philocalyst: Very useful to know!

asyncgit/src/revlog.rs
philocalyst: Tried on my own nixpkgs! Seems to be quite speedy!

philocalyst:
okay @cruessler you inspired me to do a pretty big refactor so sorry for the delay here! put a ton of effort into reducing the amount of redundant work as far as both clones and passing through the same commits/lanes twice or more. should be totally lazy as far as the graph goes, which i'm very happy to report! there's also a set of tests that should test the basics... (disclosure that i used $llm for the test writing, let me know if this goes against a policy and i will rewrite!)

Also I kind of overpromised with stashes, although I will do in a follow-up if all goes well :)

philocalyst:
OH! Also I managed to make the graph rendering the nicest it's ever looked!!! Excited for your thoughts :)

philocalyst:
Also started working on some **more** improvements, like mega-merges and path highlighting, but I would rather submit that after we get this one in!

cruessler:
> Also started working on some **more** improvements, like mega-merges and path highlighting, but I would rather submit that after we get this one in!

Makes sense, this PR is already quite large. 😀

I hope to get to do some more reviewing over the weekend or at the start of next week!

asyncgit/src/graph/buffer.rs
```diff
@@ -0,0 +1,281 @@
...
+
+	/// Scratch list of the [`DeltaOp`]s recorded for processing commit
+	pending_delta: Vec<DeltaOp>,
+}
+
+impl Default for Buffer {
+	fn default() -> Self {
+		Self::new()
+	}
+}
```
cruessler: Could this be replaced by `#[derive(Default)]` on `Buffer`?

asyncgit/src/graph/buffer.rs
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+	/// Remember `alias` as a merge commit whose second parent must be
+	/// given its own lane.
+	pub fn track_merge_commit(&mut self, alias: AliasId) {
+		self.merge_commits.push(alias);
+	}
+
+	pub fn update(&mut self, new_chunk: &Chunk) {
+		// Phase 1: place the new chunk into the lane array.
```
cruessler: The comments in `update` feel a bit redundant as the method names already carry some context.

asyncgit/src/graph/buffer.rs
```diff
@@ -0,0 +1,281 @@
...
+			self.record_insert(target, Some(new_chunk.clone()));
+		}
+		target
+	}
+
+	fn find_lane_awaiting_parent(
+		&self,
+		alias: Option<AliasId>,
+	) -> Option<usize> {
+		let alias = alias?;
```
cruessler: What do you think about turning this into `alias.map(…)`?

asyncgit/src/graph/buffer.rs
```diff
@@ -0,0 +1,281 @@
...
+
+		// Phase 3: flush any pending merge commits into new lanes.
+		self.flush_merge_commits();
+
+		// Phase 4: commit the delta and maybe checkpoint.
+		self.commit_delta();
+	}
+
+	fn place_chunk(&mut self, new_chunk: &Chunk) -> usize {
+		// Prefer a lane whose current occupant is waiting for this chunk as parent_a.
```
cruessler: What does `parent_a` refer to?

asyncgit/src/graph/oids.rs
```diff
@@ -0,0 +1,35 @@
...
+use std::collections::HashMap;
+
+/// mapping of `CommitId` to a numeric alias
+pub struct GraphOids(HashMap<CommitId, AliasId>);
+
+impl Default for GraphOids {
+	fn default() -> Self {
+		Self::new()
+	}
+}
```
cruessler: Could this be turned into `#[derive(Default)]` on `GraphOids`?

asyncgit/src/graph/mod.rs
```diff
@@ -0,0 +1,108 @@
...
+// On most terminals can more than 256 lanes even be represneted usefully? Not really.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct LaneIndex(u8);
+
+/// Numeric alias assigned to each commit in the graph.
+///
+/// The alias is a dense integer index created by [`GraphOids`](super::oids::GraphOids)
+/// that avoids storing full [`CommitId`](crate::sync::CommitId)s inside the lane state.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash)]
+pub struct AliasId(usize);
```
cruessler: The name `AliasId` was not immediately obvious to me when I first saw it. Is there maybe a different name that carries more context? Maybe `AliasForCommitId` or is that too clunky?

asyncgit/src/graph/chunk.rs
```diff
@@ -0,0 +1,15 @@
+use super::AliasId;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum Markers {
+	Uncommitted,
+	Commit,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Chunk {
```
cruessler: Is there a way to make the name `Chunk` more descriptive? Even after reviewing this PR for a bit, it doesn’t really click. 😀

asyncgit/src/graph/chunk.rs
```diff
@@ -0,0 +1,15 @@
...
+	pub alias: Option<AliasId>,
```
cruessler: Can you add some context on what `alias` being `None` means?

asyncgit/src/sync/commits_info.rs
```diff
@@ -122,6 +132,8 @@ pub struct CommitInfo {
 	pub author: String,
 	///
 	pub id: CommitId,
+	///
+	pub parents: SmallVec<[CommitId; 2]>,
```
cruessler, line 136: I’m wondering whether there is an easy way to also compute this information on demand. This could potentially also be explored in a follow-up.

asyncgit/src/graph/buffer.rs
```diff
@@ -0,0 +1,281 @@
...
+	fn first_empty_lane(&self) -> Option<usize> {
+		self.current.iter().position(Option::is_none)
+	}
+
+	fn consume_alias_in_other_chunks(
+		&mut self,
+		alias: AliasId,
+		skip_index: usize,
+	) {
+		for index in 0..self.current.len() {
```
cruessler: Is there a way to turn this loop into something like `for (index, chunk) in self.current.iter().enumerate()`? I’d like to avoid the extra bounds check inside the loop if possible.

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,851 @@
...
+			alias: Some(commit_alias),
+			parent_a: first_parent,
+			parent_b: second_parent,
+			marker: Markers::Commit,
+		};
+
+		if let Some(second) = second_parent {
+			self.merge_parents.insert(commit_alias, second);
+
+			if first_parent.is_some()
```
cruessler: Are there cases where `second_parent.is_some()`, but `first_parent.is_none()`? Apart from that, do you think something like `match (first_parent, second_parent)` could make the possible cases clearer?

cruessler (changes requested):
All my comments were supposed to part of a single review, but apparently I was not familiar enough yet with the tool I’m using to make that happen, so all of them arrived as individual comments. Sorry for causing that many notifications!

Also, would it be possible to add a couple more tests, so that all symbols in `graphrow.rs` are covered?

After having gone through maybe a third of half of the PR, I can say that I like the changes so far, but I was wondering whether there was a way to calculate `parents` on demand as well?

cruessler:
I’m not sure whether there’s something wrong with the algorithm. In my copy of https://github.com/GitoxideLabs/gitoxide, scrolling to the beginning of the history shows a few lanes that I think should be empty, but have lines in them. (This looks different in `tig`.)

image

philocalyst:
Hey! @cruessler worked through your comments, and the graph issue I think uncovered a nice sum-type abstraction that should cover more invariants (and also feels much nicer).

Rest of your feedback should be addressed. Pretty weak-sauce still on the tests, although they do cover all of the issues you've uncovered so far. Just not my strong suit, so if you have direction... I will implement.

asyncgit/src/graph/walker.rs
philocalyst: Match definitely helped here!

asyncgit/src/graph/chunk.rs
philocalyst: Yes yes!

asyncgit/src/graph/oids.rs
philocalyst: 100%

asyncgit/src/graph/buffer.rs
philocalyst: I like it!

asyncgit/src/sync/commits_info.rs
philocalyst: This is very fruitful, so not moving it to a follow-up! Thanks for paying such close attention.

philocalyst:
Fixed conflicts !

asyncgit/src/graph/buffer.rs
cruessler: You might have missed this comment.

asyncgit/src/graph/buffer.rs
cruessler: You might have missed this comment.

asyncgit/src/graph/buffer.rs
```diff
@@ -0,0 +1,316 @@
...
+		skip_index: usize,
+	) {
+		let current = self.current.clone();
+		for (index, slot) in current.into_iter().enumerate() {
+			let chunk = match slot {
+				Some(chunk) if index != skip_index => chunk,
+				_ => continue,
+			};
+
+			let new = match chunk {
```
cruessler: Can you make `new` more descriptive? Should this be `new_chunk`?

asyncgit/src/graph/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+/// was minted.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct UnwalkedAlias(CommitAlias);
+
+impl UnwalkedAlias {
+	/// The underlying alias, for comparisons against walked commits.
+	pub const fn get(self) -> CommitAlias {
+		self.0
+	}
+}
```
cruessler: Would it make sense to implement `Deref` for `UnwalkedAlias` with `Target = CommitAlias` as well? (Currently, there’s a few `.get()` calls that feel like they might not strictly be necessary.)

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+use super::oids::GraphOids;
+use super::{
+	CommitAlias, ConnectionType, GraphRow, LaneIndex, UnwalkedAlias,
+	MAX_LANE_COLORS,
+};
+use crate::sync::CommitId;
+use core::cmp::Ordering;
+use std::collections::{HashMap, HashSet};
+
+/// Get the lanes color index, which cycles through the ste palette.
```
cruessler: Is `ste palette` a typo?

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+
+	#[allow(clippy::missing_const_for_fn)]
+	fn vertical(self) -> bool {
+		self.intersects(Self::UP | Self::DOWN)
+	}
+}
+
+/// The sub network of an existing connection glyph
+/// `None` represents commit markers, which are never drawn over.
+fn conn_dirs(conn: ConnectionType) -> Option<Directions> {
```
cruessler: `conn_dirs` always returns `Some(x)`. Could the return type be changed to just `Directions`?

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+		(false, true, true, true) => ConnectionType::TeeDown,
+		(true, false, true, false) => ConnectionType::BranchUp,
+		(true, false, false, true) => ConnectionType::BranchUpRight,
+		(false, true, true, false) => {
+			ConnectionType::MergeBridgeStart
+		}
+		(false, true, false, true) => ConnectionType::MergeBridgeEnd,
+		(false, false, _, _) => ConnectionType::MergeBridgeMid,
+		(true, true, _, _)
+		| (true | false, false | true, false, false) => {
```
cruessler: `true | false` can be replaced by `_` unless I’m missing something.

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+		| ConnectionType::CommitStash
+		| ConnectionType::CommitUncommitted => return None,
+	})
+}
+
+/// Determine the glyph for a cell's connectivity.
+/// Vertical lines take precedence in crossed cells.
+/// Yet the horizontal bridge continues in
+/// the spacer columns either side, so we retain wholeness.
+const fn dirs_conn(dirs: Directions, dotted: bool) -> ConnectionType {
```
cruessler: The relation between `conn_dirs` and `dirs_conn` is not immediately obvious. Are there maybe more telling names?

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+/// Draw `add` into a cell, merging with whatever is already there.
+/// The line with a vertical component is the chosen way with color
+/// ensuring lanes stay visually continuous
+fn overlay_cell(
+	cell: &mut Option<(ConnectionType, LaneIndex)>,
+	add: Directions,
+	color: LaneIndex,
+) {
+	if let Some((conn, existing_color)) = cell {
+		if let Some(existing) = conn_dirs(*conn) {
```
cruessler: Unless I’m missing something `conn_dirs` always returns `Some(x)`, so the `else` branch is never taken. Is that an issue?

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+			let is_dotted =
```
cruessler: Minor: you could make this `conn.is_dotted()`.

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+	/// by [`Self::drawable_parent`], which refuses to mint an
+	/// [`UnwalkedAlias`] for any of them.
+	processed: HashSet<CommitAlias>,
+}
+
+impl Default for GraphWalker {
+	fn default() -> Self {
+		Self::new()
+	}
+}
```
cruessler: Could this also be `#[derive(Default)]`? (I would assume that’s more idiomatic.)

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+			buffer: Buffer::new(),
+			oids: GraphOids::new(),
+			merge_parents: HashMap::new(),
+			processed: HashSet::new(),
+		}
+	}
+
+	/// Mint the drawable alias for a parent commit, or `None` if the
+	/// walk already passed it.
+	fn drawable_parent(
```
cruessler: What about renaming to `mint_drawable_parent` to draw attention to the fact this method is about minting?

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+				parent,
+				second,
+			},
+			(Some(parent), None) => LaneSlot::Flowing {
+				alias: commit_alias,
+				parent,
+			},
+			// `second_parent` is only ever `Some` once
+			// `first_parent` has already been consumed from the
+			// identical iterator, so a merge always has both parents.
```
cruessler: What do you think about turning this comment into `(None, Some(_)) => unreachable!("…")` to capture that context more explicitly?

cruessler (changes requested):
This is a partial review (as I want to spend some time with friends now 😀). I’m leaving it, hoping to unblock you at least a bit. @philocalyst Again, thanks for your efforts!

From what I can tell, the PR is already in a good state. I tried it in my copy of the `gitoxide` repository which has a few interesting merges, and I didn’t spot any issues.

This comes with a few caveats, though:

- Since the PR was force-pushed, I was not able to pick up the review where I had left off after the previous one. This makes it harder to reconcile what’s new and what I’ve already seen which requires extra effort to review.
- This is a very large PR. I can manage to review maybe 700 lines per hour, but there’s limits to my spare time, in particular to my ability to dedicate a few hours uninterrupted to reviewing code. :-)

In light of that, I’m worried that I’m blocking this PR from moving forward. We might want to consider adopting a different strategy for getting it merged, such as merging behind a feature flag/an option and then finding the remaining issues. Another option would be to try and write more test cases and try to cover this feature as completely as reasonably possible.

@extrawurst What do you think?

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+				self.buffer.track_merge_commit(commit_alias);
+			}
+		}
+
+		self.processed.insert(commit_alias);
+		self.buffer.update(&chunk);
+	}
+
+	/// Number of commits already folded into the graph buffer.
+	pub const fn processed_commits(&self) -> usize {
```
cruessler: What do you think about making `number` or `len` part of the method’s name?

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+	) -> Vec<GraphRow> {
+		if commit_range.is_empty() {
+			return Vec::new();
+		}
+
+		// Decompress one row before the range to establish predecessor state
+		let snapshot_start_index =
+			global_start_index.saturating_sub(1);
+		let snapshot_end_index =
+			global_start_index + commit_range.len() - 1;
```
cruessler, line 256: Is this equivalent to `snapshot_start_index + commit_range.len()`? (If so replacing it would make it slightly easier to follow, I think.)

cruessler, line 256: Why is `global_start_index` shifted by `1` compared to `snapshot_start/end_index`? I’m not sure I understand based on context alone.

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+		commit_lane: usize,
+		current_snapshot: &[Option<LaneSlot>],
+		previous_snapshot: Option<&[Option<LaneSlot>]>,
+	) {
+		let Some((source_lane, target_lane)) = merge_bridge else {
+			return;
+		};
+		if source_lane == target_lane {
+			return;
+		}
```
cruessler, line 309: Can this happen in practice? If so, would it make sense to prevent the creation of such a merge bridge or would that have unintended consequences?

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+					LaneIndex::from(end_lane),
+				)
+			})
+			.collect()
+	}
+
+	/// Checks if tracking a merge commit would be redundant based on
+	/// current buffer state: some lane already flows to the target
+	/// parent without owing a second parent of its own.
+	fn is_redundant_merge_track(
```
cruessler: What do you think about `has_lane_to_parent`? Asking because I find `is_redundant_merge_track` harder to understand without looking at the implementation.

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+	) -> bool {
+		let exists_in_current =
+			current_snapshot.get(target_lane).is_some();
+		let matches_previous =
+			previous_snapshot.is_some_and(|previous| {
+				previous.get(target_lane)
+					== current_snapshot.get(target_lane)
+			});
+
+		exists_in_current && matches_previous
```
cruessler: What do you think about starting with the guard, like so:

```rust
if let Some(previous_snapshot) = previous_snapshot {
}
```

Maybe this would allow us to remove the second call to `current_snapshot.get(…)` as well.

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+			&mut lanes,
+			&branching_lanes,
+			commit_lane,
+		);
+
+		let active_lane_count =
+			current_snapshot.iter().flatten().count();
+
+		GraphRow {
+			lane_count: LaneIndex::from(active_lane_count),
```
cruessler: Not sure whether to add this comment, but this could also be `active_lane_count.into()`.

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+	}
+
+	/// Locates the primary lane for the current commit.
+	fn find_commit_lane(
+		current_snapshot: &[Option<LaneSlot>],
+		commit_alias: Option<CommitAlias>,
+	) -> usize {
+		let Some(target_alias) = commit_alias else {
+			return 0;
+		};
```
cruessler: I’m wondering whether this method should return `None` when `commit_alias.is_none()` as I’m not sure what the implications of returning `0` are for the caller.

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+	fn calculate_merge_bridge(
+		current_snapshot: &[Option<LaneSlot>],
+		commit_lane: usize,
+		second_parent_alias: CommitAlias,
+	) -> Option<(usize, usize)> {
+		current_snapshot
+			.iter()
+			.position(|chunk_option| {
+				chunk_option.as_ref().is_some_and(|chunk| {
+					chunk.awaits() == Some(second_parent_alias)
```
cruessler: What do you think about changing `awaits` to `waits_for_second_parent(second_parent_alias)`? This line would be slightly easier to read, I think.

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+		previous
+			.iter()
+			.enumerate()
+			.filter(|(index, previous_chunk)| {
+				previous_chunk.is_some()
+					&& current_snapshot
+						.get(*index)
+						.is_none_or(Option::is_none)
+			})
+			.map(|(index, _)| index)
```
cruessler: Could this be combined to `filter_map`?

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+		chunk: &LaneSlot,
+		lane_index: usize,
+		head_alias: Option<CommitAlias>,
+	) -> Option<ConnectionType> {
+		// A settled lane draws nothing below its commit.
+		if matches!(chunk, LaneSlot::Settled { .. }) {
+			return None;
+		}
+
+		let is_dotted = lane_index == 0
```
cruessler: Why is this limited to `lane_index == 0`?

cruessler (changes requested):
New batch of comments. I expect to be able to have more time tomorrow as I’ll be on a train for a couple of hours. I want to spend them on wrapping up the review!

philocalyst:
Thank you SO much for your review! Your thoroughness is deeply admired -- it's humbling to fix these mistakes.

As far as merge strategy goes, I am okay with writing more test-cases. Not really a fault of anyone in particular, but the response times in this project are measured in weeks/months in my experience, and I would rather not have to justify the removal of the flag-gating to eyes months removed (or have you guys have to continously re-acquaint).

asyncgit/src/graph/walker.rs
philocalyst: No that's fair

asyncgit/src/graph/walker.rs
philocalyst: It would, yes.

asyncgit/src/graph/walker.rs
philocalyst: i like this

asyncgit/src/graph/walker.rs
philocalyst: I will comment, forgot myself.

asyncgit/src/graph/walker.rs
philocalyst: I like it

asyncgit/src/graph/walker.rs
philocalyst: Much cleaner yes

asyncgit/src/graph/walker.rs
philocalyst: definitely

asyncgit/src/graph/walker.rs
philocalyst: After some cleanup yes

asyncgit/src/graph/walker.rs
philocalyst: love this

asyncgit/src/graph/walker.rs
philocalyst: Great catch, was refactored

asyncgit/src/graph/walker.rs
philocalyst: Yes :(

asyncgit/src/graph/mod.rs
philocalyst: I appreciate it

asyncgit/src/graph/buffer.rs
philocalyst: i was thinking implicitly, am okay with that form too

philocalyst:
And apologies about the force push, have been doing a lot of work on nixpkgs where it's preferred. I believe github preserves the original, but if you would like to see, I have it all in my reflog.

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+		);
+
+		// row of commit 3: lane 1 is crossed by the merge bridge but
+		// keeps both its vertical glyph and its own lane color
+		let crossed = computed[2].lanes[1]
+			.expect("crossed lane should not be empty");
+		assert_eq!(crossed.0, ConnectionType::Vertical);
+		assert_eq!(crossed.1, lane_color(1));
+	}
+}
```
cruessler, line 973: Would it make sense to add a test for a history with more than one root?

asyncgit/src/graph/walker.rs
```diff
@@ -0,0 +1,915 @@
...
+			&HashSet::new(),
+			None,
+		);
+
+		// row of commit 3: lane 1 is crossed by the merge bridge but
+		// keeps both its vertical glyph and its own lane color
+		let crossed = computed[2].lanes[1]
+			.expect("crossed lane should not be empty");
+		assert_eq!(crossed.0, ConnectionType::Vertical);
+		assert_eq!(crossed.1, lane_color(1));
```
cruessler, line 931: This looks like a test for a very particular edge case. Are there other similar edge cases that would benefit from such a test as well?

asyncgit/src/sync/commits_info.rs
```diff
@@ -156,6 +168,11 @@ pub fn get_commits_info(
 				author,
 				time: c.time().seconds(),
 				id: CommitId(c.id()),
+				parents: c
+					.parents()
+					.take(2)
+					.map(|p| CommitId(p.id()))
+					.collect(),
```
cruessler, line 175: I haven’t gotten yet to profiling this addition against current `master`. It’s on my list, though.

src/main.rs
```diff
@@ -33,7 +33,6 @@
 
 #![forbid(unsafe_code)]
 #![deny(
-	mismatched_lifetime_syntaxes,
```
cruessler, line 36: I’m wondering whether this removal was necessary.

src/gitui.rs
```diff
@@ -226,7 +226,7 @@ mod tests {
 			// Linux Temp Folder
 			settings.add_filter(r" */tmp/\.tmp\S+-insta/", "[TEMP_FILE]");
 			// Commit ids that follow a vertical bar
-			settings.add_filter(r"│[a-z0-9]{7} ", "│[AAAAA] ");
+			settings.add_filter(r"│[^a-f0-9]*[a-f0-9]{7} ", "│[AAAAA] ");
```
cruessler: What was this change motivated by? The regex has become less restrictive and I’d like to reduce the risk of false positives as much as possible.

src/components/utils/logitems.rs
```diff
@@ -78,12 +83,14 @@ impl LogEntry {
...
 
-///
+/// A batch of parsed log entries with an index offset.
 #[derive(Default)]
 pub struct ItemBatch {
 	index_offset: Option<usize>,
 	items: Vec<LogEntry>,
 	highlighting: bool,
+	pub graph_ready: bool,
+	pub max_lane: usize,
```
cruessler, line 93: There’s now a mismatch between some fields being `pub` while others are not. Are there downsides to making the new fields not `pub`?

asyncgit/Cargo.toml
```diff
@@ -29,6 +30,8 @@ rayon = "1.11"
 rayon-core = "1.13"
 scopetime = { path = "../scopetime", version = "0.1" }
 serde = { version = "1.0", features = ["derive"] }
+smallvec = "1"
+ssh-key = { version = "0.6.7", features = ["crypto", "encryption"] }
```
cruessler: `ssh-key` has been removed from `master` recently.

asyncgit/Cargo.toml
```diff
@@ -20,6 +20,7 @@ fuzzy-matcher = "0.3"
 git2 = { version = "0.21", features = ["https"] }
 git2-hooks = { path = "../git2-hooks", version = "0.7" }
 gix = { version = "0.84.0", default-features = false, features = ["mailmap", "max-performance", "revision", "sha1", "sha256", "status"] }
+im = "15"
```
cruessler: What is this required for? On my machine, `cargo check` doesn’t complain if I remove this line.

cruessler (changes requested):
Another batch of comments. There’s only 7 files left to review now, most importantly `asyncgit/src/revlog.rs` and `src/components/commitlist.rs`.

src/gitui.rs
philocalyst: That's leftover from an old debugging session !

asyncgit/Cargo.toml
philocalyst: Leftovers from old work! Been so long, we've started to forget...

src/components/utils/logitems.rs
philocalyst: Fixed this where there's not a real use for a getter

asyncgit/src/sync/commits_info.rs
philocalyst: Thanks :0

asyncgit/src/graph/walker.rs
philocalyst: Added two more :)

asyncgit/src/graph/walker.rs
philocalyst: It does

philocalyst:
@cruessler All good?

philocalyst:
@cruessler Still progressing?

andras-tim:
_sorry for the offtopic, but I can't wait this feature for months..._ :sweat:
Could you please to merge it if there's no serious issues? This is an awesome feature that would make the `gitui` tui complete! (I'm going to finally switch from `git-gui` & `gitk` to `gitui` :heart_eyes:)

philocalyst:
Will fortune and fate it should be complete, just waiting on final review, btw @cruessler that force push is with lease, it was just a cargo.toml change! Please don't fret.

cruessler:
This PR is on top of my TODO list for the coming week! I have drafted a benchmark to make sure there’s no performance regressions, but I have not yet gotten to committing and pushing that code.

cruessler:
@philocalyst The benchmark is available in this commit: https://github.com/cruessler/gitui/commit/d372bfafbcf3f1dbce83538584c6330bb856fcaa. This commit can be applied on top of your PR. The benchmark exercises a history walk with and without creating the history graph. Would you mind double-checking that it actually measures your changes? (I had it drafted by an LLM and want to make sure I didn’t miss anything in my review.)

Also, could you run the part without graph creation on your machine and compare against `master`? (This requires some modifications to the benchmark, but should not be too complicated.)

Example result on my machine:

```
❯ GITUI_BENCH_REPO=~/github/extrawurst/gitui/ cargo bench -p asyncgit --bench full_history
    Finished `bench` profile [optimized] target(s) in 0.13s
     Running benches/full_history.rs (target/release/deps/full_history-1e929556af4de21d)
full_history/without_graph
                        time:   [17.477 ms 17.660 ms 17.786 ms]
                        thrpt:  [146.18 Kelem/s 147.23 Kelem/s 148.77 Kelem/s]
                 change:
                        time:   [+0.4870% +1.6229% +2.7906%] (p = 0.02 < 0.05)
                        thrpt:  [−2.7148% −1.5970% −0.4847%]
                        Change within noise threshold.
full_history/with_graph time:   [35.523 ms 35.901 ms 36.392 ms]
                        thrpt:  [71.444 Kelem/s 72.421 Kelem/s 73.193 Kelem/s]
                 change:
                        time:   [−0.2869% +1.1908% +2.8456%] (p = 0.17 > 0.05)
                        thrpt:  [−2.7669% −1.1768% +0.2878%]
                        No change in performance detected.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
```

So it looks like creating the graph requires about the same amount of time as walking the history in this example. Running the `full_history/without_graph` part on `master` requires about the same amount of time as on this branch, so it looks like your changes don’t affect the “graph is disabled”-case in this synthetic benchmark.

```
❯ GITUI_BENCH_REPO=~/github/extrawurst/gitui/ cargo bench -p asyncgit --bench full_history
    Finished `bench` profile [optimized] target(s) in 0.10s
     Running benches/full_history.rs (target/release/deps/full_history-164177db5604f630)
full_history/without_graph
                        time:   [17.510 ms 17.691 ms 17.808 ms]
                        thrpt:  [146.01 Kelem/s 146.97 Kelem/s 148.49 Kelem/s]
                 change:
                        time:   [+0.3010% +1.5661% +2.7256%] (p = 0.03 < 0.05)
                        thrpt:  [−2.6532% −1.5420% −0.3001%]
                        Change within noise threshold.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) low mild
```

PR #2906, opened 2026-04-12
Auto return to the tree view if the last hunk in a file was staged

### LINKED ISSUE #2748: After staging, next unstaged change is janky

WaterWhisperer:
This Pull Request fixes/closes #2748.

It changes the following:
- auto return to the tree view after staging the last hunk from the status diff view

[录屏 2026年04月12日 16时40分45秒.webm](https://github.com/user-attachments/assets/9f598f71-d31a-4deb-a6b6-bc20da0ca4e9)

I followed the checklist:
- [ ] I added unittests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 * use [tombi](https://github.com/tombi-toml/tombi) for all toml file formatting
 * open the external editor from the status diff view [[@WaterWhisperer](https://github.com/WaterWhisperer)] ([#2805](https://github.com/gitui-org/gitui/issues/2805))
+* auto return to the tree view after staging the last hunk from the status diff view [[@WaterWhisperer](https://github.com/WaterWhisperer)] ([#2748](https://github.com/gitui-org/gitui/issues/2748))
 
 ### Fixes
 * crash when opening submodule ([#2895](https://github.com/gitui-org/gitui/issues/2895))
--- a/src/components/utils/statustree.rs
+++ b/src/components/utils/statustree.rs
@@ -64,7 +64,7 @@ impl StatusTree {
 		self.update_visibility(None, 0, true);
 		self.available_selections = self.setup_available_selections();
 
-		//NOTE: now that visibility is set we can make sure selection is visible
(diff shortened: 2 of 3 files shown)


WaterWhisperer:
About unittests: I started by adding simple unit tests to the new helper function, but eventually removed it because it was redundant and affected the code layout of the `status.rs` itself.

WaterWhisperer:
#2748 also mentions the issue of jumping to the directory level instead of the first file in the directory, which is the file that follows, but I think it depends on personal habits, and in some cases it is necessary to stage the entire folder directly, and I feel that this is reasonable behavior for the user, "if I want to stage the file separately then I just need to move it to it", so I didn't make any changes to this part

PR #2918, opened 2026-04-25
asyncgit: honour branch.sort config when listing branches

### LINKED ISSUE #2024: feature request for support branch.sort

Scolliq:
Closes #2024.

**Why:** `get_branches_info` hardcodes a lex-asc sort on the branch name, so users with `branch.sort = -committerdate` (or `authordate`, or `-refname`) in their git config see GitUI's branch list ignore that preference.

**What changed:** read `branch.sort` from the repo config and apply it. Supports `refname`, `committerdate`, `authordate`, each with the optional `-` prefix for descending. Anything else (including `version:refname`, which I left out for scope) falls back to the existing refname-ascending behaviour. Public API unchanged.

**Tests:** 6 parser unit tests covering recognised values, the descending prefix, whitespace, and unknown-falls-back. 2 integration tests with `write_commit_file_at` to pin the actual sort behaviour for `-committerdate` and the unset-config baseline.

`cargo test -p asyncgit --lib sync::branch` → 36 pass. `cargo clippy -p asyncgit --all-features` clean. `cargo fmt --check` clean.

**Future scope:** `version:refname` would need a semver-aware comparator — happy to follow up if there's interest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Changed
+* honour `branch.sort` git config when listing branches, including support for comma-separated multi-key values ([#2918](https://github.com/gitui-org/gitui/pull/2918))
 * use [tombi](https://github.com/tombi-toml/tombi) for all toml file formatting
 * open the external editor from the status diff view [[@WaterWhisperer](https://github.com/WaterWhisperer)] ([#2805](https://github.com/gitui-org/gitui/issues/2805))
 
--- a/asyncgit/src/sync/branch/mod.rs
+++ b/asyncgit/src/sync/branch/mod.rs
@@ -250,10 +250,12 @@ impl BranchSort {
 	}
 
 	fn parse(raw: &str) -> Self {
-		let trimmed = raw.trim();
-		let (descending, key) = trimmed
+		// git accepts comma-separated lists (e.g. `-committerdate,refname`);
+		// honour only the primary (first) key.
+		let primary = raw.split(',').next().unwrap_or("").trim();
+		let (descending, key) = primary
 			.strip_prefix('-')
-			.map_or((false, trimmed), |rest| (true, rest));
+			.map_or((false, primary), |rest| (true, rest));
 		match key {
 			"committerdate" => Self {
 				field: BranchSortField::CommitterDate,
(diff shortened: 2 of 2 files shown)


extrawurst:
Screenshot 2026-04-27 at 08 58 29

Thanks for taking care of this. The only problem i see is that we make the remote branch list very hard to read if there are different remotes. I know this is exactly as `git branch --remote` lists them but there must be a better way. 

I can see a better solution to do a hirarchical tree view first and only sort by this factor inside the same "folder". Is that something you could see yourself doing as a followup?

Right now this is likely still better than the status quo. Please add a changelog entry and rebase to current upstream master.

Also it would be great to update https://github.com/gitui-org/gitui/pull/2146 on top of this where we store the GUI selected sorting into the local git config and use that as the source of truth.

extrawurst:
Git accepts a comma-separated list (e.g. -committerdate,refname). `BranchSort::parse` treats the whole string as one key, so `-committerdate,refname` falls through to the default. Either support it or document the subset.

Scolliq:
yeah fair point about the remote display. i'm thinking hierarchical tree view as a followup PR rather than blocking this one — does that work?

added the changelog entry, already rebased on top of upstream master. happy to pick up #2146 after this lands too

also went ahead and added support for comma-separated sort keys (e.g. `-committerdate,refname`) — just uses the first key as primary. added a test for it as well

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Changed
+* honour `branch.sort` git config when listing branches, including support for comma-separated multi-key values ([#2918](https://github.com/gitui-org/gitui/pull/2918))
```
extrawurst, line 11: * please adopt the format we use (your authorship is missing)
* and clarify we only respect the first sort key though.

asyncgit/src/sync/branch/mod.rs
```diff
@@ -744,6 +814,154 @@ mod tests_branches {
...
+			"older commit",
+			git2::Time::new(1_000_000, 0),
+		);
+		create_branch(repo_path, "older").unwrap();
+
+		write_commit_file_at(
+			&repo,
+			"b.txt",
+			"b",
+			"newer commit",
```
extrawurst, line 923: this test works wrong. you commit `newer commit` actually to the branch `older`

extrawurst (changes requested):

PR #2919, opened 2026-04-25
feat(navigation): vim like keybindings for basic nav

blagoySimandov:
It changes the following:
- Adds  vim-like navigation: `j`/`k` for up/down, `l` to open preview

I followed the checklist:
- [X] I added unittests


blagoySimandov:
I was considering adding h as well 
- when in preview mode h exits it
- when exploring the file tree h closes a folder

I scraped it since currently h is bound to 'help'

PR #2924, opened 2026-05-16
fix: restrict `e` to HEAD, add `O` to open historical files read-only (#2147)

### LINKED ISSUE #2147: edit only for head revision otherwise open as temple
right now we show a `edit [e]` cmd for any file at any revision in file tree views. but we only open the file as it is currently on disk.

this means you don't see then what you expected when you looked at an old diff/revision. or worst case: the file is no longer present and will error.

we should only show `edit [e]` when you are at the head revision and otherwise two other options:
1. `open [o]` which will dump the revision of the file in a tempfile and open that
2. `go to head [g]` to jump the that file in the current file-tree (if it present)

2. is not as trivial and should probably be a separate issue (feel free to open that)

Metbcy:
This Pull Request fixes/closes #2147.

It changes the following:
- `e` (edit) on the Files-at-revision view is now only enabled when the revision being viewed is HEAD. Previously `e` always opened the current worktree copy regardless of which commit you were browsing, which was confusing and could lead to edits being misattributed to the wrong revision.
- A new `O` (shift+o) command is enabled only on non-HEAD revisions. It writes the blob at the selected revision into a fresh tempdir (prefixed with the full commit SHA), marks the file read-only, and opens it in the external editor.
- The tempdir is owned by the `RevisionFilesComponent` and is cleaned up when the next `O` replaces it or when the component is dropped, so there is no on-disk leak.
- On `O`, the view switches to the Status tab so it is visually clear that edits to the tempfile will not be persisted to the historical revision.

### Testing

A new unit test (`dumps_blob_to_readonly_tempfile_and_cleans_up_on_drop`) covers the dump-to-readonly + drop-on-replace behavior. Full suite passes:

```
cargo nextest run --workspace
# 310 tests, 0 failed
```

I also smoke-tested manually on a small repo:

| Scenario | Expected | Actual |
|---|---|---|
| At HEAD, `e` on a file | Opens worktree copy (writable) | ✅ perms 644, real path |
| At HEAD, `O` on a file | Ignored | ✅ no-op |
| At older commit, `e` | Ignored | ✅ no-op |
| At older commit, `O` | Opens tempfile, read-only, historical content | ✅ perms 444, `/tmp/<sha><rand>/<file>`, correct blob |
| Open a second file at the same revision | Previous tempdir dropped first | ✅ old tempdir gone |
| Quit gitui | Tempdir removed | ✅ |

### Notes

- Previous attempt at this issue (#2384) was closed by its author with the checklist unfinished; this PR is an independent implementation.
- `is_head()` returns `false` on lookup error, which keeps the safe behavior of preferring `open` over `edit`.

I followed the checklist:
- [x] I added unittests
- [x] I ran `make check` without errors (skipped `deny`/`tombi` locally  CI will likely cover)


Metbcy:
Closing in favor of #3012, which now covers #2147 on current master.

PR #2962, opened 2026-06-13
Update `git2` to 0.21 across crates

cruessler:
This PR updates the `git2` dependency to 0.21.

Among other things, this brings a new feature flag: `unstable-sha256`. This feature flag is not enabled in this PR. This could be done in a follow-up.

Breaking changes in upstream changelog: https://github.com/rust-lang/git2-rs/blob/main/CHANGELOG.md#changed

Relevant for `gitui`:

- “The `ssh`, `https`, and `cred` Cargo features are no longer enabled by default.” `asyncgit` now explicitly enables `https`.
- “Many string accessors that previously returned `Option<&str>` now return `Result<&str, Error>` or `Result<Option<&str>, Error>`, so callers can distinguish a missing value from a non-UTF-8 one.” In this PR, I’ve decided to keep the existing behaviour by treating `Err` values as `None`. Let me know if you want any changes here!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1273,16 +1273,14 @@ dependencies = [
 
 [[package]]
 name = "git2"
-version = "0.20.4"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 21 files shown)


extrawurst:
Awesome! Thank you!

PR #2967, opened 2026-06-17
Update `gix` to 0.84

cruessler:
This is a follow-up to #2962.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1079,7 +1079,7 @@ version = "0.7.0"
 dependencies = [
  "git2",
  "git2-testing",
- "gix-path 0.11.0",
+ "gix-path",
  "log",
  "pretty_assertions",
  "shellexpand",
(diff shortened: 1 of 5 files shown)


extrawurst:
@cruessler what does it take to mark this as out-of-draft?

cruessler:
@extrawurst `ssh-key` is still a release candidate at `0.7.0-rc.10`. I’m fine with merging as-is, but I wanted to at least have mentioned it. With respect to changes related to `gitoxide`, there’s nothing left on my to-do list.

extrawurst:
@cruessler there is a new ssh-key rc. still no release. a gitui crates release would fail with this - so i opt to wait :(

cruessler:
Since #2464 got merged which removed the dependency on `ssh-key`, there’s no need any longer to wait for its new release, so this PR is now finally ready for review!

extrawurst:
@cruessler any particular improvements or unlocks we will get from this now?

cruessler:
Yes! This enables `sha256` for `gix`. I’ll soon open a follow-up PR that enables `unstable-sha256` for `git2` at which point we should have support for SHA-256 repositories in `gitui`. Based on what I know, most, if not all, APIs we use should already work with SHA-256 without us having to change anything.

PR #2979, opened 2026-06-18
fix(staging): preserve no-trailing-newline when staging partial hunks

Noethix55555:
## Summary

`finish()` in `NewFromOldContent` unconditionally appended `\n` to the reconstructed file content, even when the original file had no trailing newline. The `DeleteEOFNL` / `AddEOFNL` diff markers were silently discarded with a bare `break`, so staging or discarding a single line in a no-newline file would silently corrupt it by adding a newline.

- Add `trailing_newline: bool` field to `NewFromOldContent` (default `true`).
- In `apply_selection`, when `DeleteEOFNL` or `AddEOFNL` is encountered, set `trailing_newline` based on whether the change is being applied or reverted, then `break`.
- In `finish()`, only append `\n` when `self.trailing_newline` is true.
- Add `test_stage_preserves_no_trailing_newline` in `stage_tracked.rs` that stages a single changed line in a two-line file with no trailing newline and asserts the staged blob content is also newline-free.

### DIFF
--- a/asyncgit/src/sync/staging/mod.rs
+++ b/asyncgit/src/sync/staging/mod.rs
@@ -13,10 +13,20 @@ use std::{collections::HashSet, fs::File, io::Read};
 
 const NEWLINE: char = '\n';
 
-#[derive(Default)]
 struct NewFromOldContent {
 	lines: Vec<String>,
 	old_index: usize,
+	trailing_newline: bool,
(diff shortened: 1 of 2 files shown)


Bolt4243:
Subsumed by #2977, which already includes this fix.

PR #2980, opened 2026-06-23
git LFS support

### LINKED ISSUE #2812: git-lfs support
Our current bug #2809 (our broken pre-push hook support) led me down the rabbit hole of "what else is required to properly support git-lfs in gitui.

We are looking at basically three things:
1. properly run pre-push hook (thats #2809)
2. hook into filter logic and run them on all checkout-like operations (smudge)
3. hook into filter logic and run them on all worktree->index operations (clean)

Now libgit2 exposes an API to hook into filters but git2-rs does not expose this right now ([open issue](https://github.com/rust-lang/git2-rs/issues/442)).

Since our end goal is to be libgit2 independent I call out whats needed to do this all using gitoxide:

1. we need to migrate all our git-checkout locations to gitoxide
2. migrate all worktree->index operations to gitoxide
3. ~~use gitoxide machinery to hook into filters~~ (turns out gitoxide will already invoke filters for us)
4. ~~forward filter logic via shellout (`git-lfs clean`, `git-lfs smudge` or `git-lfs filter-process`)~~
5. finally we support basic git-lfs
6. add support for more file locking relevant lfs features via hooks: `post-checkout`, `post-merge`, `post-commit`

This issue is more of an epic, I would love to see contributions to take on individual steps as separate PRs. 

cc @Byron @cruessler correct me if I am wrong but from the gitoxide side all this is ready to be used or am I wrong?

Further reading: https://github.com/gitui-org/gitui/discussions/1089

philocalyst:
This Pull Request fixes #2812 

It changes the following:
- This adds Git LFS support through the [git-lfs](https://gitlab.com/rustutils/git-lfs) implementation
	- There ARE some [failing test-cases](https://gitlab.com/rustutils/git-lfs/-/blob/master/tests/SCOREBOARD.md) within the implementation, it's not as perfect as shelling out. But it DOES happen to work!
	- Most of the fails are related to bulletproofing, like ssh, credentials, etc. So... kind of yikes, but I know this will improve. (And we don't have another option)
- This is a very basic/introductory PR, laying the groundwork for the finer features like lock/unlock, pruning, status, etc.
- I've elected to go with a simple [LFS] marker rather than some new symbol.
- - Respects "GIT_LFS_SKIP_SMUDGE"
- Tries to be pretty careful with attempting smudge and related LFS operations
- In retrospect perhaps a bit log-heavy compared to what I see in related code, part of that was debugging, so let me know if it's overwhelming.
- Went with git2 in the asyncgit portion of this PR because of the embedded nature of a lot of these new functions.

I followed the checklist:
- [x] I added unittests


philocalyst:
Converting to draft because everytime I try to add new features I go up against the brutal limitations of the library. It's the only one available, but I find myself reimplementing transport mechanisms and that's where it's too messy for me.

Are we okay with shelling out to the git-lfs binary for those parts? And keeping it mixed? Think that's the most pragmatic in the short-term.

extrawurst:
> Are we okay with shelling out to the git-lfs binary for those parts

Thats ok, let's just make sure to abstract behind a structure that allows us to collect the shelled-out parts in a single place we could exchange

PR #2991, opened 2026-07-09
Change walk order from ByCommitTime to topological order

peso:
This is preparation for a graph implementation.

Currently, commits are sorted by commit time without regard to parent-child relations. This means a parent with a date more recent than its child will appear before its child. `git log` has options `--date-order` and `--author-date-order` both of which show no parents before all its children, before ordering commits by date.

If sorted by commit time, a graph may have to draw a parent before its child. This is confusing to the user and require extra memory for the graph render algorithm. Topology order makes graph visualisation much simpler as all children are now visited before their parents.

This Pull Request prepares a PR for issue #81. It is relevant to both PR #2628 and PR #2890.

It changes the following:
- Commits was ordered by commit time, but are now in topology order. 

I followed the checklist:
- [x] I added unittests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * open the external editor from the status diff view [[@WaterWhisperer](https://github.com/WaterWhisperer)] ([#2805](https://github.com/gitui-org/gitui/issues/2805))
 * automatically convert spaces to dashes when creating or renaming a branch [[@pbouillon]](https//pbouillon.github.io)] ([#2916](https://github.com/gitui-org/gitui/pull/2916))
 * support rewording non-HEAD commits when `commit.gpgsign` is enabled (gpg format only) [[@guerinoni](https://github.com/guerinoni)] ([#2959](https://github.com/gitui-org/gitui/pull/2959))
+* always preserve topology order before sorting commits by date
 
 ### Fixes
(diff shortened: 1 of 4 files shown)


extrawurst:
@cruessler do you recall why we changed this in the first place?

cruessler:
I don’t remember why this change was made in the first place, but some digging through history revealed these commits that are related to the introduction of `TimeOrderedCommit` and what came before (from older to newer):

- Commit that ordered a regular revwalk by time: https://github.com/gitui-org/gitui/commit/680b178a376898d127372dcc301153b4549db444
- Commit that turned date-ordering off shortly after: https://github.com/gitui-org/gitui/commit/3adcb4b8c8aaaf9d3947a2dc61defacc191d7403
- Commit that switched from a `git2` revwalk to a custom one and introduced `TimeOrderedCommit`: https://github.com/gitui-org/gitui/commit/6006e7aebdc4bb5a14442c0f1508e4bf242e84b6

I presume there were cases where you preferred date-based ordering in the past and it has stayed that way ever since. 😀

cruessler:
I’ve created two benchmarks for this change in #3013, and it seems that `gitoxide`’s topo walk is significantly slower than the manual walk we did before. On my machine, it was ~120 ms on `master` vs. more than 400 ms in this PR. From what I remember, no-one has put a lot of effort into optimizing this walk yet which of course is unfortunate. Can you double-check my results?

The filtered log walker that still relies on `git2` was not affected as far as I can tell.

PR #3001, opened 2026-07-19
Enable `unstable-sha256` for `git2`

cruessler:
This PR is a follow-up to #2962 and #2967. It enables the `unstable-sha256` feature for `git2`. This change should enable `gitui` to work with SHA-256 repos since SHA-256 support for `gitoxide` was enabled via #2967 already.

### DIFF
--- a/asyncgit/Cargo.toml
+++ b/asyncgit/Cargo.toml
@@ -17,7 +17,7 @@ crossbeam-channel = "0.5"
 dirs = "6.0"
 easy-cast = "0.5"
 fuzzy-matcher = "0.3"
-git2 = { version = "0.21", features = ["https"] }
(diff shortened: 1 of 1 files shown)


cruessler:
@extrawurst I tested this branch locally and was able to open a SHA-256 repository. I’m wondering what amount of testing we want before we want to merge this change. If we wanted, we could run all tests twice, once with SHA-1 and once with SHA-256. This is something we do in `gitoxide` for the large majority of tests. `gix-testtools`, in particular, has support for both hashes, so we could also add `gix-testtools` as a dependency.

That said, though, I’m also fine with not adding much in terms of testing in this PR, knowing that test coverage in `gitoxide` at least is already quite thorough and assuming the same is true for `git2`. What do you think?

PR #3003, opened 2026-07-21
feat: add Graph tab with SourceTree-style branch lanes

tiagoyamashita:
## Summary
- Add a new top-level Graph [6] tab after Stashes that walks commits from HEAD, local branches, remotes, and tags
- Draw colored Unicode lane glyphs beside each commit so branch/merge topology is visible (SourceTree-style)
- Reuse existing commit-list ref badges ({*branch}, [remote/branch], <tag>) and Enter-to-inspect

## Test plan
- [ ] `cargo test -p asyncgit graph_`
- [ ] `cargo build`
- [ ] Open gitui, press `6` (or Tab to Graph)
- [ ] Confirm diverged local/remote branches show multiple colored lanes
- [ ] Confirm tags and HEAD branch labels appear on the correct commits
- [ ] Confirm Enter opens commit inspect from the Graph tab


src/app.rs
peso, line 351: You added a key_config.keys.tab_graph, but forgot to match against it to the if statement inside src/app.rs fn App.event that calls switch_tab. This means that the shortcut '6' does not activate the tab as intended.

peso (changes requested):

peso:
What are your thoughts behind placing the graph on a separate tab? 
I would expect it to be on the log tab. 
Fewer tabs gives a simpler interface that is faster to understand.

Perhaps with an option that toggles the graph on and off. Default would be on. 
The option would be in the OptionsPopup as shown in the CommandBar.
It could also be a dedicated key, if it was important to toggle it often.

PR #3005, opened 2026-07-21
fix(ui): restore log search option selection highlight

### LINKED ISSUE #3004: Search options no longer highlight as cursor moves.
**Describe the bug**
Search options no longer highlight as one scrolls through the options, making it hard to know what/where you're typing.

**To Reproduce**
1. Open the find tool with f
2. Try to change some options, notice it works, but no longer highlights.

**Expected behavior**
1. Open the find tool with f
2. See where your cursor is so you can toggle options.

**Screenshots**

https://github.com/user-attachments/assets/0546c506-46d5-4f8a-866a-665e268bf31e

 - OS/Distro + Version: Arch Linux
 - GitUI Version 0.28.1
 - Rust version: 1.97.1

sjuxax:
Fixes #3004 .

Option rows used theme.text(focused, false), which only dimmed non-focused lines and never applied selection_bg.

Use theme.text(true, selected) for option focus, refresh textarea style/cursor on enabled(), and keep jump-to-SHA field focused (invalid SHA still uses block danger styling).

This code was written by Grok 4.5.

Followed the checklist:
- [x] I added unittests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * crash when opening submodule ([#2895](https://github.com/gitui-org/gitui/issues/2895))
 * when staging the last file in a directory, the first item after the directory is no longer skipped [[@Tillerino](https://github.com/Tillerino)] ([#2748](https://github.com/gitui-org/gitui/issues/2748))
 * index-out-of-bounds panic when unstaging lines near the end of a diff ([#2953](https://github.com/gitui-org/gitui/issues/2953))
+* restore highlights on find tool ([#3004](https://github.com/gitui-org/gitui/issues/3004))
 
 ## [0.28.1] - 2026-03-21
 
--- a/src/components/textinput.rs
+++ b/src/components/textinput.rs
@@ -131,9 +131,34 @@ impl TextInputComponent {
 		self.embed = true;
 	}
 
-	///
-	pub const fn enabled(&mut self, enable: bool) {
+	/// Focus/unfocus the text input visually (style + cursor).
+	/// Used by popups that share the field with other selectable rows.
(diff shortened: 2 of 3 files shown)


CAOShurong (commented):
Code-grounded verification against the regression mechanism (couldn't build locally on Windows — openssl-sys wants a source build with perl/VC; CI covers the build).

I traced the regression independently of the issue text: since the tui-textarea migration (b9a2e131, first shipped in v0.25.0), TextInputComponent's TextArea is built once in show_inner_textarea() with the style captured at build time, and enabled() only flipped self.selected without touching the cached widget — so the input kept its "active" style and block cursor forever, and the option rows only ever had a default-vs-DarkGray foreground difference (theme.text(x, false) never hits the selection_bg branch). That matches the reported "no visible highlight while moving through options".

This PR addresses both halves of that mechanism:

1. enabled() now calls apply_enabled_style(), which mutates the cached textarea in place (set_style / set_placeholder_style / set_cursor_style REVERSED-vs-none), and show_inner_textarea() applies it after (re)builds — the focus state is finally visible on the input itself, not just the flag.
2. option_line() uses theme.text(true, selected), so the focused option gets the real selection_bg instead of relying on the fg-only cue. Nice side effect: the seven duplicated blocks collapse into one helper.

The new unit test asserting cursor_style REVERSED toggling covers the core of (1).

One deliberate behavior change worth a maintainer's explicit eye (not a defect): in the JumpCommitSha path, find_text.enabled(false) becomes enabled(true) with the comment "Field stays focused; invalid SHA is shown via block style" — the field now stays visually focused in that mode where it was dimmed before. If that's the intended UX for SHA jump mode, this is fine; just making sure it's a conscious choice rather than fallout from the refactor.

Test result on the head: not run locally (build blocker above); CI on this PR is the runtime evidence.

PR #3012, opened 2026-08-08
only offer edit on head revision, open older revisions as tempfile

### LINKED ISSUE #2147: edit only for head revision otherwise open as temple

MsfPablo:
Closes #2147

In the revision file tree `edit [e]` was offered for any file at any revision, but it always opened the file as it currently exists on disk. That means you don't see what you expected when looking at an old revision, or the editor errors out because the file no longer exists.

Now:
- `edit [e]` is only shown when the browsed revision is HEAD (existing behaviour, unchanged).
- For any other revision, `open [o]` is shown instead: it writes that revision's blob to a temporary file (`gitui-<short-id>-<name>`) and opens that in the external editor.

`go to head [g]` from the issue is left for a separate issue as suggested.

New key: `open_file_revision` (defaults to `o`). The revision file tree consumes it before the global options popup binding.

`cargo build`, `cargo clippy --all-features` and `cargo fmt --check` are clean.


MsfPablo:
Checking in on this one. If opening older revisions as a tempfile isn't the direction you want, I'm glad to rework it — just didn't want it to sit unnoticed.

